PR #22, opened 2022-06-01
Add versioning and release with changesets

kylesmartin:


packages-ts/gauntlet-starknet-cli/package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "@chainlink/gauntlet-starknet-cli",
+  "name": "@chainlink/starknet-gauntlet-cli",
```
archseer, line 2: The package names changed, but the directories didn't which seems to confuse my LSP.

PR #32, opened 2022-06-14
Multisig integration example

taco-paco:


packages-ts/eqlabs-multisig-integration/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
```
krebernisak: No need to include `./test`, we can use just our standard config here: https://github.com/smartcontractkit/chainlink-starknet/blob/develop/packages-ts/starknet-gauntlet/tsconfig.json (c/p from here)

packages-ts/eqlabs-multisig-integration/test/Multisig.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  it('should submit & confirm transaction', async () => {
+    const nonce = 0
+    const selector = getSelectorFromName('get_signers')
+    const payload = {
+      to: multisig.address,
+      function_selector: selector,
+      calldata: [],
+      nonce,
+    }
```
krebernisak: Please do a `set_treshold` here instead which is a more realistic example. 
Change it from 2 to 3, then execute, and verify the change in the end.

packages-ts/eqlabs-multisig-integration/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "@types/elliptic": "^6.4.14",
+    "@types/mocha": "^9.0.0",
+    "@types/node": "^16.11.10",
+    "chai": "^4.3.4",
+    "ethers": "^5.6.8",
+    "ts-node": "^10.7.0",
+    "typescript": "^4.7.2"
+  },
+  "dependencies": {
+    "@toruslabs/starkware-crypto": "^1.0.0",
```
krebernisak: Do we use this dependency for anything here?

packages-ts/eqlabs-multisig-integration/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@chainlink/starknet-integration-tests-eqlabs-multisig",
```
krebernisak: One additional structure change request: please rename these integration folders to start with `integration-` prefix:

- `packages-ts/eqlabs-multisig-integration` -> `packages-ts/integration-eqlabs-multisig`
- `packages-ts/starkgate-integration` -> `packages-ts/integration-starkgate`

krebernisak (changes requested):

packages-ts/eqlabs-multisig-integration/tsconfig.json
taco-paco: We don't have `src` here or it will be added in the future (?) + we need to add path to the apt hardhat.config.ts

packages-ts/eqlabs-multisig-integration/test/Multisig.test.ts
taco-paco: Done

packages-ts/eqlabs-multisig-integration/package.json
taco-paco: Done

packages-ts/eqlabs-multisig-integration/package.json
taco-paco: No, removed

packages-ts/integration-eqlabs-multisig/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "@types/elliptic": "^6.4.14",
+    "@types/mocha": "^9.0.0",
+    "@types/node": "^16.11.10",
+    "chai": "^4.3.4",
+    "ethers": "^5.6.8",
+    "ts-node": "^10.7.0",
+    "typescript": "^4.7.2"
+  },
+  "dependencies": {
+    "hardhat": "2.9.6",
```
archseer: Can you upgrade hardhat? `2.10` should work with node 18.x

packages-ts/eqlabs-multisig-integration/tsconfig.json
krebernisak: This is just an internal integration package, will not be published ever and probably no src will be added, just tests.

Not sure if there is some problem here with my original suggestion, but it’s a minor thing so feel free to keep it as is so we get this merged.

krebernisak (approved):
Thank you @taco-paco, this looks great!

Maybe just update the hardhat version as the final change, and good to go.

packages-ts/integration-eqlabs-multisig/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "main": "index.js",
+  "directories": {
+    "test": "test"
+  },
+  "scripts": {
+    "test": "npx hardhat test"
+  },
+  "devDependencies": {
+    "@nomiclabs/hardhat-ethers": "^2.0.5",
+    "@chainlink-dev/starknet-multisig": "^0.1.0",
```
krebernisak: Deprecated!

Please double check from Lauri, but I think this is the one we should be using: https://www.npmjs.com/package/starsign-multisig

krebernisak (changes requested):
@PacoEdwin can we please rebase and merge this one, today?

It's a pretty simple PR, and it has been hanging in here for a while. Let's move this forward.

archseer:
I'm actually working on rebasing this now, we also need to update the main gauntlet package

packages-ts/integration-eqlabs-multisig/package.json
archseer: We'll switch in a follow-up, starsign-multisig doesn't contain pre-compiled artifacts

PR #33, opened 2022-06-14
transaction manager

aalu1418:
Initial components for a simple transaction manager (batches and sends) to unblock and connection of other components


pkg/starknet/keys/starkkey.go
```diff
@@ -0,0 +1,125 @@
...
+	return key.PublicKeyStr()
+}
+
+// PublicKeyStr
+func (key Key) PublicKeyStr() string {
+
+	// TODO: what if this is different per network?
+	// https://github.com/Shard-Labs/starknet-devnet/blob/master/starknet_devnet/account.py
+	classHash, _ := new(big.Int).SetString("1803505466663265559571280894381905521939782500874858933595227108099796801620", 10)
+	salt := big.NewInt(20)
```
aalu1418: this will be addressed separately...this used for generating the keys for devnet

pkg/chainlink/ocr2/client_test.go
```diff
@@ -16,7 +18,8 @@ func TestOCR2Client(t *testing.T) {
 	ocr2ContractAddress := "0x756ce9ca3dff7ee1037e712fb9662be13b5dcfc0660b97d266298733e1196b"
 	lggr := logger.Test(t)
 
-	client, err := NewClient(chainID, lggr)
+	cfg := starknet.NewConfig(db.ChainCfg{}, lggr)
+	client, err := NewClient(chainID, "", lggr, cfg)
```
aalu1418: note: using a remotely deployed contract for test cases here makes it a bit flaky with the introduction of underlying configuration/timeouts - so the `RequestTimeout` config is set to 10s to help with this

this would eventually be good to have a test done locally

relayer/pkg/chainlink/ocr2/client.go
```diff
@@ -27,8 +27,8 @@ type Client struct {
 	lggr           logger.Logger
 }
 
-func NewClient(chainID string, lggr logger.Logger) (*Client, error) {
-	client, err := starknet.NewClient(chainID, lggr)
+func NewClient(chainID string, url string, lggr logger.Logger, cfg starknet.Config) (*Client, error) {
```
sdrug, line 30: Since we already pass `url`, do we need cfg in the client at all?

relayer/pkg/chainlink/ocr2/utils.go
```diff
@@ -126,7 +126,7 @@ func parseConfigEventData(eventData []*caigotypes.Felt) (types.ContractConfig, e
 		if i%2 == 0 {
 			signers = append(signers, member.Bytes())
 		} else {
-			transmitters = append(transmitters, types.Account(member.Hex()))
+			transmitters = append(transmitters, types.Account(member.String()))
```
sdrug, line 129: Thanks for fixing this!

relayer/pkg/chainlink/txm/test_helpers.go
```diff
@@ -0,0 +1,97 @@
...
+	"0x23a7711a8133287637ebdcd9e87a1613",
+	"0x1846d424c17c627923c6612f48268673",
+	"0xfcbd04c340212ef7cca5a5a19e4d6e3c",
+	"0xb4862b21fb97d43588561712e8e5216a",
+	"0x259f4329e6f4590b9a164106cf6a659e",
+}
+
+// SetupLocalStarkNetNode sets up a local starknet node via cli, and returns the url
+func SetupLocalStarkNetNode(t *testing.T) string {
+	port := mustRandomPort()
```
sdrug: <just wondering> Why random port? If it's taken, the server won't start anyways..

relayer/pkg/starknet/client.go
```diff
@@ -11,69 +13,99 @@ import (
...
-	if err != nil {
-		return id, errors.Wrap(err, "couldn't get chain id")
-	}
-
-	return id, nil
-}
-
-func (c *Client) LatestBlockHeight(ctx context.Context) (height uint64, err error) {
+func (c *Client) LatestBlockHeight(parentCtx context.Context) (height uint64, err error) {
+	ctx, cancel := context.WithTimeout(parentCtx, c.cfg.RequestTimeout())
```
sdrug: Ah, now I see where `cfg` is used. My idea initially was that the timeout is coming with original context passed to the function (`parentCtx` here). Is there a reason why this idea is not valid?

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -0,0 +1,196 @@
...
+			start := time.Now()
+
+			// calculate total txs to process
+			txLen := len(txm.queue)
+			if txLen > txm.cfg.TxMaxBatchSize() {
+				txLen = txm.cfg.TxMaxBatchSize()
+			}
+
+			// fetch batch and split by sender
+			txsBySender := map[string][]types.Transaction{}
```
sdrug, line 94: <just wondering> What is the reason of grouping batches by sender?

relayer/pkg/chainlink/ocr2/client.go
aalu1418: i think you answered it yourself, but just writing it out here again :)
* the config is used for query timeouts
* the URL will eventually be passed in using the chains/nodes implementation

relayer/pkg/starknet/client.go
aalu1418: i think that would also work! 

personally, I like having the client with a built in timeout that way there is one place (`cfg`) where the timeout is defined for any query. this also allows for the `chainCfg` from the CL node to define parameters that can be later be changed without code changes :)

also it helps because you don't have to worry about defining any query timeouts when you use the client - and you can pass other contexts directly in

relayer/pkg/chainlink/txm/txm.go
aalu1418: it's grouped by sender so each sender can have it's own privatekey/signer (CL nodes can have multiple private keys for a key type)

relayer/pkg/chainlink/txm/test_helpers.go
aalu1418: the idea is that randomness would reduce the chances of a port being taken. but yeah it's a potential risk that the random port may be taken and cause a test flake

relayer/pkg/starknet/client.go
sdrug: My opinion on that is that the client is a utility module and it's behaviour is to be defined by the consumer (such as contract_reader) by passing the context in. It also removes the dependency on cfg that is imo unnecessary here. Definitely we don't want to change the code every time, I just mean that the timeout is to be defined upstream (in contract_reader or even earlier because contract_reader also receives the context from its consumer).

relayer/pkg/chainlink/txm/txm.go
sdrug: I didn't get it, sorry. What's the purpose of that? Why not, let's say, pack the batch up to it's limit every time?

aalu1418: hmmm let me try explaining it with more detail :)

the transaction manager is built in a way that it can sign any transaction assuming it has a key for it. each transaction that is placed in the queue has a `senderAddress` parameter. so if i mixed many txs with different senderAddresses into the queue the txm manager should still be able to send all of them - once the txm pulls a batch from the queue it has to split by address so it can use the correct private key to sign those txs.

if it were to pack the batch up to the limit each time there's two problems:
* we want to get txs on chain in a timely manner, so it's unlikely that a batch would become filled
* if we were to put all txs into a single batch (assuming they have mixed senderAddresses), then it would be only using 1 private key to sign the whole batch. and that private key may not be valid for the other senderAddresses

relayer/pkg/starknet/client.go
aalu1418: that's a valid point!

my main concern is if the upstream implementation doesn't include a timeout, then the request has the potential to hang and block execution. so by having a parameter set on a lower level of the client it reduces the chance of that happening?

if you feel strongly about it, I'm happy to change it :) we also specify a client level timeout in the solana and terra implementations

relayer/pkg/chainlink/txm/test_helpers.go
```diff
@@ -0,0 +1,97 @@
...
+		require.NoError(t, err)
+		raw := keys.Raw(keyBytes)
+
+		keyMap[raw.Key().PublicKeyStr()] = raw.Key()
+
+	}
+	return keyMap
+}
+
+func mustRandomPort() string {
```
archseer: Can't you just bind onto port 0?

> Binding to port 0 is the official documented way to bind to a OS-assigned random port. It is in the bind() documentation: "if the port is specified as zero, the service provider assigns a unique port to the application from the dynamic client port range.

archseer (changes requested):

shell.nix
```diff
@@ -24,6 +24,6 @@
   venvDir = "./.venv";
```
archseer, line 24: I wonder if this needs the same fix btw, otherwise the `.venv` folder might end up under `relayer/`

relayer/pkg/chainlink/txm/test_helpers.go
aalu1418: assigning it to port 0 would work with selecting a new port, but how would we pass the port/url back to the client?

I'll add something to also check if the port is in use like @sdrug suggested

relayer/pkg/chainlink/txm/txm.go
sdrug: Thanks for explanation :)
The batch not being able to contain separately signed txs from different senders - is it our limitation of the tx manager or the network one?

relayer/pkg/starknet/client.go
sdrug: I'm totally fine with you giving it another thought and deciding it's fine this way :)

I noticed that our modules are tightly coupled (i.e. every stores its copy of `cfg`) and it's hard to move them around. I think we should continuously address it since the app is growing. On example of this client, if let's say another module would require to use it (on Solana and Terra there was a getter for it in the `Chain` interface, so the core module can potentially call it), that module would have to comply with this timeout although it didn't set it. Or if this module sets the timeout itself through the context passed to the client, there could be unexpected results.

aalu1418: that's a good point! maybe there is a better way to pass in a default timeout with the option to pass in a null value that indicates to not add an additional timeout. I will take a second look

relayer/pkg/chainlink/txm/txm.go
aalu1418: I believe it's a network limitation where each "payload" sent to the endpoint has one signature set (r + s components) associated with it, but each "payload" can have multiple txs
- caigo implementation: https://github.com/dontpanicdao/caigo/blob/0a04d64ffe5e6792cef20a53ccc545505cd1ee4b/account.go#L127

shell.nix
aalu1418: tried out the change but had to undo it since the CI immediately failed 🤷🏼

relayer/pkg/chainlink/txm/test_helpers.go
aalu1418: added some additional func so it will retry finding open ports if there's a conflict

relayer/pkg/starknet/client.go
aalu1418: changed the interface!
```
NewClient(chainID string, baseURL string, lggr logger.Logger, timeout *time.Duration)
```
now there doesn't have to be a timeout specified for the client - and it can be used directly from the parent context. but if there is a default passed in, then it will use that

sdrug (approved):
LGTM, great job as always :thumbsup:

relayer/go.mod
```diff
@@ -5,7 +5,8 @@ go 1.18
 require (
 	github.com/NethermindEth/juno v0.0.0-20220630151419-cbd368b222ac
 	github.com/btcsuite/btcd/chaincfg/chainhash v1.0.1
-	github.com/dontpanicdao/caigo v0.2.1-0.20220517132056-e34006317632
+	github.com/dontpanicdao/caigo v0.3.0
+	github.com/ethereum/go-ethereum v1.10.18
```
jmank88: Core just bumped to 1.10.20 - or better yet, could this import be avoided?

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -0,0 +1,196 @@
...
+	return txm.starter.StopOnce("starktxm", func() error {
+		close(txm.stop)
+		txm.done.Wait()
+		return nil
+	})
+}
+
+func (txm *starktxm) Healthy() error {
+	// TODO
+	return txm.starter.Healthy()
```
jmank88: Done?
```suggestion
	return txm.starter.Healthy()
```

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -0,0 +1,196 @@
...
+
+func (txm *starktxm) Ready() error {
+	return txm.starter.Ready()
+}
+
+func (txm *starktxm) Enqueue(tx types.Transaction) error {
+	select {
+	case txm.queue <- tx:
+	default:
+		return errors.New("failed to enqueue transaction")
```
jmank88: Should the error include some tx info?

relayer/pkg/chainlink/txm/txm_test.go
```diff
@@ -0,0 +1,74 @@
...
+	for k := range localKeys {
+		for i := 0; i < 5; i++ {
+			require.NoError(t, txm.Enqueue(types.Transaction{
+				SenderAddress:      k,
+				ContractAddress:    k, // send to self
+				EntryPointSelector: "get_nonce",
+			}))
+		}
+	}
+	time.Sleep(30 * time.Second)
```
jmank88, line 71: Is there any way around this long wait? e.g. could we detect if/when the work is done sooner?

relayer/pkg/chainlink/txm/txm_test.go
```diff
@@ -0,0 +1,74 @@
+package txm
```
jmank88, line 3: Should this be behind a build tag like `integration`, since it depends on having `starknet-devnet` installed?

relayer/go.mod
aalu1418: removed! but there's still an indirect import because `caigo` imports it

relayer/pkg/chainlink/txm/txm_test.go
aalu1418: not really at the moment...but future work on the txm will include a confirmer which should then provide an avenue for determining if work is complete or not

archseer (approved):

relayer/go.mod
jmank88: :homer-shakes-fist:

jmank88 (approved):

PR #34, opened 2022-06-15
Add L2 Bridge Commands and Tests

kylesmartin:


packages-ts/starknet-gauntlet-starkgate/src/commands/bridge/setL2Token.ts
```diff
@@ -0,0 +1,65 @@
...
+import {
+  AfterExecute,
+  BeforeExecute,
+  ExecuteCommandConfig,
+  ExecutionContext,
+  makeExecuteCommand,
+  Validation,
+} from '@chainlink/starknet-gauntlet'
+import { shortString } from 'starknet'
+import { isContext } from 'vm'
```
RodrigoAD, line 11:
```suggestion
```
These don't seem to be used. I also see them on `setL1Bridge.ts` and `deploy.ts`. 
`BN` and some others from `starknet-gauntlet` either.

packages-ts/starknet-gauntlet-starkgate/test/commands/bridge.test.ts
```diff
@@ -0,0 +1,164 @@
...
+      const body = {
+        address: account,
+        amount: balance,
+        lite: true,
+      }
+      const response = await fetch(`${gateway_url}/mint`, {
+        method: 'post',
+        body: JSON.stringify(body),
+        headers: { 'Content-Type': 'application/json' },
+      })
```
RodrigoAD, line 58: Does starknet.js have any support for this? If no, could we extract this to `starknet-gauntlet`? I could see this function being used in almost every test

RodrigoAD (commented):
LGTM, just a couple of things

packages-ts/starknet-gauntlet-starkgate/test/commands/bridge.test.ts
archseer: Yeah it's specific to devnet: https://github.com/Shard-Labs/starknet-devnet#mint-token---local-faucet=

There's `fetchEndpoint` but it's `protected` so we can't use it :/ https://github.com/0xs34n/starknet.js/blob/8969decf73cb62096da8804c37be135f624ac777/src/provider/default.ts#L140-L197=

Rather than deploying our own account, we could use a specific seed for the devnet, then use one of those 10 accounts that are pre-allocated.

RodrigoAD: Definitely, that'd be perfect

aalu1418: tbh I prefer deploying our own account just to make sure that command works - also if the starknet-devnet development lags behind contract develop...not sure if any issues could show up there

aalu1418 (approved):
🚀 looks good to me!

PR #41, opened 2022-06-28
Added initial OCR setup and contract deployment to tests

smickovskid:
- [x] update GHA once this is merged: https://github.com/smartcontractkit/chainlink-github-actions/pull/26

**NOTE**: do not squash merge - need to keep commit hashes intact


.tool-versions
```diff
@@ -1,4 +1,4 @@
-nodejs 18.6.0
+nodejs 18.5.0
```
archseer: Why downgrade?

integration-tests/chainlink_models.go
```diff
@@ -0,0 +1,84 @@
...
+// P2PData holds the remote ip and the peer id and port
+type P2PData struct {
+	RemoteIP   string
+	RemotePort string
+	PeerID     string
+}
+
+// OCR2TaskJobSpec represents an OCR2 job that is given to other nodes, meant to communicate with the bootstrap node,
+// and provide their answers
+type OCR2TaskJobSpec struct {
```
krebernisak: Wasn't a type like this exposed by the testing framework?

This is generic code we should reuse across our integration efforts.

integration-tests/gauntlet_starknet.go
```diff
@@ -0,0 +1,161 @@
+package integration_tests
```
krebernisak: This file redeclares a bunch of code and types that should be already available elsewhere, and are common across integrations. Running it is our first priority, but we should also figure out how to improve this situation.

@smickovskid, please keep this in mind and try to reuse it if there is existing code available. If not, let's try to split the code into:

- common code (shared across integrations)
- custom StarkNet-specific code

Let's then put all common code in `relayer/pkg/common` and in-line document there our ad-hoc suggestion where should this code be moved. Then we can go back and do cleanup until we remove the `common` package altogether.

cc @archseer @aalu1418 @sdrug

relayer/ops/environment.go
```diff
@@ -1,45 +1,101 @@
...
+type HelmProps struct {
+	Name   string
+	Path   string
+	Values *map[string]interface{}
+}
+
+type Chart struct {
+	HelmProps *HelmProps
+	Props     *Props
+}
```
krebernisak: Why doesn't the `chainlink-env` module expose these types?

They are also duplicated there: https://github.com/smartcontractkit/chainlink-env/search?q=HelmProps

integration-tests/common.go
```diff
@@ -2,74 +2,212 @@ package integration_tests
...
+type BridgeInfo struct {
+	ObservationSource string
+	JuelsSource       string
+}
+
+type Chart struct {
+	Name   string
+	Path   string
+	Values *map[string]interface{}
+}
```
krebernisak: Again redeclaring the `Chart` type

integration-tests/common.go
```diff
@@ -2,74 +2,212 @@ package integration_tests
...
+	Env             *environment.Environment
+	sc              *StarkNetClient
+	chainlinkNodes  []*client.Chainlink
+	mockServer      *ctfClient.MockserverClient
+	nodeKeys        []NodeKeysBundle
+	starknetNetwork *blockchain.EVMNetwork
+	bTypeAttr       *client.BridgeTypeAttributes
+)
+
+type StarkNetClient struct {
```
krebernisak: I don't really understand the scope of this "client":

- It interacts with the CL node
- It interacts with the StarkNet node
- It also expects/asserts so acts as a test

We should have a `ChainlinkClient` a `StarkNetDevnetClient` and a test, instead of combining all of that functionality into one god object.

.tool-versions
smickovskid: Fixed

integration-tests/chainlink_models.go
smickovskid: Removed

relayer/ops/environment.go
smickovskid: Removed and adjusted

integration-tests/common.go
smickovskid: God object decoupled

integration-tests/common.go
smickovskid: Removed

relayer/ops/charts/starknet/templates/deployment.yaml
```diff
@@ -22,7 +22,8 @@ spec:
 {{- if eq .Values.real_node true }}
         image: "{{ .Values.repository | default "eqlabs/pathfinder"}}:{{ .Values.tag | default "v0.1.8-alpha"}}"
 {{- else }}
-        image: "{{ .Values.repository | default "shardlabs/starknet-devnet"}}:{{ .Values.tag | default "0.1.23"}}"
+        image: "{{ .Values.repository | default "shardlabs/starknet-devnet"}}:{{ .Values.tag | default "0.2.6"}}"
```
archseer: Should be `0.2.8` to pull in the latest fixes

relayer/ops/environment.go
```diff
@@ -1,45 +1,88 @@
...
+
+func defaultProps() *ethereum.Props {
+	return &ethereum.Props{
+		NetworkName: "starknet-dev",
+		Values: map[string]interface{}{
+			"replicas": "1",
+			"starknet-dev": map[string]interface{}{
+				"image": map[string]interface{}{
+					"image":   "shardlabs/starknet-devnet",
+					"version": "v0.2.6",
```
archseer: Same here, `0.2.8`

integration-tests/gauntlet_starknet.go
krebernisak: More resources on figuring out commons:

- Solana: https://github.com/smartcontractkit/chainlink-solana/blob/develop/tests/e2e/common/common.go
- Terra: https://github.com/smartcontractkit/chainlink-terra/blob/develop/tests/e2e/common/common.go

cc @skudasov

relayer/ops/environment.go
```diff
@@ -1,45 +1,88 @@
...
 					},
 				},
+				"seed":      "123",
+				"real_node": "false",
 			},
 		},
 	}
 }
+
+func New(props *ethereum.Props) environment.ConnectedChart {
```
krebernisak:
```suggestion
func New(props *Props) environment.ConnectedChart {
```

relayer/ops/environment.go
```diff
@@ -1,45 +1,88 @@
...
+	if props == nil {
+		props = defaultProps()
+	}
+	return Chart{
+		HelmProps: &ethereum.HelmProps{
```
krebernisak:
```suggestion
		HelmProps: &HelmProps{
```

relayer/ops/environment.go
```diff
@@ -1,45 +1,88 @@
...
-	return &environment.Config{
-		NamespacePrefix: "chainlink-starknet",
-		Charts: environment.Charts{
-			"geth": {Index: 1},
-			"starknet": {
-				Index: 1,
-				Path:  "../../relayer/ops/charts/starknet",
-				Values: map[string]interface{}{
-					"real_node": false,
+type Chart ethereum.Chart
```
krebernisak:
```suggestion
// TODO: remove type aliases once upstream is refactored
type Chart ethereum.Chart
```

integration-tests/smoke/suite_test.go
```diff
@@ -4,12 +4,12 @@ package smoke_test
 import (
 	"testing"
 
-	"github.com/smartcontractkit/chainlink-testing-framework/actions"
+	"github.com/smartcontractkit/chainlink/integration-tests/actions"
```
krebernisak, line 7: Hmm, this seems weird that this fn is sourced from the Chainlink core repo instead of the CTF 🤔

relayer/ops/environment.go
```diff
@@ -1,45 +1,88 @@
...
+
+func defaultProps() *ethereum.Props {
+	return &ethereum.Props{
+		NetworkName: "starknet-dev",
+		Values: map[string]interface{}{
+			"replicas": "1",
+			"starknet-dev": map[string]interface{}{
+				"image": map[string]interface{}{
+					"image":   "shardlabs/starknet-devnet",
+					"version": "v0.2.8",
```
krebernisak: We'll need to think of a way to keep all these versions in check.

integration-tests/gauntlet_starknet.go
```diff
@@ -0,0 +1,244 @@
...
+	err = json.Unmarshal(gauntletOutput, &payload)
+	if err != nil {
+		return payload, err
+	}
+
+	return payload, nil
+}
+
+// Loads and returns the default starknet gauntlet config
+func LoadOCR2Config(nKeys []ctfClient.NodeKeysBundle) (*OCR2Config, error) {
```
krebernisak: This function probably should not be in this file, looks more like a test setup and should find a better home.

krebernisak: Move it to `integration-tests/smoke/ocr2_test.go`, load it there and pass it in as a parameter.

integration-tests/common.go
```diff
@@ -2,77 +2,173 @@ package integration_tests
...
-	Type                      string        `mapstructure:"type" yaml:"type"`
-	PrivateKeys               []string      `mapstructure:"private_keys" yaml:"private_keys"`
-	ChainlinkTransactionLimit uint64        `mapstructure:"chainlink_transaction_limit" yaml:"chainlink_transaction_limit"`
-	Timeout                   time.Duration `mapstructure:"transaction_timeout" yaml:"transaction_timeout"`
-	MinimumConfirmations      int           `mapstructure:"minimum_confirmations" yaml:"minimum_confirmations"`
-	BlockGasLimit             uint64        `mapstructure:"block_gas_limit" yaml:"block_gas_limit"`
+const (
+	gethClient      = "Hardhat"
+	devnetClient    = "starknet-dev"
+	chainLinkClient = "chainlink"
```
krebernisak:
```suggestion
	serviceKeyL1        = "Hardhat"
	serviceKeyL2        = "starknet-dev"
	serviceKeyChainlink = "chainlink"
```

integration-tests/common.go
```diff
@@ -2,77 +2,173 @@ package integration_tests
...
 		cancel: cancel,
 		cfg:    cfg,
-		urls:   urls,
-		client: resty.New().SetBaseURL(urls[0].String()),
+		client: resty.New().SetBaseURL(t.GetStarkNetAddress()),
+	}
+	// Currently not needed since we are testing on L2
+	// if err := c.init(); err != nil {
+	// 	return nil, err
+	// }
```
krebernisak: This logic should probably not be part of this constructor, but a specific test should do it if it requires it + the fn name `init` should probably be made less generic.

integration-tests/common.go
```diff
@@ -2,77 +2,173 @@ package integration_tests
...
+
+}
+
+// Sets up the starknet client
+func (t *Test) SetupClients() {
+	t.sc, err = t.NewStarkNetDevnetClient(&blockchain.EVMNetwork{
+		Name: "starknet-dev",
+		URL:  t.Env.URLs[devnetClient][1],
+		PrivateKeys: []string{
+			"c4da537c1651ddae44867db30d67b366",
```
krebernisak: Also redeclared in `smoke/ocr2_test.go`, consider extracting this as a specific env (default) somewhere.

relayer/ops/environment.go
smickovskid: This will not go in the final PR, we will add it to the deployment yml, now it is hardcoded for testing purposes

integration-tests/common.go
smickovskid: Added `ConfigureL1Messaging()` func

integration-tests/common.go
smickovskid: Moved to `starknet` package and set default values as well as getters.

integration-tests/gauntlet_starknet.go
smickovskid: Moved and adjusted

relayer/pkg/starknet/starknet.go
```diff
@@ -25,6 +25,9 @@ const (
 	serviceKeyChainlink = "chainlink"
 	ChainName           = "starknet"
 	ChainId             = "devnet"
+	// These are one of the the default addresses based on the seed we pass to devnet which is 123
```
krebernisak: These are defined by a specific test env we seed with 123.

They shouldn't be part of our core starknet package of the relayer module, but part of the integration-test module where we should define our env where we test.

relayer/pkg/starknet/starknet.go
```diff
@@ -119,10 +122,10 @@ func (t *Test) DeployEnv(nodes int) {
...
 func (t *Test) SetupClients() {
 	t.sc, err = t.NewStarkNetDevnetClient(&blockchain.EVMNetwork{
-		Name: "starknet-dev",
+		Name: serviceKeyL2,
 		URL:  t.Env.URLs[serviceKeyL2][1],
 		PrivateKeys: []string{
-			"c4da537c1651ddae44867db30d67b366",
+			defaultWalletPrivKey,
 		},
 	})
```
krebernisak: How did this code end up in the `relayer` module? :)

We should move it to the `integration-tests` module.

relayer/go.mod
```diff
@@ -196,3 +139,13 @@ require (
...
+
+replace github.com/smartcontractkit/chainlink-starknet/relayer => ../../chainlink-starknet/relayer
+
+replace github.com/smartcontractkit/chainlink-env => ../../chainlink-env
+
+replace github.com/smartcontractkit/chainlink-testing-framework => ../../chainlink-testing-framework
+
+replace github.com/smartcontractkit/chainlink => ../../chainlink
+
+replace github.com/smartcontractkit/chainlink/integration-tests => ../../chainlink/integration-tests
```
krebernisak: We should have an pre-prepared  but commented out `go.work` file added to the repo, so we can do those replaces there and avoid doing them in module `go.mod` so we don't accidentally forget it.

relayer/pkg/chainlink/ocr2/client.go
```diff
@@ -159,11 +160,13 @@ func (c *Client) LatestTransmissionDetails(ctx context.Context, address string)
 }
 
 func (c *Client) ConfigFromEventAt(ctx context.Context, address string, blockNum uint64) (cc ContractConfig, err error) {
-	block, err := c.r.BlockByNumberGateway(ctx, blockNum)
+	block, err := c.r.BlockByNumberGateway(ctx, blockNum+1) // HAXX: temporary workaround for devnet 0.2.8
```
krebernisak: `starknet-devnet` v0.2.9 should now be released 🎉 🤞

relayer/pkg/chainlink/ocr2/client_test.go
```diff
@@ -15,12 +17,12 @@ import (
 
 func TestOCR2Client(t *testing.T) {
```
krebernisak, line 28: This test the way it is currently setup (depends on real infra) should be moved to `integration-tests` module.

relayer/pkg/starknet/deployer_starknet.go
```diff
@@ -1,4 +1,4 @@
-package integration_tests
+package starknet
```
krebernisak: Why this move?

The deployer should be moved back to `integration_tests` or removed completely if not used anymore, as I think we have a new Gauntlet deployer now.

relayer/pkg/starknet/starknet.go
```diff
@@ -0,0 +1,70 @@
+package starknet
```
krebernisak: This source file should be moved to `ops` module (currently under `relayer/ops`) to a `devnet` package, as this is the context of this file.

integration-tests/common/test_common.go
```diff
@@ -0,0 +1,253 @@
...
+	ctfClient "github.com/smartcontractkit/chainlink-testing-framework/client"
+	"github.com/smartcontractkit/chainlink/integration-tests/client"
+	"strings"
+	"time"
+)
+
+const (
+	ChainName = "starknet"
+	ChainId   = "devnet"
+	// These are one of the default addresses based on the seed we pass to devnet which is 123
```
krebernisak: We should define the `defaultSeed = 123` here and use it for our setup.

integration-tests/common/test_common.go
```diff
@@ -0,0 +1,253 @@
...
+)
+
+var (
+	err error
+)
+
+type Test struct {
+	sc         *StarkNetDevnetClient
+	cc         *ChainlinkClient
+	mockServer *ctfClient.MockserverClient
```
krebernisak, line 55: What is the mock server used for?

integration-tests/common/common.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func GetServiceKeyL1() string {
+	return serviceKeyL1
+}
+func GetServiceKeyL2() string {
+	return serviceKeyL2
+}
+
+func SetDefaultP2PPort(port string) {
```
krebernisak: We should avoid having mutable global state like this.

We should rather encapsulate it within a specific type, then mutate and run functions on a specific instance of that type.

relayer/pkg/chainlink/ocr2/client.go
archseer: Looks like this PR is rebased on an older version of my PR

integration-tests/common/common.go
smickovskid: Adjusted, added common state to handle variables

relayer/pkg/starknet/starknet.go
smickovskid: Moved and renamed

relayer/pkg/starknet/deployer_starknet.go
smickovskid: Removed since unused

relayer/pkg/starknet/starknet.go
smickovskid: Adjusted and moved

relayer/pkg/starknet/starknet.go
smickovskid: Adjusted and moved to the setup in the test itself

integration-tests/gauntlet_starknet.go
smickovskid: Adjusted and re-structured

relayer/ops/charts/hardhat/values.yaml
```diff
@@ -8,7 +8,7 @@ service:
 resources:
   limits:
     cpu: 5000m
-    memory: 5120Mi
+    memory: 8000Mi
```
archseer: Hmm is this necessary? The current values for all the pods are so high that my 16GB RAM laptop refused to spawn 5 core nodes (last one was Pending, waiting for more system resources to free up). I've had to modify the settings locally to only spawn 3 nodes.

relayer/ops/devnet/devnet.go
```diff
@@ -34,16 +35,17 @@ func AutoSyncL1(ctx context.Context) {
...
 
-// FundAccounts Funds provided accounts with 500000 eth each
+// FundAccounts Funds provided accounts with 100 eth each
 func FundAccounts(l2AccList []string) error {
 	for _, key := range l2AccList {
-		_, err := rClient.R().SetBody(map[string]interface{}{
+		res, err := rClient.R().SetBody(map[string]interface{}{
 			"address": key,
-			"amount":  500000,
+			"amount":  1e21,
```
archseer: 👍🏻 I had the same issue in tests

.github/workflows/relayer.yml
```diff
@@ -29,6 +29,14 @@ jobs:
...
       - name: Checkout sources
         uses: actions/checkout@f25a3a9f25bd5f4c5d77189cab02ff357b5aedeb # v2.4.1
+      - uses: smartcontractkit/tool-versions-to-env-action@v1.0.7
+        id: tool-versions
+      - name: Setup Node ${{ steps.tool-versions.outputs.nodejs_version }}
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ steps.tool-versions.outputs.nodejs_version }}
+      - name: Install
+        run: cd .. && yarn install --frozen-lockfile
```
archseer: This is now mixing Nix and tool-versions. I could just replace these with the Nix shell?

relayer/ops/charts/hardhat/values.yaml
aalu1418: i'll drop it back down :) i was testing something to see if I could make devnet more reliable

aalu1418: also not sure if this is related, but looks like I changed the `hardhat` network values which isn't what we're using 🤔

.github/workflows/relayer.yml
aalu1418: sounds good to me!

.github/workflows/e2e_custom_cl.yml
```diff
@@ -6,7 +6,7 @@ on:
       cl_branch_ref:
         description: Chainlink repo branch to integrate with
         required: true
-        default: develop
+        default: starknet-bump
```
archseer: Note: Let's change this back after the PR is merged in core

.github/workflows/relayer.yml
```diff
@@ -29,17 +29,22 @@ jobs:
...
+      - uses: smartcontractkit/tool-versions-to-env-action@v1.0.7
+        id: tool-versions
 
       - name: Install Nix
         uses: cachix/install-nix-action@v17
         with:
           nix_path: nixpkgs=channel:nixos-unstable
 
+      - name: Install Yarn dependencies
+        run: cd .. && nix develop -c yarn install --frozen-lockfile
```
archseer, line 41: I think we can drop this again, relay tests don't use npm deps

integration-tests/common/test_common.go
```diff
@@ -0,0 +1,216 @@
...
+	"github.com/smartcontractkit/chainlink-starknet/relayer/ops"
+	"github.com/smartcontractkit/chainlink-starknet/relayer/ops/devnet"
+	"github.com/smartcontractkit/chainlink-testing-framework/blockchain"
+	ctfClient "github.com/smartcontractkit/chainlink-testing-framework/client"
+	"github.com/smartcontractkit/chainlink/integration-tests/client"
+)
+
+const (
+	ChainName = "starknet"
+	ChainId   = gateway.GOERLI_ID
```
archseer: Let's add a note that this is the default chain ID used on devnet

integration-tests/smoke/ocr2_test.go
```diff
@@ -1,85 +1,241 @@
...
+		accessControllerAddress string
+		ocrAddress              string
+		sg                      *ops.StarknetGauntlet
+		t                       *common.Test
+		nAccounts               []string
+		serviceKeyL1            = "Hardhat"
+		serviceKeyL2            = "starknet-dev"
+		serviceKeyChainlink     = "chainlink"
+		chainName               = "starknet"
+		chainId                 = gateway.GOERLI_ID
```
archseer, line 49: I wonder if we could share this with test_common so the two vars are in sync

relayer/ops/charts/devnet/values.yaml
```diff
@@ -7,10 +7,11 @@ service:
...
   limits:
-    cpu: 1000m
-    memory: 1024Mi
+    cpu: 5000m
+    memory: 5120Mi
   requests:
-    cpu: 1000m
-    memory: 1024Mi
+    cpu: 5000m
+    memory: 5120Mi
```
archseer: Let's try `2GB` instead of 5

relayer/ops/devnet/environment.go
```diff
@@ -0,0 +1,88 @@
...
+				"resources": map[string]interface{}{
+					"requests": map[string]interface{}{
+						"cpu":    "1000m",
+						"memory": "1024Mi",
+					},
+					"limits": map[string]interface{}{
+						"cpu":    "1000m",
+						"memory": "1024Mi",
+					},
+				},
```
archseer: What's the difference between the chart above and these limits here?

relayer/pkg/chainlink/keys/ocr2key.go
```diff
@@ -121,14 +108,12 @@ func (sk *OCR2Key) MaxSignatureLength() int {
 
 func (sk *OCR2Key) Marshal() ([]byte, error) {
 	// https://github.com/ethereum/go-ethereum/blob/07508ac0e9695df347b9dd00d418c25151fbb213/crypto/crypto.go#L159
-	return starknet.PadBytesBigInt(sk.privateKey.D, sk.privateKeyLen()), nil
+	return starknet.PadBytesBigInt(sk.priv, sk.privateKeyLen()), nil
 }
 
 func (sk *OCR2Key) privateKeyLen() int {
 	// https://github.com/NethermindEth/juno/blob/3e71279632d82689e5af03e26693ca5c58a2376e/pkg/crypto/weierstrass/weierstrass.go#L377
```
archseer, line 114: comment can be dropped

relayer/pkg/chainlink/ocr2/utils.go
```diff
@@ -96,6 +99,7 @@ func isEventFromContract(event *caigotypes.Event, address string, eventName stri
 	return event.Keys[0].Cmp(eventKey) == 0
 }
```
archseer, line 128: Need to re-enable the address check in the method above (isEventFromContract)

archseer: For address sanitization, I like the `HexToBytes` function we have in the ocr2keys test and we probably want a few helpers for the reverse: `"0x"+hex.EncodeToString(starknet.PadBytes(felt.Bytes(), 32))` etc

relayer/pkg/chainlink/ocr2/client_test.go
```diff
@@ -1,49 +1,87 @@
...
 	lggr := logger.Test(t)
+	g, err := ops.NewStarknetGauntlet("../../../../")
+	g.SetupNetwork(url)
+
+	// set env vars
+	require.NoError(t, os.Setenv("PRIVATE_KEY", "0x"+hex.EncodeToString(maps.Values(keys)[0].Raw())))
+	require.NoError(t, os.Setenv("ACCOUNT", maps.Keys(keys)[0]))
+	require.NoError(t, os.Setenv("BILLING_ACCESS_CONTROLLER", ZERO_ADDRESS))
+
+	// clean up env vars
```
aalu1418, line 44: TODO

.github/workflows/relayer.yml
aalu1418: actually it's still needed :)

this test is now an integration test that uses gauntlet underneath to deploy the contract: https://github.com/smartcontractkit/chainlink-starknet/blob/gauntlet_ocrv2_test_setup/relayer/pkg/chainlink/ocr2/client_test.go

relayer/ops/devnet/environment.go
aalu1418: @smickovskid or @skudasov might be able to answer this one

.github/workflows/e2e_custom_cl.yml
aalu1418: actually, we can point this directly at `develop` right now since that's what i've been testing against locally so it should work. the flow will pull the chainlink repo, `go get & go mod tidy` the latest commit on this branch, then build

integration-tests/smoke/ocr2_test.go
aalu1418: hmmm not sure if we want to share these, since eventually `common` would be used for `smoke` and `soak`
In my mind:
* `smoke`: quick test against local devnet
* `soak`: can be pointed to any network - devnet, testnet, mainnet

integration-tests/common/test_common.go
```diff
@@ -0,0 +1,214 @@
...
+	ctfClient "github.com/smartcontractkit/chainlink-testing-framework/client"
+	"github.com/smartcontractkit/chainlink/integration-tests/client"
+)
+
+const (
+	ChainName = "starknet"
+	ChainId   = gateway.GOERLI_ID // default chainID for local devnet and live testnet
+	// These are one of the default addresses based on the seed we pass to devnet which is 123
+	defaultWalletAddress = "0x6e3205f9b7c4328f00f718fdecf56ab31acfb3cd6ffeb999dcbac41236ea502"
+	defaultWalletPrivKey = "0xc4da537c1651ddae44867db30d67b366"
```
aalu1418: this can also be cleaned up to use the keys defined in `ops/test_helpers.go` with `--seed 0`

integration-tests/common/test_common.go
aalu1418: i think it's used to mock an EA, but i wasn't sure how it worked

integration-tests/common/test_common.go
```diff
@@ -0,0 +1,214 @@
...
+				"version": clVersion,
+			},
+		}
+	}
+
+	t.Env = environment.New(&environment.Config{
+		NamespacePrefix: "chainlink-smoke-ocr-starknet-ci",
+		InsideK8s:       false,
+	}).
+		//AddHelm(hardhat.New(nil)).
```
aalu1418: TODO: reenable with lower resource usage

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/setConfig.ts
```diff
@@ -29,8 +29,19 @@ const makeContractInput = async (input: SetConfigInput): Promise<ContractInput>
...
+    k.replace('ocr2off_starknet_', ''),
+  )
+  input.offchainConfig.configPublicKeys = input.offchainConfig.configPublicKeys.map((k) =>
+    k.replace('ocr2cfg_starknet_', ''),
+  )
+
   const { offchainConfig } = await encoding.serializeOffchainConfig(input.offchainConfig, input.secret)
-  return [oracles, new BN(input.f).toNumber(), input.onchainConfig, 2, bytesToFelts(offchainConfig)]
+  // TODO: encode input.onchainConfig to a sequence of felts
+  let onchainConfig = []
```
aalu1418: Open TODO

relayer/ops/charts/hardhat/values.yaml
```diff
@@ -0,0 +1,14 @@
...
+  type: ClusterIP
+  externalPort: 8545
+  internalPort: 8545
+resources:
+  limits:
+    cpu: 5000m
+    memory: 5120Mi
+  requests:
+    cpu: 5000m
+    memory: 5120Mi
```
aalu1418: adjust these to similar starknet values `2000m cpu` `2048Mi memory`

relayer/pkg/chainlink/ocr2/client_test.go
aalu1418: resolving this comment as it has been turned into an integration test :)

relayer/pkg/chainlink/ocr2/config_digester.go
```diff
@@ -7,7 +7,6 @@ import (
 	"strings"
 
 	"github.com/NethermindEth/juno/pkg/crypto/pedersen"
-	"github.com/NethermindEth/juno/pkg/rpc"
```
aalu1418, line 10: note: removing this `juno/pkg/rpc` cleaned up a lot of dependencies and also fixed some build contraints - it's also just simpler types now

relayer/pkg/starknet/client_test.go
```diff
@@ -12,16 +15,29 @@ import (
...
+	var wg sync.WaitGroup
+	wg.Add(1) // mock endpoint only called once
+	mockServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		_, err := w.Write([]byte(`{"parent_block_hash": "0x0", "status": "ACCEPTED_ON_L2", "timestamp": 1660840417, "gas_price": "0x174876e800", "sequencer_address": "0x4bbfb0d1aab5bf33eec5ada3a1040c41ed902a1eeb38c78a753d6f6359f1666", "transactions": [], "transaction_receipts": [], "state_root": "030c9b7339aabef2d6c293c40d4f0ec6ffae303cb7df5b705dce7acc00306b06", "starknet_version": "0.9.1", "block_hash": "0x0", "block_number": 0}`))
+		require.NoError(t, err)
+		wg.Done()
+	}))
+	defer mockServer.Close()
+
 	// todo: adjust for e2e tests
```
aalu1418: done - need to remove todo comment

aalu1418 (commented):
notes on still open items

integration-tests/common/test_common.go
aalu1418: update: can't be re-enabled yet b/c in local testing it never shows a ready state

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/setConfig.ts
aalu1418: gauntlet should not be passing in the onchainConfig - it is generated onchain and emitted via event. onchainConfig has been implemented

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -143,24 +138,32 @@ func (txm *starktxm) run() {
...
+	fee, err := account.EstimateFee(ctx, txs, caigo.ExecuteDetails{})
 	if err != nil {
 		return txhash, errors.Errorf("failed to estimate fee: %s", err)
 	}
 
+	details := caigo.ExecuteDetails{
+		MaxFee: &types.Felt{
+			Int: new(big.Int).SetUint64((fee.OverallFee * FEE_MARGIN) / 100),
+		},
+	}
```
archseer, line 160: This just duplicates the default internal logic: if you provide no max fee and don't call EstimateFee it will do the same thing internally. I inlined this since I figured we might want to weak the FEE_MARGIN a bit

contracts/test/ocr2/aggregator-test.ts
```diff
@@ -188,10 +188,27 @@ describe('aggregator.cairo', function () {
...
+      observers,
+      observations.length,
+      ...observations,
+      juels_per_fee_coin,
+    ])
+    console.log(msg)
     for (let oracle of oracles.slice(0, f + 1)) {
       let [r, s] = ec.sign(oracle.signer, msg)
+      console.log(
+        `privKey ${oracle.signer.getPrivate()} r ${r} s ${s} pubKey ${number.toBN(ec.getStarkKey(oracle.signer))}`,
```
sdrug, line 207: Is it necessary to log a private key here? 
Seems unsafe even for testing env..

aalu1418: hmmm i think this is ok since we use the info from this log statement to test the key signing algorithms - so the logged information is then just hard coded in the go test - so it's already exposed there.

also, i believe it is basically impossible to generate a private key attached to a real account...so generating and logging one-off keys should be ok

integration-tests/smoke/ocr2_test.go
```diff
@@ -1,85 +1,250 @@
...
+				}
+
+				time.Sleep(5 * time.Second)
+
+				res, err := client.LatestTransmissionDetails(ctx, ocrAddress)
+				if err != nil {
+					log.Error().Err(err)
+
+					// exit loop early
+					if stuck && stuckCount > 10 {
```
sdrug: Why also check for `stuckCount` here? 
Maybe cut the stuck part here in `if` and break the loop in the end when `stuckCount` exceeds certain value. This way the `stuck` variable isn't needed and the loop is broken immediately..

relayer/pkg/chainlink/ocr2/utils.go
```diff
@@ -86,21 +106,31 @@ func caigoFeltsToJunoFelts(cFelts []*caigotypes.Felt) (jFelts []*big.Int) {
...
 
 func isEventFromContract(event *caigotypes.Event, address string, eventName string) bool {
-	// temp: disable because of address format differences (with/without leading zeros)
-	// if event.FromAddress != address {
-	// 	return false
-	// }
+	fromBytes, err := keys.HexToBytes(event.FromAddress)
```
sdrug: Address compare func to utils?

relayer/pkg/chainlink/txm/txm.go
sdrug: +1

sdrug:
@aalu1418 @archseer 
All in all looks good to me 👍

integration-tests/smoke/ocr2_test.go
aalu1418: moved the end condition logic to the beginning to match the other end condition logic :)

aalu1418 (approved):
Approving!

@sdrug said it looks good 🎉 - https://github.com/smartcontractkit/chainlink-starknet/pull/41#issuecomment-1222559238

any future changes should be done in separate PRs - this PR is getting a bit too unwieldy

PR #48, opened 2022-07-11
move aggregator consumer into examples/contracts/aggregator-consumer

qperrot:


archseer:
Looks like you need to modify test.sh to run tests from the new directory

archseer (approved):

PR #52, opened 2022-07-18
pin build-custom since it is being deprecated

tateexon:
Pinning this for now since it will be removed shortly and I don't want development on starknet to accidentally be halted. There is a ticket to use the new actions in progress by @AnieeG so this pin should just be a short term stop gap.


tateexon:
Looks like there is more broken then just the build pin so pinning this branch would do nothing.

PR #56, opened 2022-07-22
L2 emergency protocol

taco-paco:


aiagentdev-1990 (dismissed):
LGTM after our previous discussions on this PR

contracts/contracts/SimpleWriteAccessController/library.cairo
```diff
@@ -0,0 +1,142 @@
...
+            return (TRUE)
+        end
+
+        return (FALSE)
+    end
+
+    # TODO: Why having address here as an argument?
+    func check_access{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
+        address : felt
+    ):
```
archseer: Where is this sourced from? The address here should be `user` and is used to verify the sender's access

contracts/contracts/SimpleWriteAccessController/simple_write_access_controller.cairo
```diff
@@ -0,0 +1,41 @@
...
+) -> (bool : felt):
+    let (has_access) = simple_write_access_controller.has_access(user, data_len, data)
+    return (has_access)
+end
+
+# implements IAccessController
+@view
+func check_access{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
+    address : felt
+):
```
archseer: Same here

contracts/contracts/SimpleWriteAccessController/library.cairo
```diff
@@ -0,0 +1,142 @@
...
+        Ownable_initializer(owner_address)
+        s_check_enabled.write(TRUE)
+
+        return ()
+    end
+
+    # Virtual method needs wrapper
+    func has_access{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
+        user : felt, data_len : felt, data : felt*
```
archseer: I wonder if we should remove `data` from here since on Cairo we can't pass `calldata` through in a generic way anyway and it's not checked anywhere

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
```diff
@@ -0,0 +1,250 @@
...
+func s_l2_cross_domain_messenger() -> (address : felt):
+end
+
+@storage_var
+func s_rounds(id : felt, field : felt) -> (res : felt):
+end
+
+@storage_var
+func s_latest_round_id() -> (res : felt):
+end
```
archseer: I think `library` should always namespace storage variables (`SequencerUptimeFeed_latest_round_id`) etc. to avoid name collisions if it's imported elsewhere. I saw this pattern from OpenZeppelin's `library.cairo` implementations

contracts/contracts/ocr2/aggregator.cairo
```diff
@@ -984,6 +984,8 @@ func billing_set(
 ):
 end
 
+# TODO: Why here and not in OffchainAggregatorBilling
+# Looks like they are merged now
```
archseer: Yeah we just have a single implementation outside of EVM

archseer:
Can you also resolve the `yarn.lock` conflict?

contracts/contracts/SimpleWriteAccessController/library.cairo
```diff
@@ -0,0 +1,142 @@
+%lang starknet
```
krebernisak: @archseer are there other contracts as candidates to be restructured like this, to enable composability later?

contracts/contracts/SimpleWriteAccessController/library.cairo
```diff
@@ -0,0 +1,142 @@
...
+
+from starkware.cairo.common.alloc import alloc
+from starkware.cairo.common.bool import TRUE, FALSE
+from starkware.cairo.common.cairo_builtins import HashBuiltin
+
+from ownable import Ownable_only_owner, Ownable_initializer
+
+struct Phase:
```
krebernisak: This struct looks unused.

contracts/contracts/SimpleWriteAccessController/library.cairo
krebernisak: It's part of an interface that is unused now but could be used in the future by a non-simple access controller, for example, to verify a signature or something like that.

I'd suggest we leave it to keep our interfaces consistent.

contracts/contracts/SimpleWriteAccessController/library.cairo
```diff
@@ -0,0 +1,142 @@
...
+
+        let (check_enabled) = s_check_enabled.read()
+        if check_enabled == FALSE:
+            return (TRUE)
+        end
+
+        return (FALSE)
+    end
+
+    # TODO: Why having address here as an argument?
```
krebernisak: Because a caller could check on behalf of someone else if the question is why don't we just verify access for the sender?

contracts/contracts/SimpleReadAccessController/library.cairo
```diff
@@ -0,0 +1,50 @@
...
+
+    func has_access{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
+        user : felt, data_len : felt, data : felt*
+    ) -> (bool : felt):
+        let (has_access) = simple_write_access_controller.has_access(user, data_len, data)
+        if has_access == TRUE:
+            return (TRUE)
+        end
+
+        let (tx_info) = get_tx_info()
```
krebernisak: @archseer @taco-paco I'm not sure this will work as intended.

1. On Ethereum we check if `userAddr == tx.origin` which gives access only to EOA (we use it to check sender, so it checks if sender equals origin), meaning off-chain access. An Ethereum account (signer) can call this method directly and read, while indirectly via other contract calls it will fail (unless explicit whitelist).

2. Now on StarkNet, we check if `userAddr == tx.account` which gives access only to StarkNet accounts (we use it to check sender, so it checks if sender equals account), which are contracts on StarkNet. 

So now someone can deploy a custom account contract, and sign a transaction that will first read from the feed (successfully) and can then use that data to share it with another contract using this custom implemented account.

We should reconsider this workflow. Any thoughts?

contracts/contracts/SimpleWriteAccessController/library.cairo
archseer: I don't think so, I already extracted `Ownable`

contracts/contracts/SimpleWriteAccessController/library.cairo
taco-paco: That's namings from [develop](https://github.com/smartcontractkit/chainlink-starknet/blob/develop/contracts/contracts/ocr2/interfaces/IAccessController.cairo). Will rename everywhere

contracts/contracts/SimpleReadAccessController/library.cairo
krebernisak: Thinking on it some more, it should be ok to leave as is for now:

- We can easily update the AC if necessary
- We'll ask for suggestions during audit
- Even if custom account can be used to read prices, they would be of little use as any other contract couldn't verify the data but would need to trust the user. It could accept calls from only specific custom classes of contracts, but then the downstream protocol takes a lot of risk by depending of infrastructure Chainlink can update.

Let's add a comment here:

```suggestion
        // NOTICE: access is granted to account contracts, to enable off-chain reads. The account abstraction architecture can be used to deploy a custom contract that could access the data and pass it forward to a contract, but a workflow like that would be of little use because the contract would need to trust user inputs as it can't verify the data itself. While the financial contract could accept calls from verified account contracts only (to verify data is correct), it would be a risky strategy to depend on this part of infrastructure that can easily be upgraded.
        let (tx_info) = get_tx_info()
```

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
taco-paco: I'm not sure that this is necessary. I think if child class has intersecting variable names then probably smth off with architecture or those variables named improperly incorrectly reflecting their meaning. If this is for the case where we make functions from library reusable not in class that inherit from it then such functions shall be placed in `utils.cairo`

contracts/contracts/SimpleReadAccessController/library.cairo
taco-paco: Added comment

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
```diff
@@ -0,0 +1,258 @@
...
+
+func require_access{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}():
+    let (address) = get_caller_address()
+    simple_read_access_controller.check_access(address)
+
+    return ()
+end
+
+# TODO: this methods are virtual in .sol: clarify
```
krebernisak: A Solidity keyword:

> A function that allows an inheriting contract to override its behavior will be marked at virtual

contracts/contracts/SimpleWriteAccessController/library.cairo
```diff
@@ -0,0 +1,136 @@
...
+    func constructor{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
+        owner_address : felt
+    ):
+        Ownable_initializer(owner_address)
+        s_check_enabled.write(TRUE)
+
+        return ()
+    end
+
+    # Virtual method needs wrapper
```
krebernisak: ?

contracts/contracts/SimpleWriteAccessController/library.cairo
```diff
@@ -0,0 +1,136 @@
...
+
+# Adds an address to the access list
+@external
+func add_access{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(user : felt):
+    Ownable_only_owner()
+    _add_access(user)
+    return ()
+end
+
+func _add_access{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(user : felt):
```
krebernisak: Why do you have an `_add_access` internal fn but no `_remove_access` internal fn? Let's be consistent here.

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
krebernisak: Namespace examples: https://github.com/OpenZeppelin/cairo-contracts/blob/main/src/openzeppelin/access/accesscontrol/library.cairo

@archseer please make a decision here. We should be consistent across our codebase.

@taco-paco these `s_` prefixes are c/p from our Solidity implementation, right?

contracts/contracts/SimpleReadAccessController/library.cairo
```diff
@@ -0,0 +1,57 @@
...
+        # it would be a risky strategy to depend on this part of infrastructure that can easily be upgraded.
+        let (tx_info) = get_tx_info()
+        if tx_info.account_contract_address == user:
+            return (TRUE)
+        end
+
+        return (FALSE)
+    end
+
+    # TODO: remove when starkware adds get_class_hash_at
```
krebernisak: How would we use `get_class_hash_at` here?

contracts/contracts/SimpleReadAccessController/library.cairo
```diff
@@ -0,0 +1,57 @@
...
+    func check_access{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
+        user : felt
+    ):
+        alloc_locals
+
+        let empty_data_len = 0
+        let (empty_data) = alloc()
```
krebernisak: We can't fetch `msg.data` for this call, as we do in Solidity, right? 
If not please add an in-line comment here, provide some more info to the reader.

contracts/contracts/SimpleReadAccessController/simple_read_access_controller.cairo
```diff
@@ -0,0 +1,33 @@
+%lang starknet
+
+from starkware.cairo.common.alloc import alloc
+from starkware.cairo.common.cairo_builtins import HashBuiltin, SignatureBuiltin
+from starkware.cairo.common.bool import TRUE, FALSE
+
+from ocr2.interfaces.IAccessController import IAccessController
+from SimpleReadAccessController.library import simple_read_access_controller
+from SimpleWriteAccessController.library import simple_write_access_controller
```
krebernisak: This import doesn't seem to be used in this specific file, is it required?

contracts/contracts/SimpleReadAccessController/library.cairo
```diff
@@ -0,0 +1,57 @@
...
+
+from starkware.cairo.common.alloc import alloc
+from starkware.cairo.common.cairo_builtins import HashBuiltin
+from starkware.starknet.common.syscalls import get_tx_info
+from starkware.cairo.common.bool import TRUE, FALSE
+
+from SimpleWriteAccessController.library import simple_write_access_controller
+
+namespace simple_read_access_controller:
+    func constructor{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
```
krebernisak: Didn't you mention that contracts with constructors are not composable, so we restructured them, but now this library also has a constructor. 

Asking questions here, not sure how to best structure this.

contracts/contracts/mocks/uptime_feed_mock.cairo
```diff
@@ -0,0 +1,55 @@
+%lang starknet
```
krebernisak: Why was this mock contract added? What is it used for?

The only usage I see is in a test that checks the interface of a mocked contract, why?

contracts/contracts/ocr2/aggregator.cairo
```diff
@@ -1020,10 +1023,14 @@ func has_billing_access{
...
+    let (empty_data) = alloc()
 
     let (has_access: felt) = IAccessController.has_access(
         contract_address=access_controller,
-        address=caller
+        user=caller,
+        data_len=empty_data_len,
+        data=empty_data
     )
     return (has_access)
```
krebernisak: 1. `IAccessController.check_access` is available and does exactly this, checks against an empty calldata.
2. `has_billing_access` returns a boolean, but is used across the `aggregator` contract as an access modifier. As the fn is `require_billing_access`, but it is not. It will not error, but just return true/false which is currently not handled.

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
```diff
@@ -0,0 +1,258 @@
...
+end
+
+@event
+func UpdateIgnored(
+    latest_status : felt, latest_timestamp : felt, incoming_status : felt, incoming_timestamp : felt
+):
+end
+
+@event
+func L1SenderTransferred(prev : felt, cur : felt):
```
krebernisak: There is a subtile API change introduced here: it should be to/from not prev/cur.

Context: https://github.com/smartcontractkit/chainlink/search?q=L1SenderTransferred

Please be careful to keep the original (or close to original) API where appropriate. Please double check the code for those inconsistencies. Thx!

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
```diff
@@ -0,0 +1,258 @@
...
+end
+
+func require_valid_round_id{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
+    round_id : felt
+):
+    let (lateset_round_id) = s_latest_round_id.read()
+
+    with_attr error_message("invalid round_id"):
+        assert_not_zero(round_id)
+        # TODO: do we need to check if uint80 is overflown?
```
aiagentdev-1990: is this still a todo?

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
```diff
@@ -0,0 +1,258 @@
...
+        assert l1_sender = address
+    end
+
+    return ()
+end
+
+func require_valid_round_id{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
+    round_id : felt
+):
+    let (lateset_round_id) = s_latest_round_id.read()
```
aiagentdev-1990: `lateset_round_id` should be `latest_round_id`

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
```diff
@@ -0,0 +1,258 @@
...
+        assert_boolean(status)
+
+        let (latest_round_id) = s_latest_round_id.read()
+        let (latest_observation_timestamp) = s_rounds.read(
+            latest_round_id, Round.observation_timestamp
+        )
+        let (local latest_status) = s_rounds.read(latest_round_id, Round.answer)
+
+        let (lt) = is_le(timestamp, latest_observation_timestamp - 1)  # timestamp < latest_observation_timestamp
+        if lt == TRUE:
```
aiagentdev-1990: can we just do `if lt` in Cairo instead of checking `== TRUE`?

aiagentdev-1990: `lt` is also not used anywhere else so it might be cleaner to do

```
if is_le(timestamp, latest_observation_timestamp - 1):
```

instead of assigning `lt` to a variable.

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
```diff
@@ -0,0 +1,258 @@
...
+        require_valid_round_id(round_id)
+
+        let (round) = _get_round(round_id)
+        return (round)
+    end
+
+    func description{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}() -> (
+        description : felt
+    ):
+        const description = 'L2 Sequencer Uptime Status Feed'
```
aiagentdev-1990: we can just do `return L2 Sequencer Uptime Status Feed` instead of declaring `description`

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
```diff
@@ -0,0 +1,258 @@
...
+        return (description)
+    end
+
+    func decimals{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}() -> (
+        decimals : felt
+    ):
+        const decimals = 0
```
aiagentdev-1990: same here.  Can we just do `return 0` or is this not possible with Cairo?

contracts/contracts/SimpleWriteAccessController/library.cairo
```diff
@@ -0,0 +1,136 @@
...
+
+    return ()
+end
+
+# makes the access check enforced
+@external
+func enable_access_check{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}():
+    Ownable_only_owner()
+    let (check_enabled) = s_check_enabled.read()
+    if check_enabled == FALSE:
```
aiagentdev-1990: do we need to explicitly do `== FALSE`?  Can we just do

```
if( !s_check_enabled.read())

```

contracts/contracts/SimpleWriteAccessController/library.cairo
```diff
@@ -0,0 +1,136 @@
...
+    if has_access == FALSE:
+        s_access_list.write(user, TRUE)
+        AddedAccess.emit(user)
+        return ()
+    end
+
+    return ()
+end
+
+# # Removes an address from the access list
```
aiagentdev-1990: There is an extra `#`

aiagentdev-1990 (changes requested):

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
taco-paco: Saw this somewhere in other cairo contracts. This just lets know that this is storage_var from name + solves issues with redefinition like
`let (latest_round_id) = s_latest_round_id.read()`

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
taco-paco: Had this question because couldn't find where those methods ore overriden in solidity contracts. So was wondering why they were declared as virtual

contracts/contracts/SimpleWriteAccessController/library.cairo
taco-paco: Zombie comment.

contracts/contracts/SimpleReadAccessController/library.cairo
taco-paco: Then we won't need to have a copy of `SimpleWriteAccessController::check_access` here in `SimpleReadAccessController`. In `SimpleWriteAccessController::check_access` we could fetch class_hash and use [library_call](https://www.cairo-lang.org/docs/hello_starknet/calling_contracts.html#library-calls) which would execute has_access in corresponding context: if class_hash returns hash of `SimpleReadAccessController` (or parent) then we will execute `SimpleReadAccessController::has_access` otherwise we will execute `SimpleWriteAccessController::has_access`

taco-paco: [last part & c++ example of 3 here](https://github.com/smartcontractkit/chainlink-starknet/pull/49). in `Base::other_f`

contracts/contracts/SimpleWriteAccessController/library.cairo
taco-paco: Removed `_add_access`

contracts/contracts/SimpleReadAccessController/library.cairo
taco-paco: Discussed on meeting. Probably need to create issue for this  once its merged

contracts/contracts/SimpleReadAccessController/simple_read_access_controller.cairo
taco-paco: Removed

contracts/contracts/SimpleReadAccessController/library.cairo
taco-paco: Theoretical preambule:
- For cairo to perceives func as constructor when it needs to have [`@constructor` decorator + func name is constructor](https://github.com/starkware-libs/cairo-lang/blob/167b28bcd940fd25ea3816204fa882a0b0a49603/src/starkware/starknet/compiler/external_wrapper.py#L212)
- Functions with decorators can't be used in `namespace`s
- `namespace`'s name prepends to functions name + [some magic static string values](https://github.com/starkware-libs/cairo-lang/blob/167b28bcd940fd25ea3816204fa882a0b0a49603/src/starkware/starknet/compiler/external_wrapper.py#L723)

That means that if file contains constructor (according to definition above) and then includes file that contains constructor we will have `redefinition` error because we can't move it to namespace or name not as `constructor`. So what we do to avoid this is restructure code to have `library.cairo` & `some_contract.cairo`. Then use namespace trickery to avoid naming functions smth like that `simple_read_access_controller_constructor` (inside `library.cairo`)

taco-paco: > Theoretical preambule:
> 
> * For cairo to perceives func as constructor when it needs to have [`@constructor` decorator + func name is constructor](https://github.com/starkware-libs/cairo-lang/blob/167b28bcd940fd25ea3816204fa882a0b0a49603/src/starkware/starknet/compiler/external_wrapper.py#L212)
> * Functions with decorators can't be used in `namespace`s
> * `namespace`'s name prepends to functions name + [some magic static string values](https://github.com/starkware-libs/cairo-lang/blob/167b28bcd940fd25ea3816204fa882a0b0a49603/src/starkware/starknet/compiler/external_wrapper.py#L723)
> 
> That means that if file contains constructor (according to definition above) and then includes file that contains constructor we will have `redefinition` error because we can't move it to namespace. So what we do to avoid this is restructure code to have `library.cairo` & `some_contract.cairo`. Then use namespace trickery to avoid naming functions smth like that `simple_read_access_controller_constructor`

contracts/contracts/mocks/uptime_feed_mock.cairo
taco-paco: To ensure work of interfaces and propagation of values through `libraries`. ie if deployed `SimpleReadAccessController` and calling `IAggregator::has_access` then `SimpleReadAccessController::has_access` will be called. Same with `SimpleWriteAccessController`

contracts/contracts/ocr2/aggregator.cairo
taco-paco: 1. `check_access` in current form panics if one doesn't have an access.
2. I think regarding implementation details of `aggregator.cairo` @archseer can answer better.

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
taco-paco: `from` is a [system token](https://github.com/starkware-libs/cairo-lang/blob/13cef109cd811474de114925ee61fd5ac84a25eb/src/starkware/cairo/lang/compiler/parser_errors_test.py#L173) and if used not in import(maybe other uses are possible but I've never seen one) results in such error:
`Unexpected token Token('FROM', 'from')`

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
taco-paco: `Solidity` implementation uses `uint80` and as I remember ensures that it doesn't overflow. I'm not sure how critical that's why a left this todo. So answer would depend on your requirements :)

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
taco-paco: Thanks! Fixed everywhere

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
taco-paco: No :) even this isn't possible
`if is_le(timestamp, latest_observation_timestamp - 1) == TRUE`
cairo 💪

contracts/contracts/SimpleWriteAccessController/library.cairo
taco-paco: No

contracts/contracts/SimpleWriteAccessController/library.cairo
taco-paco: Removed

contracts/contracts/solidity/StarkNetValidator.sol
```diff
@@ -0,0 +1,89 @@
...
+contract StarkNetValidator is TypeAndVersionInterface, AggregatorValidatorInterface, SimpleWriteAccessController {
+    int256 private constant ANSWER_SEQ_OFFLINE = 1;
+    // Selector hardcoded because StarkNet hash function is not available in this environment
+    uint256 constant STARK_SELECTOR_UPDATE_STATUS =
+        1585322027166395525705364165097050997465692350398750944680096081848180365267;
+
+    IStarknetMessaging public immutable STARKNET_MESSAGING;
+    uint256 public immutable L2_UPTIME_FEED_ADDR;
+
+    /// Invalid StarkNet messaging contract address - the address is 0.
```
aiagentdev-1990: Let's try use NatSpec comments here

```suggestion
    /// @notice Invalid StarkNet messaging contract address - the address is 0.
```

contracts/contracts/solidity/StarkNetValidator.sol
```diff
@@ -0,0 +1,89 @@
...
+    error InvalidStarkNetMessaging ();
+
+    /// Invalid StarkNet uptime feed address - the address is 0.
```
aiagentdev-1990: Natspec comment

contracts/contracts/solidity/StarkNetValidator.sol
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * @title StarkNetValidator - makes cross chain call to update the Sequencer Uptime Feed on L2
+ */
+contract StarkNetValidator is TypeAndVersionInterface, AggregatorValidatorInterface, SimpleWriteAccessController {
+    int256 private constant ANSWER_SEQ_OFFLINE = 1;
+    // Selector hardcoded because StarkNet hash function is not available in this environment
+    uint256 constant STARK_SELECTOR_UPDATE_STATUS =
+        1585322027166395525705364165097050997465692350398750944680096081848180365267;
+
+    IStarknetMessaging public immutable STARKNET_MESSAGING;
```
aiagentdev-1990: Can we rename `STARKNET_MESSAGING`?

```suggestion
    IStarknetMessaging public immutable STARKNET_CROSS_DOMAIN_MESSENGER;
```

contracts/contracts/solidity/StarkNetValidator.sol
```diff
@@ -0,0 +1,89 @@
...
+    function validate(
+        uint256, /* previousRoundId */
+        int256 previousAnswer,
+        uint256, /* currentRoundId */
+        int256 currentAnswer
+    ) external override checkAccess returns (bool) {
+        bool status = currentAnswer == ANSWER_SEQ_OFFLINE;
+        uint256[] memory payload = new uint256[](2);
+
+        // File payload with `status` and `timestamp`
```
aiagentdev-1990:
```suggestion
        // Fill payload with `status` and `timestamp`
```

contracts/test/starkNetValidator/StarkNetValidator.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    mockStarkNetMessenger = await mockStarkNetMessengerFactory.deploy()
+    await mockStarkNetMessenger.deployed()
+
+    starkNetValidator = await starknetValidatorFactory.deploy(mockStarkNetMessenger.address, l2Contract.address)
+
+    await account.invoke(l2Contract, 'set_l1_sender', { address: starkNetValidator.address })
+  })
+
+  describe('starknetValidator', () => {
+    it('should get the selector from name successfully', async () => {
```
aiagentdev-1990:
```suggestion
    it('should get the selector from the name successfully', async () => {
```

contracts/test/starkNetValidator/StarkNetValidator.test.ts
```diff
@@ -0,0 +1,85 @@
...
+      const setSelector = getSelectorFromName('update_status')
+      expect(BigInt(setSelector)).to.equal(
+        1585322027166395525705364165097050997465692350398750944680096081848180365267n,
+      )
+    })
+
+    it('reverts if called by account with no access', async () => {
```
aiagentdev-1990, line 54:
```suggestion
    it('reverts if called by an account with no access', async () => {
```

contracts/test/starkNetValidator/StarkNetValidator.test.ts
```diff
@@ -0,0 +1,85 @@
...
+      expect(deployedTo).not.to.be.undefined
+      expect(L1Provider).to.equal(networkUrl)
+
+      const { address: loadedFrom } = await starknet.devnet.loadL1MessagingContract(
+        networkUrl,
+        mockStarkNetMessenger.address,
+      )
+
+      expect(mockStarkNetMessenger.address).to.equal(loadedFrom)
+
```
aiagentdev-1990, line 72: remove extra lines

examples/contracts/aggregator-consumer/contracts/Price_Consumer.cairo
```diff
@@ -0,0 +1,84 @@
+%lang starknet
```
aiagentdev-1990: why is this file named `Price_Consumer.cairo`?

examples/contracts/aggregator-consumer/scripts/getLatestPrice.ts
```diff
@@ -0,0 +1,25 @@
+import { starknet } from 'hardhat'
```
aiagentdev-1990, line 1: why is this file named `getLatestPrice`?

examples/contracts/aggregator-consumer/scripts/consumerValidator.ts
```diff
@@ -0,0 +1,72 @@
...
+let validator: Contract
+let mockStarkNetMessengerFactory: ContractFactory
+let mockStarkNetMessenger: Contract
+let deployer: SignerWithAddress
+let eoaValidator: SignerWithAddress
+let networkUrl: string
+let account: Account
+
+async function main() {
+  // account = await starknet.getAccountFromAddress(
```
aiagentdev-1990: is this still used?  We should remove it if not.

contracts/test/ocr2/sequencer-uptime-feed.ts
```diff
@@ -0,0 +1,121 @@
...
+        user,
+      })
+
+      {
+        const res = await mockContract.call('has_access', { user: user, data: [] })
+        expect(res.bool).to.equal(1n)
+      }
+    })
+
+    // No l1_handler testing for now
```
aiagentdev-1990: let's just remove this for now since we don't have a way to do it

contracts/test/ocr2/sequencer-uptime-feed.ts
```diff
@@ -0,0 +1,121 @@
+import { expect } from 'chai'
+import { starknet } from 'hardhat'
+import { StarknetContract, StarknetContractFactory, Account } from 'hardhat/types/runtime'
+import { number } from 'starknet'
+import { assertErrorMsg } from '../utils'
+
+describe('SequencerUptimeFeed test', function () {
+  this.timeout(300_000)
```
aiagentdev-1990, line 8: why do we need this?

contracts/test/ocr2/sequencer-uptime-feed.ts
```diff
@@ -0,0 +1,121 @@
...
+        owner_address: number.toBN(account.starknetContract.address),
+      })
+    })
+
+    it('Test grainting access', async function () {
+      await account.invoke(uptimeFeedContract, 'add_access', {
+        user,
+      })
+
+      try {
```
aiagentdev-1990, line 36: not sure if we can do this for Starknet, but is there something similar to this?

```
await expect(nonOwner.invoke(uptimeFeedContract, 'add_access', {
  user,
}).to.be.revertedWith('Ownable: caller is not the owner')
```

contracts/test/ocr2/sequencer-uptime-feed.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    before(async function () {
+      const uptimeFeedFactory = await starknet.getContractFactory('sequencer_uptime_feed')
+      uptimeFeedContract = await uptimeFeedFactory.deploy({
+        initial_status: 0,
+        owner_address: number.toBN(account.starknetContract.address),
+      })
+    })
+
+    it('Test grainting access', async function () {
```
aiagentdev-1990, line 31: We should structure our tests where there is a describe block for each function and have describe blocks for each scenario.  This is an example.

```
describe('#add_access', () => {
  describe('when called by non owner', () => {
    it('reverts', async () => {
       // Add test here
    })
  })
})
```

contracts/test/ocr2/sequencer-uptime-feed.ts
```diff
@@ -0,0 +1,121 @@
...
+        })
+
+        expect.fail()
+      } catch (err: any) {
+        assertErrorMsg(err?.message, 'Ownable: caller is not the owner')
+      }
+    })
+
+    it('Test has_access', async function () {
+      {
```
aiagentdev-1990, line 48: This should be two separate tests.

```
describe('#has_access', () => {
  describe('when called by a user that has access', () => {
    it('returns true', async () => {
       // Add test here
    })
  })

 describe('when called by a user that has no access', () => {
    it('returns false', async () => {
       // Add test here
    })
  })
})
```

contracts/test/ocr2/sequencer-uptime-feed.ts
```diff
@@ -0,0 +1,121 @@
...
+        expect(res.bool).to.equal(1n)
+      }
+
+      {
+        const res = await uptimeFeedContract.call('has_access', { user: user + 1, data: [] })
+        expect(res.bool).to.equal(0n)
+      }
+    })
+
+    it('Test check_access', async function () {
```
aiagentdev-1990, line 59:
```
describe('#check_access', () => {
  describe('when called by an address that has no access', () => {
    it('reverts', async () => {
       // Add test here
    })
  })
})
```

contracts/contracts/cairo/ocr2/SequencerUptimeFeed/library.cairo
```diff
@@ -0,0 +1,258 @@
...
+        assert_boolean(status)
+
+        let (latest_round_id) = s_latest_round_id.read()
+        let (latest_observation_timestamp) = s_rounds.read(
+            latest_round_id, Round.observation_timestamp
+        )
+        let (local latest_status) = s_rounds.read(latest_round_id, Round.answer)
+
+        let (lt) = is_le(timestamp, latest_observation_timestamp - 1)  # timestamp < latest_observation_timestamp
+        if lt == TRUE:
```
aiagentdev-1990, line 122: can we not do `if is_le(timestamp, latest_observation_timestamp - 1)` in Cairo?

aiagentdev-1990 (changes requested):

contracts/test/starkNetValidator/StarkNetValidator.test.ts
taco-paco: done

contracts/contracts/solidity/StarkNetValidator.sol
taco-paco: Done

contracts/contracts/solidity/StarkNetValidator.sol
taco-paco: Done

contracts/contracts/solidity/StarkNetValidator.sol
taco-paco: Done

contracts/contracts/cairo/ocr2/SequencerUptimeFeed/library.cairo
aiagentdev-1990: Just saw comment above.

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
archseer: I think we should follow the OpenZeppelin guidelines: https://docs.openzeppelin.com/contracts-cairo/0.3.1/extensibility#libraries

> Must prefix storage functions with the name of the namespace to prevent clashing with other libraries (e.g. ERC20balances)

I agree that properly named variables are a good practice, but without namespacing subtle bugs can be introduced since it won't actually error, it'll just behave incorrectly.

contracts/contracts/solidity/StarkNetValidator.sol
taco-paco: Done

contracts/test/ocr2/sequencer-uptime-feed.ts
taco-paco: Let's create an extra PR for test refactorings

contracts/test/ocr2/sequencer-uptime-feed.ts
taco-paco: No

taco-paco:
Hey everyone 👋 ! Please resolve existing conversation so I would now that you're satisfied with answer and we can continue discussion in the ones that left. Thanks!

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
ivpavici: Discussed on the meeting at 18.8. and agreed? @archseer

contracts/contracts/SimpleWriteAccessController/library.cairo
ivpavici: @krebernisak , @archseer  resolved?

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
ivpavici: Question: do we turn this function into "virtual" in cairo?

contracts/contracts/SimpleReadAccessController/library.cairo
ivpavici: Shall we leave or remove this TODO ?

contracts/contracts/SimpleReadAccessController/library.cairo
ivpavici: @krebernisak this one can also be resolved I guess?

contracts/contracts/SimpleReadAccessController/library.cairo
ivpavici: What do we do with this one? :)

contracts/contracts/mocks/uptime_feed_mock.cairo
ivpavici: Remove this mock or leave it?

contracts/contracts/ocr2/aggregator.cairo
taco-paco: @archseer Shall we address this in this PR or create an issue for future fix?  (Discussed in dm on slack)

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
taco-paco: Used @archseer suggestion changed to: `from_addr`, `to_addr`

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
taco-paco: Discussed with @archseer
Leaving todo

ivpavici: @archseer please resolve if ok?

contracts/test/ocr2/sequencer-uptime-feed.ts
ivpavici: We'll add this in the same PR after all :)

contracts/test/ocr2/sequencer-uptime-feed.ts
taco-paco: So describe contains only one test every time? I thinks test file and log will turn into large matreshka. I would suggest having describes just for functions and then write scenarios in `it`

contracts/test/ocr2/sequencer-uptime-feed.ts
ivpavici: @qperrot

examples/contracts/aggregator-consumer/scripts/getLatestPrice.ts
ivpavici: @qperrot

examples/contracts/aggregator-consumer/contracts/Price_Consumer.cairo
ivpavici: @qperrot

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
ivpavici: Done

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
ivpavici: Done

contracts/test/starkNetValidator/StarkNetValidator.test.ts
ivpavici: Done I guess, please check

examples/contracts/aggregator-consumer/scripts/consumerValidator.ts
ivpavici: @qperrot , we uncommented it

contracts/test/ocr2/sequencer-uptime-feed.ts
ivpavici: removed

examples/contracts/aggregator-consumer/contracts/Price_Consumer.cairo
qperrot: I had to do the same implementation that you did here https://docs.chain.link/docs/l2-sequencer-flag/ so I keep the same names.

examples/contracts/aggregator-consumer/scripts/getLatestPrice.ts
qperrot: Same thing. But this is currently not a good implementation, I had to finish the ERC677 first, then I will go back on those exemple.

contracts/test/ocr2/sequencer-uptime-feed.ts
aiagentdev-1990: Sure that’s fine as long as we are descriptive about what scenario is being tested.

taco-paco: Seems like we state same thing twice:
```
describe('when called by a user that has access', () => {
    it('returns true', async () => {
```
In context of `has_access` function first and second line say the same thing.
Same for case with `false`

While that would be explicit enough in my opinion:
```
describe('#has_access', () => {
  it('when called by a user that has access', async () => {
```

contracts/contracts/SimpleReadAccessController/library.cairo
archseer: Cairo 0.9.0 added this so we should fix the TODO: start using `get_class_hash_at`

contracts/contracts/ocr2/aggregator.cairo
archseer: Fixed in the e2e PR

examples/contracts/aggregator-consumer/contracts/Price_Consumer.cairo
ivpavici: @cds95

examples/contracts/aggregator-consumer/scripts/getLatestPrice.ts
ivpavici: @cds95

contracts/test/ocr2/sequencer-uptime-feed.ts
ivpavici: @cds95 what will we conclude here? your decision! :)

contracts/test/ocr2/sequencer-uptime-feed.ts
ivpavici: pending decision from https://github.com/smartcontractkit/chainlink-starknet/pull/56#discussion_r951099148

contracts/contracts/SimpleReadAccessController/library.cairo
taco-paco: That's an API it's not present in syscalls as [get_caller_address](https://github.com/starkware-libs/cairo-lang/blob/13cef109cd811474de114925ee61fd5ac84a25eb/src/starkware/starknet/common/syscalls.cairo#L196) for example

taco-paco: So if we want to change it we would have to change interface adding extra argument for `class_hash`

contracts/test/ocr2/sequencer-uptime-feed.ts
aiagentdev-1990: I still prefer us to explicitly callout how the function should behave.  Saying `when called by a user that has access` doesn't really tell the reader what the expected behavior is.

contracts/test/ocr2/sequencer-uptime-feed.ts
aiagentdev-1990: Same case as above

contracts/test/ocr2/sequencer-uptime-feed.ts
ivpavici: @taco-paco will remove this one (after testing it)

examples/contracts/aggregator-consumer/contracts/Price_Consumer.cairo
aiagentdev-1990: ok then we should rename it to `PriceConsumerWithSequencerCheck` to match the example then.

qperrot: Done

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
archseer: I think we can just drop this TODO (change it to possibly make it overridable in the future since we don't need this right now)

contracts/contracts/SimpleReadAccessController/library.cairo
archseer: Let's keep the TODO then

contracts/contracts/SimpleReadAccessController/library.cairo
archseer: Rename to `initializer`

contracts/test/ocr2/sequencer-uptime-feed.ts
ivpavici: Will most likely be addressed in a separate PR after all, since tests are not audited and can change more easily :)

contracts/contracts/ocr2/SequencerUptimeFeed/library.cairo
taco-paco: Done

contracts/contracts/SimpleReadAccessController/library.cairo
taco-paco: Done. Renamed to initialize as suggested in slack dms

aiagentdev-1990:
Approved from my end assuming we have a separate PR after this to refactor tests.  I'd still like @krebernisak to give the final thumbs up on this though as he is more familiar with the Cairo code.

contracts/package.json
```diff
@@ -19,13 +21,15 @@
     "bn.js": "^5.2.0",
     "cairo-ls": "^0.0.4",
     "chai": "^4.3.4",
+    "ethereum-waffle": "^3.4.4",
```
archseer, line 24: Is this used anywhere?

archseer (dismissed):
Looks okay to me, but @krebernisak should review as well

contracts/package.json
krebernisak: I don't think so, removing it in subsequent review PR

krebernisak: Ah, it is required after all. It brings the Ethereum specific chai matchers. We use them to assert on errors in tests.

PR #57, opened 2022-07-26
implement reader interface, connecting components together

aalu1418:
corresponding PR in core: https://github.com/smartcontractkit/chainlink/pull/7079

merge this PR in `chainlink-starknet` first!


Makefile
```diff
@@ -40,7 +40,7 @@ ifeq ($(OSFLAG),$(LINUX))
 	curl -s https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | bash
 ifneq ($(CI),true)
 	# install nix
-	sh <(curl -L https://nixos-nix-install-tests.cachix.org/serve/vij683ly7sl95nnhb67bdjjfabclr85m/install) --daemon --tarball-url-prefix https://nixos-nix-install-tests.cachix.org/serve --nix-extra-conf-file ./nix.conf
+	# sh <(curl -L https://nixos-nix-install-tests.cachix.org/serve/vij683ly7sl95nnhb67bdjjfabclr85m/install) --daemon --tarball-url-prefix https://nixos-nix-install-tests.cachix.org/serve --nix-extra-conf-file ./nix.conf
```
aalu1418, line 43: @archseer do you know why this is failing (or if the logic to run it is correc)? I commented it out since it was blocking things - but also didn't seem to need it?

archseer: Yeah I'm not sure why this is here, maybe it was intended to be used on the CI

relayer/pkg/chainlink/relay.go
```diff
@@ -67,8 +67,11 @@ func (r *relayer) NewConfigProvider(args relaytypes.RelayArgs) (relaytypes.Confi
 		return nil, errors.Wrap(err, "couldn't initilize Chain")
 	}
 
-	url := "" // TODO: retrieve from reader from nodes/chains config
-	configProvider, err := ocr2.NewConfigProvider(relayConfig.ChainID, args.ContractID, url, chain.Config(), r.lggr)
+	reader, err := chain.Reader()
```
archseer, line 70: 👍🏻

Makefile
krebernisak: @tateexon originally added this in AFAIK

relayer/pkg/chainlink/ocr2/client.go
```diff
@@ -18,30 +18,24 @@ type OCR2Reader interface {
...
 
-	BaseClient() *starknet.Client
+	BaseClient() starknet.Reader
 }
 
 var _ OCR2Reader = (*Client)(nil)
 
 type Client struct {
-	starknetClient *starknet.Client
+	starknetClient starknet.Reader
```
krebernisak: Consider renaming `starknetClient` to maybe just `reader`

relayer/pkg/chainlink/relay.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/smartcontractkit/chainlink-starknet/relayer/pkg/chainlink/ocr2"
-	"github.com/smartcontractkit/chainlink-starknet/relayer/pkg/starknet"
+	chainset "github.com/smartcontractkit/chainlink-starknet/relayer/pkg/starknet/chain"
```
jmank88: I don't have a better suggestion, but `chainset` seems too specific as the whole packge alias.

aalu1418: `starkchain` ?

relayer/pkg/chainlink/ocr2/client.go
krebernisak: Here we have an OCR2 client using the StarkNet reader to read and write from contracts.

Let's clean these keywords - what's a reader vs. what's a client

I'd suggest a `Client is Reader, Writer`, a `Reader` is just a reader, while a `Writer` is just a writer. Let's structure it like that.

The original idea was to have:

- base `starknet.Client` abstracting network APIs (Juno vs Caigo, Gateway vs RPC)
- ocr2 specific `ocr2.Client` using the undelying `starknet.Client`

Now these `Reader` keywords were introduced and now we have code like:

```
type Client struct {
	starknetClient starknet.Reader
	lggr           logger.Logger
}

res, err := c.starknetClient.CallContract(ctx, ops)
```

Which is very confusing.

It should either be:

```go
type Client struct {
	r    starknet.Reader
	w    starknet.Writer
	lggr logger.Logger
}
```
or
```go
type Client struct {
	c    starknet.Client
	lggr logger.Logger
}
```

Please try to clean this up slightly in this PR 🙏

relayer/pkg/chainlink/ocr2/providers.go
```diff
@@ -28,8 +28,8 @@ type configProvider struct {
 	lggr logger.Logger
 }
 
-func NewConfigProvider(chainID string, contractAddress string, url string, cfg starknet.Config, lggr logger.Logger) (*configProvider, error) {
-	chainReader, err := NewClient(chainID, url, lggr, cfg)
+func NewConfigProvider(chainID string, contractAddress string, basereader starknet.Reader, cfg starknet.Config, lggr logger.Logger) (*configProvider, error) {
```
krebernisak: The `ConfigProvider` should depend just on `ocr2.Reader` as it just reads, doesn't write anything.

The `MedianProvider` on the other hand should depend on `ocr2.Client` which it uses for both reading and writing to the chain.

While these functions construct their own readers/clients (I'm lost here lol) now, ideally we do IoC and they depend on them, we provide/inject them.

relayer/pkg/chainlink/relay.go
krebernisak: `ChainSet` is a Chainlink concept, not a StarkNet concept.

IMHO the `ChainSet` type should be moved form `pkg/starknet` to `pkg/chainlink`.
I'd suggest we leave `client.go`, `client_test.go` and `types.go` in `pkg/starknet` and move everything else to `pkg/chainlink` or `pkg/chainlink/ocr2` depending on domain the code handles.

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -30,6 +29,14 @@ type StarkTXM interface {
...
 }
 
+//go:generate mockery --name TxConfig --output ./mocks/ --case=underscore --filename txconfig.go
+
+type TxConfig interface {
+	TxTimeout() time.Duration
+	TxSendFrequency() time.Duration
+	TxMaxBatchSize() int
+}
+
```
krebernisak: Consider extracting to `txm/config.go`.

Additionally, there is a [test_helpers.go](https://github.com/smartcontractkit/chainlink-starknet/tree/f4d2e6b4d47188fd5b9fa15fd377fc907910ece3/relayer/pkg/chainlink/txm)/test_helpers.go) file in this package which doesn't really belong here. It should be moved to `ops` or `integration-tests` module, potentially even split between the two.

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -38,14 +45,14 @@ type starktxm struct {
 	queue   chan types.Transaction
 	curve   *caigo.StarkCurve
 	ks      keys.Keystore
-	cfg     starknet.Config
+	cfg     TxConfig
 
 	// TODO: use lazy loaded client
 	client    types.Provider
-	getClient func() types.Provider
+	getClient func() (types.Provider, error)
```
krebernisak, line 44: Hmm, why do we again provide client like this (lazy fetch) instead of injecting it as a dependency to the txm constructor?

relayer/pkg/chainlink/ocr2/providers.go
aalu1418: why does the `MedianProvider` need the `writer` component? afaik the only part that writes to the chain is the txm

relayer/pkg/chainlink/txm/txm.go
aalu1418: so if there are multiple nodes configured, and one endpoint goes down the txm can try picking a new one

Makefile
aalu1418: i'll leave this commented out for now as it doesn't seem to be used - it can be revisited later on if needed

relayer/pkg/chainlink/ocr2/providers.go
krebernisak: The `MedianProvider` provides the `ContractTransmitter` that depends on TXM.

So the `MedianProvider` either requests TXM and uses it to construct the `ContractTransmitter`, or it requests the client and uses it to construct the TXM, and in turn the `ContractTransmitter`.

As currently the TXM is provided externally by the Chain type, we'll probably inject the TXM here. In that case, it doesn't need a `writter` but just an additional `reader`.

Does that make sense?

aalu1418: yup makes sense! I also realized that I forgot to connect the txm and and `ContractTransmitter` so I will add that implementation and then that should make more sense.

aalu1418: ok! now `Chain.TxManager` is now used to pass the txm to `MedianProvider` => `ContractTransmitter`

the txmanager is started together with the `Chain` in core

relayer/pkg/chainlink/ocr2/providers.go
```diff
@@ -77,14 +79,14 @@ type medianProvider struct {
...
 
-func NewMedianProvider(chainID string, contractAddress string, senderAddress string, url string, cfg starknet.Config, lggr logger.Logger) (*medianProvider, error) {
-	configProvider, err := NewConfigProvider(chainID, contractAddress, url, cfg, lggr)
+func NewMedianProvider(chainID string, contractAddress string, senderAddress string, basereader starknet.Reader, cfg config.Config, txm txm.TxManager, lggr logger.Logger) (*medianProvider, error) {
+	configProvider, err := NewConfigProvider(chainID, contractAddress, basereader, cfg, lggr)
 	if err != nil {
 		return nil, errors.Wrap(err, "couldn't initialize ConfigProvider")
 	}
 
 	cache := NewTransmissionsCache(cfg, configProvider.reader, configProvider.lggr)
```
krebernisak, line 88: It's confusing to use the ConfigProvider just as a dependency provider to construct the MedianProvider.

Consider constructing the dependencies inline without using the `NewConfigProvider` here.

krebernisak: `errors.Wrap(err, "couldn't initialize ConfigProvider")` is a weird error to get when constructing the `MedianProvider`

relayer/pkg/chainlink/ocr2/transmissions_cache.go
```diff
@@ -27,11 +26,11 @@ type transmissionsCache struct {
 	stop, done chan struct{}
 
 	reader *contractReader
-	cfg    starknet.Config
+	cfg    config.Config
```
krebernisak, line 29: Nesting many packages we get types that read like `config.Config` which is not ideal. We should iterate on how to structure this better.

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -101,7 +100,11 @@ func (txm *starktxm) run() {
 
 			// fetch client if needed
 			if txm.client == nil {
-				txm.client = txm.getClient()
+				var err error
+				txm.client, err = txm.getClient()
+				if err != nil {
+					txm.lggr.Errorw("unable to fetch client", "error", err)
```
jmank88, line 106: Should this `continue` the for loop to try again on next tick, rather than moving on with `txm.client == nil`?

krebernisak, line 106: How do we plan to recover from this?

krebernisak (approved):

relayer/pkg/chainlink/txm/txm.go
krebernisak: Deleted my other comment to group it with a similar one here.

Basically, "How do we plan to recover from this?" is the Q.

aalu1418: yeah there should be a `continue` with that...I'll fix that in a follow up PR with the DB fix

relayer/pkg/chainlink/ocr2/providers.go
aalu1418: yeah it's a bit confusing - afaik ConfigProvider is used for bootstrap nodes, and MedianProvider is used for oracle nodes. so the ConfigProvider is still needed for the oracle to func - hence the dependency.

at the very least this could be cleaned up a bit

krebernisak: There is no dependency on the `ConfigProvider` per se, we don't use the things it actually provides `ContractConfigTracker` or `OffchainConfigDigester`, but just internal deps the `ConfigProvider` constructs.

I'm suggesting it's cleaner to construct them ourselves in the `NewMedianProvider` or have a separate dependency provider that both can use to construct their dependencies (if the idea is to share code).

relayer/pkg/chainlink/ocr2/transmissions_cache.go
aalu1418: this one is tricky since moving it out of it's own module was causing circular dependencies, but I think I might be able to set up a cleaner way - similar to how the txm config is contained to the txm module. the ocr2 cache configs could also be defined in the ocr2 module

relayer/pkg/chainlink/ocr2/providers.go
aalu1418: ohhh interesting...did a little more digging and I understand what you are saying now. the way the caches are set up means we don't actually require most of ConfigProvider (different than solana and terra)

aalu1418: wait...on second thought: it does require the config provider dependencies? https://github.com/smartcontractkit/chainlink-relay/blob/24530b163a6f03e71846e669b2c219d7a0f52ede/pkg/types/types.go#L55

PR #58, opened 2022-07-26
export ocr2 keys

aalu1418:
forgot to export it when bringing over the files from core

### DIFF
--- a/relayer/pkg/chainlink/keys/ocr2key.go
+++ b/relayer/pkg/chainlink/keys/ocr2key.go
@@ -118,7 +118,7 @@ func (sk *OCR2Keyring) MaxSignatureLength() int {
 	return 32 + 32 + 32 // publickey + r + s
 }
 
-func (sk *OCR2Keyring) marshal() ([]byte, error) {
+func (sk *OCR2Keyring) Marshal() ([]byte, error) {
 	// https://github.com/ethereum/go-ethereum/blob/07508ac0e9695df347b9dd00d418c25151fbb213/crypto/crypto.go#L159
 	return PadBytes(sk.privateKey.D, sk.privateKeyLen()), nil
 }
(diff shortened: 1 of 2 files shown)


relayer/pkg/chainlink/keys/ocr2key.go
```diff
@@ -16,25 +16,25 @@ import (
...
-var _ ocrtypes.OnchainKeyring = &ocr2Keyring{}
+var _ ocrtypes.OnchainKeyring = &OCR2Keyring{}
 
-type ocr2Keyring struct {
+type OCR2Keyring struct {
 	privateKey starksig.PrivateKey
 }
 
-func newStarkNetKeyring(material io.Reader) (*ocr2Keyring, error) {
+func NewStarkNetKeyring(material io.Reader) (*OCR2Keyring, error) {
```
jmank88, line 25: nbd but curious if you considered returning the already public interface instead?
```suggestion
func NewStarkNetKeyring(material io.Reader) (ocrtypes.OnchainKeyring, error) {
```

jmank88 (dismissed):

relayer/pkg/chainlink/keys/ocr2key.go
jmank88: I know there are differing opinions on when you should return the underlying type vs. the interface, but FWIW this can be error prone by accidentally leading to a typed nil (`(*OCR2Keyring)(nil)`) being set in a `type ocrtypes.OnchainKeyring` var, which passes nil checks unexpectedly.

aalu1418: that's fair! at the moment, the `OCR2Keyring` also needs to export marshal and unmarshal functions which are not part of `ocrtypes.OnchainKeyring` interface

another option could be to wrap the `OnchainKeyring` interface?

jmank88: Yeah, sounds like that would work. Really nbd though :shrug:

jmank88 (approved):

PR #61, opened 2022-07-29
bug fixes

aalu1418:
dependency for: https://github.com/smartcontractkit/chainlink/pull/7079

+ addresses miscellaneous feedback from other PRs
+ fixes the chainID in db struct


relayer/pkg/chainlink/config/config.go
```diff
@@ -35,8 +36,7 @@ type Config interface {
 	txm.TxConfig // txm config
```
krebernisak: nit:
```suggestion
	txm.Config // txm config
```

relayer/pkg/chainlink/keys/key.go
```diff
@@ -11,101 +11,101 @@ import (
...
+func (key Key) ID() string {
+	return key.ContractAddressStr()
 }
 
-// Not actually public key, this is the derived contract address
+// this is the derived contract address, the contract is deployed using the StarkKeyStr
 // This is the primary identifier for onchain interactions
 // the private key is identified by this
-func (key StarkKey) PublicKeyStr() string {
+func (key Key) ContractAddressStr() string {
```
krebernisak: Let's call this an account:
```suggestion
func (key Key) AccountAddressStr() string {
```

relayer/pkg/chainlink/keys/ocr2key.go
```diff
@@ -16,25 +16,25 @@ import (
 	ocrtypes "github.com/smartcontractkit/libocr/offchainreporting2/types"
 )
 
-var _ ocrtypes.OnchainKeyring = &OCR2Keyring{}
+var _ ocrtypes.OnchainKeyring = &OCR2Key{}
 
-type OCR2Keyring struct {
+type OCR2Key struct {
```
krebernisak, line 21: Do we need this public?

relayer/pkg/chainlink/ocr2/contract_cache.go
```diff
@@ -31,11 +30,11 @@ type contractCache struct {
 	stop, done chan struct{}
 
 	reader *contractReader
-	cfg    config.Config
+	cfg    Config
 	lggr   logger.Logger
 }
 
-func NewContractCache(cfg config.Config, reader *contractReader, lggr logger.Logger) *contractCache {
+func NewContractCache(cfg Config, reader *contractReader, lggr logger.Logger) *contractCache {
```
krebernisak: Weird to depend on a local struct here `reader *contractReader` instead of a proper interface. We should probably define an `ocr2.Reader` interface, the `ContractCache` should require it to read, cache the data and also implement so it can be used as an `ocr2.Reader`.

krebernisak (dismissed):
Looking better and better 🙌

relayer/pkg/chainlink/keys/ocr2key.go
aalu1418: yeah we do. it's used as part of defining generics - so we need the type to properly set up the generic struct

relayer/pkg/chainlink/ocr2/contract_cache.go
aalu1418: nice and clean fix:
```
type Reader interface {
	types.ContractConfigTracker
	median.MedianContract
}

var _ Reader = (*contractReader)(nil)
```

relayer/pkg/chainlink/txm/txm_test.go
```diff
@@ -75,6 +89,7 @@ func TestTxm(t *testing.T) {
 		}
 	}
 	time.Sleep(30 * time.Second)
+	wg.Wait()
```
jmank88, line 92: nit/ why sleep 30s if we are just going to `Wait()` anyways? doesn't that just artificially extend the length of the test? Can we drop the time.Sleep?

relayer/pkg/chainlink/keys/utils.go
```diff
@@ -28,7 +28,7 @@ func PubKeyToContract(pubkey starksig.PublicKey, classHash, salt *big.Int) []byt
 		pedersen.ArrayDigest(pubkey.X),
 	)
 
-	// pad big.Int to 32 bytes if needed
+	// pad big.Int to 32 bytes if needed:w
```
jmank88:
```suggestion
	// pad big.Int to 32 bytes if needed
```

aalu1418: ah dang it...still occasionally forgetting to hit `esc` before exiting terminal based editors lol

relayer/pkg/chainlink/txm/txm_test.go
aalu1418: if the `time.Sleep` is dropped then the txm would be closed right after the `wait` is fulfilled...but there are still txs in the queue at that point. so there would have to be a sleep afterward to make sure the the txm is done.

it's not ideal, but that `wait()` should immediately pass once the sleep time is done

jmank88: Ah, OK, two different things to wait on :up:

jmank88 (approved):

PR #68, opened 2022-08-05
Minor tooling tweaks

archseer:


relayer/pkg/chainlink/ocr2/utils.go
```diff
@@ -88,16 +88,8 @@ func isEventFromContract(event *caigotypes.Event, address string, eventName stri
...
 	eventKey := caigo.GetSelectorFromName(eventName)
-	for _, key := range event.Keys {
-		if key.Cmp(eventKey) == 0 {
-			isSameEventSelector = true
-			break
-		}
-	}
-
-	return isSameEventSelector
+	return event.Keys[0].Cmp(eventKey) == 0
```
archseer, line 94: It's guaranteed to always be the first key: https://github.com/Shard-Labs/starknet-hardhat-plugin/blob/e6a811220233b4dde7f68e6f3b90516186ed0e22/src/types.ts#L734-L735

krebernisak, line 94: If the line stays after debugging, let's add a comment:

```suggestion
        // encoded event name guaranteed to be at index 0
	return event.Keys[0].Cmp(eventKey) == 0
```

shell.nix
```diff
@@ -18,6 +18,11 @@
     delve
     golangci-lint
     gotools
+
+    kube3d
+    kubectl
+    k9s
+    kubernetes-helm
```
krebernisak, line 25: The kubernetes toolchain should also be added to asdf's .tool-versions

Extracting it as a GH Issue for someone else to take over.

krebernisak: https://github.com/smartcontractkit/chainlink-starknet/issues/70

krebernisak (approved):

PR #69, opened 2022-08-08
port utils.MustRandomPort back to test_helpers + split ops folder

aalu1418:
**Depends on:**
- [x] https://github.com/smartcontractkit/chainlink-relay/pull/67


krebernisak (dismissed):

integration-tests/go.mod
```diff
@@ -197,3 +197,5 @@ require (
 replace github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 
 replace github.com/smartcontractkit/chainlink-starknet/relayer => ../../chainlink-starknet/relayer
+
+replace github.com/smartcontractkit/chainlink-starknet/ops => ../ops
```
krebernisak: We should remove this from here and use `go.work` for replacing dependencies across this workspace like this. Consider adding and committing the `go.work` with commented out local replace statements.

relayer/go.mod
```diff
@@ -196,3 +197,5 @@ require (
 )
 
 replace github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
+
+replace github.com/smartcontractkit/chainlink-starknet/ops => ../ops
```
krebernisak: Same thing here, we should avoid replaces in `go.mod` to avoid accidentally building unwanted code/versions. Add a `go.work` and add those replace statements in, but comment them out to be used only during development.

krebernisak: Also add the `go.work` to `.gitignore` once you committed it

integration-tests/go.mod
aalu1418: 🤔 I understand the reason for not wanting a `replace` in a `go.mod` file
but the alternative is to have each package import the other - and this requires the commit hashes (and for them to exist in the long run), but since we squash commits, it might have to follow a multi-step process? (since squashed/rebased commits I think can get garbage-collected?)
* in the initial PR, point `integration-tests` and `relayer` to a PR commit hash of `ops`
* squash merge PR
* open new PR pointing `integration-tests` and `relayer` to a commit hash in `develop`

krebernisak: Yeah, squashing and rebasing becomes a problem with Go workspaces, there is no way around it. But we also can't keep `replace` statements in the public repo, as it's a footgun waiting to go off.

So maybe we just move forward by disabling Github automated squash/rebase-merges? We can still do them manually and `git push -f`, but the final merge should be made by Github after a green CI - meaning the pointers are correct.

@jmank88 any ideas? :)

jmank88: :thinking:

aalu1418:
note: gauntlet + e2e test won't be passing for now - that is fixed in a different PR

jmank88 (approved):

PR #72, opened 2022-08-09
initial localenv cli

aalu1418:


integration-tests/common/test_common.go
```diff
@@ -98,6 +105,20 @@ func (t *Test) DeployCluster(nodes int) {
...
+	}
+
+	// override image parameters if necessary
+	if useLocalImg {
+		imgParams = map[string]interface{}{
+			"image":   "k3d-registry.local:12345/chainlink",
+			"version": "local",
+		}
+
+	}
```
krebernisak: Ideally we can just pass in both the image and version as flags to our test run like so: `ginkgo -r --focus @ocr integration-tests/smoke -- --chainlink-image k3d-registry.local:12345/chainlink --chainlink-version local`.

We could also have presets like so: `ginkgo -r --focus @ocr integration-tests/smoke -- --chainlink [default/default-local/latest]`

krebernisak: This is not a request, just thinking out loud :) Great job!

krebernisak (approved):

relayer/ops/localenv/main.go
```diff
@@ -0,0 +1,95 @@
...
+	// delete removes the k8s cluster
+	case "delete":
+		run("remove k8s cluster", "k3d", "cluster", "delete", "local")
+		run("remove registry", "k3d", "registry", "delete", "k3d-registry.local")
+	default:
+		panic("unrecognized command")
+	}
+}
+
+func run(name string, f string, args ...string) {
```
krebernisak, line 55: Let's move this to commons somewhere: https://github.com/smartcontractkit/chainlink-starknet/pull/41#discussion_r935711863

aalu1418: I actually think this whole file could be extracted to potentially the `relay/ops` folder since there's nothing specific to starknet at the moment

I don't know if it needs to be extracted yet, but I will leave a note about the whole file

PR #75, opened 2022-08-10
various fixes while debugging e2e env

aalu1418:


relayer/pkg/starknet/utils.go
```diff
@@ -0,0 +1,16 @@
+package starknet
+
+import "math/big"
+
+func BigIntPadBytes(a *big.Int, length int) []byte {
```
krebernisak: nit: naming

```suggestion
func PadBytesBigInt(a *big.Int, length int) []byte {
```

krebernisak: Additionally, please add 1-liner docs to these functions.

P.S. This fn is so small, not sure it requires a helper.

krebernisak (dismissed):

krebernisak (approved):

PR #77, opened 2022-08-14
Feature/ledger support

sdrug:
Notes:
- Corresponding Ledger app crashes on Ledger-S+ (upstream issue with their binary). Current PR was tested only for Ledger-S.
- Signing mode on the ledger app must be set to "insecure"
- Tx is signed twice: once for estimating fee, once for execution


packages-ts/starknet-gauntlet/src/provider/index.ts
```diff
@@ -85,9 +85,8 @@ class Provider implements IStarknetProvider {
 
   signAndSend = async (accountAddress: string, wallet: IStarknetWallet, calls: Call[], wait = false) => {
     const account = new Account(this.provider, accountAddress, wallet.wallet)
-
-    const maxFee = await account.estimateFee(calls)
```
sdrug, line 105: `estimateFee` is called from `execute` down below

packages-ts/starknet-gauntlet/src/dependencies/index.ts
```diff
@@ -6,15 +6,16 @@ export interface Env {
...
 }
 
 export interface Dependencies {
   logger: typeof logger
   prompt: typeof prompt
   makeProvider: (url: string) => IStarknetProvider
-  makeWallet: (pk: string, account?: string) => IStarknetWallet
-  makeEnv: (flags: Record<string, string>) => Env
+  makeWallet: (withLedger: boolean, pk?: string, account?: string) => Promise<IStarknetWallet>
+  makeEnv: (flags: Record<string, any>) => Env
```
RodrigoAD: Let's try to avoid `any`
```suggestion
  makeEnv: (flags: Record<string, string | boolean>) => Env
```

packages-ts/starknet-gauntlet/src/dependencies/index.ts
```diff
@@ -6,15 +6,16 @@ export interface Env {
...
+  [key: string]: any // Custom env
 }
 
 export interface Dependencies {
   logger: typeof logger
   prompt: typeof prompt
   makeProvider: (url: string) => IStarknetProvider
-  makeWallet: (pk: string, account?: string) => IStarknetWallet
-  makeEnv: (flags: Record<string, string>) => Env
+  makeWallet: (withLedger: boolean, pk?: string, account?: string) => Promise<IStarknetWallet>
```
RodrigoAD: We should accept `path` as parameter

packages-ts/starknet-gauntlet-cli/src/index.ts
```diff
@@ -48,6 +48,7 @@ const registerExecuteCommand = <UI, CI>(
         billingAccessController: process.env.BILLING_ACCESS_CONTROLLER,
         link: process.env.LINK,
         secret: flags.secret || process.env.SECRET,
+        withLedger: !!flags.withLedger,
```
RodrigoAD: Let's include `ledgerPath` here. If undefined, it will fall back to the default.

RodrigoAD: Also, `withLedger` could come defined by `process.env.WITH_LEDGER`

packages-ts/starknet-gauntlet/src/wallet/ledgerWallet.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  static create = async () => {
+    const transport = await TransportNodeHid.create()
+    const ledgerConnector = new Stark(transport)
+    return new LedgerSigner(ledgerConnector)
+  }
+
+  async getPubKey() {
+    const response = await this.ledgerConnector.getPubKey(PATH)
```
RodrigoAD: Let's move this into the `create` fn. We can then store the publicKey in the state, so we don't need to access the ledger every time we want to get the public key

packages-ts/starknet-gauntlet/src/wallet/ledgerWallet.ts
```diff
@@ -0,0 +1,94 @@
...
+}
+
+class LedgerSigner implements SignerInterface {
+  protected ledgerConnector: Stark
+
+  private constructor(lc: Stark) {
+    this.ledgerConnector = lc
+  }
+
+  static create = async () => {
```
RodrigoAD: `path` should be accepted here. If undefined, it will use the default
```suggestion
  static create = async (path?: string) => {
```

packages-ts/starknet-gauntlet/src/wallet/ledgerWallet.ts
```diff
@@ -0,0 +1,94 @@
...
+import { Abi, Invocation, InvocationsSignerDetails } from 'starknet/types'
+import { TypedData, getMessageHash } from 'starknet/utils/typedData'
+import { fromCallsToExecuteCalldataWithNonce } from 'starknet/utils/transaction'
+import { calculcateTransactionHash, getSelectorFromName } from 'starknet/utils/hash'
+import Stark from '@ledgerhq/hw-app-starknet'
+import { LedgerError } from '@ledgerhq/hw-app-starknet'
+import TransportNodeHid from '@ledgerhq/hw-transport-node-hid'
+
+// EIP-2645 path
+const PATH = "m/2645'/579218131'/0'/0'"
```
RodrigoAD:
```suggestion
const DEFAULT_PATH = "m/2645'/579218131'/0'/0'"
```

packages-ts/starknet-gauntlet/src/wallet/ledgerWallet.ts
```diff
@@ -0,0 +1,94 @@
+import { IStarknetWallet } from './'
+import { SignerInterface, encode } from 'starknet'
+import { Abi, Invocation, InvocationsSignerDetails } from 'starknet/types'
+import { TypedData, getMessageHash } from 'starknet/utils/typedData'
+import { fromCallsToExecuteCalldataWithNonce } from 'starknet/utils/transaction'
+import { calculcateTransactionHash, getSelectorFromName } from 'starknet/utils/hash'
+import Stark from '@ledgerhq/hw-app-starknet'
+import { LedgerError } from '@ledgerhq/hw-app-starknet'
```
RodrigoAD:
```suggestion
import Stark, { LedgerError }  from '@ledgerhq/hw-app-starknet'
```

packages-ts/starknet-gauntlet/src/wallet/ledgerWallet.ts
```diff
@@ -0,0 +1,94 @@
...
+import TransportNodeHid from '@ledgerhq/hw-transport-node-hid'
+
+// EIP-2645 path
+const PATH = "m/2645'/579218131'/0'/0'"
+
+function toHexString(byteArray: Uint8Array): string {
```
RodrigoAD: Doesn't `starknet.js` or `@ledgerhq/hw-app-starknet` expose this util somewhere?

RodrigoAD (commented):
Looks good, some small comments

packages-ts/starknet-gauntlet/src/wallet/ledgerWallet.ts
krebernisak: @sdrug we should predefine the `application` here: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2645.md

> application - the 31 lowest bits of sha256 of the application name. Serve as a domain separator between different applications. In the context of DeversiFi in June 2020, it is the 31 lowest bits of sha256(starkexdvf) and the value would be 1393043894.

1. I suggest we use "chainlink" as the application string
2. We should check that every path (user provided) starts with the `m/2645'/579218131'/`

packages-ts/starknet-gauntlet/src/wallet/ledgerWallet.ts
```diff
@@ -0,0 +1,94 @@
...
+    this.wallet = signer
+    this.account = account
+  }
+
+  static create = async (account?: string) => {
+    const ledgerSigner: LedgerSigner = await LedgerSigner.create()
+    return new Wallet(ledgerSigner, account)
+  }
+
+  sign = () => {}
```
krebernisak: Why do we have this `sign` method if it's always `$noop`?

RodrigoAD: It's required by the interface I think

RodrigoAD:
Something is wrong with the new dependencies. Make sure `yarn.lock` is generated correctly

sdrug:
@krebernisak @RodrigoAD 
Your comments were addressed, pipeline passed successfully.
It's good to merge imo.

RodrigoAD:
The `env` dependencies in the tests need some adjustments. You'll probably need to force `flags.pk` and `flags.account` to be `as string`: https://github.com/smartcontractkit/chainlink-starknet/blob/feature/ledger-support/packages-ts/starknet-gauntlet/test/utils/index.ts#L51
The rest LGTM

packages-ts/starknet-gauntlet/src/dependencies/index.ts
```diff
@@ -6,15 +6,23 @@ export interface Env {
...
   makeProvider: (url: string) => IStarknetProvider
-  makeWallet: (pk: string, account?: string) => IStarknetWallet
-  makeEnv: (flags: Record<string, string>) => Env
+  makeWallet: (
+    withLedger: boolean,
+    ledgerPath: string,
+    pk?: string,
+    account?: string,
+  ) => Promise<IStarknetWallet>
+  makeEnv: (flags: Record<string, string | boolean>) => Env
```
krebernisak, line 19: Like @RodrigoAD suggests, there seems to be some problems with these types. Gauntlet tests fail, check here: https://github.com/smartcontractkit/chainlink-starknet/runs/8039855018?check_suite_focus=true

packages-ts/starknet-gauntlet/src/dependencies/index.ts
```diff
@@ -6,15 +6,23 @@ export interface Env {
...
+  [key: string]: string | boolean // Custom env
 }
 
 export interface Dependencies {
   logger: typeof logger
   prompt: typeof prompt
   makeProvider: (url: string) => IStarknetProvider
-  makeWallet: (pk: string, account?: string) => IStarknetWallet
-  makeEnv: (flags: Record<string, string>) => Env
+  makeWallet: (
```
krebernisak: The idea here was to define a general framework that supports extending with custom wallet implementations. So a framework should provide a simple hook that returns a general `Promise<IStarknetWallet>` and you as a user implement your hook.

But you didn't do that here. You actually extended the framework to now include the knowledge about your specific Ledger wallet. What if we wanted to add Trezor support, how would that look like?

packages-ts/starknet-gauntlet/src/wallet/index.ts
```diff
@@ -1,34 +1,25 @@
...
-}
-
-class Wallet implements IStarknetWallet {
-  wallet: Signer
-  account: string
-
-  private constructor(keypair: KeyPair, account?: string) {
-    this.wallet = new Signer(keypair)
-    this.account = account
+export const makeWallet = async (
```
krebernisak: This is something that needs to happen at the end of the road, in the place we hook everything together. It currently happens at the beginning, at our core `starknet-gauntlet` package which now has a dependency to `starknet-gauntlet-ledger` which ideally it shouldn't know anything about.

The `starknet-gauntlet-cli` should be the package where it all happens. That package should import all other packages (plugins) and hook them in. It should depend on and import the Ledger wallet plugin, then proceed to hook it in by implementing a hook similar to below.

packages-ts/starknet-gauntlet/src/wallet/ledgerWallet.ts
```diff
@@ -0,0 +1,117 @@
...
+import { fromCallsToExecuteCalldataWithNonce } from 'starknet/utils/transaction'
+import { calculcateTransactionHash, getSelectorFromName } from 'starknet/utils/hash'
+import Stark, { LedgerError } from '@ledgerhq/hw-app-starknet'
+
+// EIP-2645 path
+//  2645 - EIP number
+//  579218131 - layer - 31 lowest bits of sha256("starkex")
+//  894929996 - application - 31 lowest bits of sha256("chainlink")
+const DEFAULT_PATH = "m/2645'/579218131'/894929996'/0'"
+const PATH_REGEX = /^\s*m\s*\/\s*2645\s*\'\s*\/\s*579218131\s*\'\s*\/\s*(\d+)\s*\'\s*\/\s*(\d+)\s*\'$/
```
krebernisak: Let's please add a test or two for this regex.

shell.nix
```diff
@@ -23,6 +23,9 @@
     kubectl
     k9s
     kubernetes-helm
+
```
krebernisak, line 26: Please add a comment here explaining this requirement

packages-ts/starknet-gauntlet/src/wallet/ledgerWallet.ts
```diff
@@ -0,0 +1,117 @@
...
+    const signer = new LedgerSigner(ledgerConnector, path, publicKey)
+
+    return signer
+  }
+
+  async getPubKey() {
+    return this.publicKey
+  }
+
+  async signTransaction(
```
krebernisak: Please add return types to signatures, and some comments would also be nice: https://github.com/0xs34n/starknet.js/blob/develop/src/signer/interface.ts

packages-ts/starknet-gauntlet/src/wallet/index.ts
```diff
@@ -1,34 +1,25 @@
...
+
 export interface IWallet<W> {
   wallet: W
-  sign: (message: any) => any
   getPublicKey: () => Promise<string>
 }
 
-export interface IStarknetWallet extends IWallet<Signer> {
+export interface IStarknetWallet extends IWallet<SignerInterface> {
   getAccountPublicKey: () => string
```
krebernisak: The public key is one thing, an account address is another:

```suggestion
  getAccountAddress: () => string
```

packages-ts/starknet-gauntlet/src/wallet/ledgerWallet.ts
```diff
@@ -0,0 +1,117 @@
...
+const DEFAULT_PATH = "m/2645'/579218131'/894929996'/0'"
+const PATH_REGEX = /^\s*m\s*\/\s*2645\s*\'\s*\/\s*579218131\s*\'\s*\/\s*(\d+)\s*\'\s*\/\s*(\d+)\s*\'$/
+
+class LedgerSigner implements SignerInterface {
+  ledgerConnector: Stark
+  path: string
+  publicKey: string
+
+  private constructor(lc: Stark, path: string, publicKey: string) {
+    this.ledgerConnector = lc
```
krebernisak: Let's just call this a client:

```suggestion
  private constructor(client: Stark, path: string, publicKey: string) {
    this.client = client
```

packages-ts/starknet-gauntlet/src/wallet/ledgerWallet.ts
```diff
@@ -0,0 +1,117 @@
...
+import { TypedData, getMessageHash } from 'starknet/utils/typedData'
+import { fromCallsToExecuteCalldataWithNonce } from 'starknet/utils/transaction'
+import { calculcateTransactionHash, getSelectorFromName } from 'starknet/utils/hash'
+import Stark, { LedgerError } from '@ledgerhq/hw-app-starknet'
+
+// EIP-2645 path
+//  2645 - EIP number
+//  579218131 - layer - 31 lowest bits of sha256("starkex")
+//  894929996 - application - 31 lowest bits of sha256("chainlink")
+const DEFAULT_PATH = "m/2645'/579218131'/894929996'/0'"
```
krebernisak: Additionally, it could be useful to enable users to just add a flag like `--path-index=3'` and we just use the `DEFAULT_PATH` with the specified index. It would be less error prone than users specifying full custom paths like `m/2645'/579218131'/894929996'/3'`

packages-ts/starknet-gauntlet/src/wallet/ledgerWallet.ts
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  async signTransaction(
+    transactions: Invocation[],
+    transactionsDetail: InvocationsSignerDetails,
+    abis?: Abi[],
+  ) {
+    if (abis && abis.length !== transactions.length) {
+      throw new Error('ABI must be provided for each transaction or no transaction')
+    }
```
krebernisak: Looks like it's an optional arg from the interface. Why do we check for it here, then do nothing with this data?

krebernisak (changes requested):
The functionality seems to be there, but let's please restructure it as a separate small package `@chainlink/starknet-gauntlet-ledger` and then bring it in via the CLI project in a more modular way.

packages-ts/starknet-gauntlet/src/dependencies/index.ts
sdrug: It makes sense. Please have another look, i think it looks better now. 
I'm not sure about separate `starknet-gauntlet-ledger` module, maybe you meant some signer module that takes care of the wallet definition?
Anyways, it's a submodule of the `starknet-gauntlet` atm, looks logical to me as it is.

krebernisak: So this is what I meant:

- `gauntlet-starknet` should know nothing about specific Ledger/Trezor/Vault/etc. wallets.
- `gauntlet-starknet` should define a Wallet API (interfaces, types required by the framework) and it can provide a default implementation (random or/and read-pk-from-env, for example)
-  Support for all other wallets should be external to `gauntlet-starknet`, they should be implemented in some other package and hooked in by the user when setting up their Gauntlet project.
- To achieve this here, I suggested to (1) remove all ledger specific code from `gauntlet-starknet` package, (2) move the code to a new `gauntlet-starknet-ledger` package, (3) hook in the Ledger wallet in the `gauntlet-starknet-cli` package (final project).

packages-ts/starknet-gauntlet-ledger/package.json
```diff
@@ -0,0 +1,33 @@
...
+      "format:check": "yarn prettier --check ./src",
+      "clean": "rm -rf ./dist/ ./bin/",
+      "build": "yarn clean && tsc -b",
+      "bundle": "yarn build && pkg ."
+    },
+    "dependencies": {
+      "@chainlink/starknet-gauntlet": "*",
+      "@ledgerhq/hw-app-starknet": "^1.0.1",
+      "@ledgerhq/hw-transport-node-hid": "^6.27.2",
+      "starknet": "^3.15.0"
```
archseer: This seems to re-import `starknet.js`, Kristijan recently moved this to the toplevel package.json: https://github.com/smartcontractkit/chainlink-starknet/blob/193c1e40be99a0016391b2cff13f53299bdc8e4c/package.json#L59

That breaks using the package as a library outside of the workspace though. I think we should move the import to`starknet-gauntlet`, then re-export it there

packages-ts/starknet-gauntlet/package.json
```diff
@@ -26,6 +26,7 @@
   },
   "dependencies": {
     "@chainlink/gauntlet-core": "0.3.1",
-    "cross-fetch": "^3.1.5"
+    "cross-fetch": "^3.1.5",
+    "starknet": "^4.4.2"
```
archseer: Ditto here, this was placed in the root `package.json`

packages-ts/starknet-gauntlet-multisig/package.json
```diff
@@ -25,7 +25,7 @@
     "bundle": "yarn build && pkg ."
   },
   "dependencies": {
-    "starsign-multisig": "0.2.3",
+    "starsign-multisig": "0.3.0",
```
archseer: This bump should wait for cairo 0.10 support since `starsign-multisig` updated their code https://github.com/eqlabs/starknet-multisig/commit/3eef643f14b82c41a97778f1d190b3b6f8cf0bd8

packages-ts/starknet-gauntlet-ledger/src/index.ts
```diff
@@ -0,0 +1,117 @@
...
+import { SignerInterface, encode } from 'starknet'
+import { Abi, Signature, Invocation, InvocationsSignerDetails } from 'starknet/types'
+import { TypedData, getMessageHash } from 'starknet/utils/typedData'
+import { fromCallsToExecuteCalldataWithNonce } from 'starknet/utils/transaction'
+import { calculcateTransactionHash, getSelectorFromName } from 'starknet/utils/hash'
+import Stark, { LedgerError } from '@ledgerhq/hw-app-starknet'
+
+// EIP-2645 path
+//  2645 - EIP number
+//  579218131 - layer - 31 lowest bits of sha256("starkex")
```
krebernisak, line 11: This should definitely not be `sha256("starkex")` which is their zkExchange product.
I guess it should be `31 lowest bits of sha256("starknet")` but please double check what is appropriate here.

packages-ts/starknet-gauntlet-ledger/src/index.ts
```diff
@@ -0,0 +1,117 @@
...
+      if (!match) {
+        throw new Error(
+          "Provided ledger path doesn't correspond the expected format m/2645'/579218131'/<application>'/<index>'",
+        )
+      }
+    }
+
+    // work around jest reimporting the package
+    // package is only allowed to be imported once
+    const TransportNodeHid = (await import('@ledgerhq/hw-transport-node-hid')).default
```
krebernisak: Not sure I understand the problem here?
What's different [than on Solana](https://github.com/smartcontractkit/chainlink-solana/blob/571a53a9eecc44fd8bec60522ec098a0f01512ff/gauntlet/packages/gauntlet-solana/src/commands/wallet.ts#L3)?

packages-ts/starknet-gauntlet/src/wallet/defaultWallet.ts
```diff
@@ -0,0 +1,25 @@
+import { ec, KeyPair, Signer } from 'starknet'
+import { IStarknetWallet } from './'
+import { Env } from '../dependencies'
+
+export class Wallet implements IStarknetWallet {
+  wallet: Signer
```
krebernisak: Wallet has a wallet?
```suggestion
  signer: Signer
```

packages-ts/starknet-gauntlet-ledger/src/index.ts
```diff
@@ -0,0 +1,117 @@
...
+    const signer = new LedgerSigner(ledgerConnector, path, publicKey)
+
+    return signer
+  }
+
+  async getPubKey(): Promise<string> {
+    return this.publicKey
+  }
+
+  async signTransaction(
```
krebernisak, line 60: Looks like the interface changed with latest version: https://github.com/0xs34n/starknet.js/blob/develop/src/signer/interface.ts

packages-ts/starknet-gauntlet-ledger/src/index.ts
```diff
@@ -0,0 +1,117 @@
...
+  static create = async (ledgerPath?: string, account?: string): Promise<Wallet> => {
+    const ledgerSigner: LedgerSigner = await LedgerSigner.create(ledgerPath)
+    return new Wallet(ledgerSigner, account)
+  }
+
+  getPublicKey = async (): Promise<string> => await this.wallet.getPubKey()
+  getAccountAddress = (): string => this.account
+}
+
+export const makeWallet = async (env: Env): Promise<IStarknetWallet> => {
```
krebernisak, line 115: We shouldn't create objects from env, but from config.
We read config from env, among other sources (file, for example).

See if there are opportunities to clean up some of these types and APIs.

packages-ts/starknet-gauntlet-starkgate/src/commands/bridge/deploy.ts
```diff
@@ -30,7 +30,7 @@ const makeContractInput = async (
   input: UserInput,
   context: ExecutionContext,
 ): Promise<ContractInput> => {
-  const defaultWallet = context.wallet.getAccountPublicKey()
+  const defaultWallet = context.wallet.getAccountAddress()
```
krebernisak, line 33: Let's fix the naming here:
```suggestion
  const accountAddr = context.wallet.getAccountAddress()
```

We should stop making the code intentionally hard to understand.

packages-ts/starknet-gauntlet-ledger/src/index.ts
RodrigoAD: Jest compiler complains if we import it by default for some reason. I dig into it at some point, and this was the best solution I could find. 
Solana doesn't have tests, so we don't have this problem...

packages-ts/starknet-gauntlet-ledger/src/index.ts
sdrug: This version of gauntlet doesn't have notion of `Config`. `flags` are being converted into `Env` that is used across the app as the source of config parameters. We inherit `makeEnv` wording from previous gauntlet versions. I agree that `Config` should be the final destination for the parameters parsed from different sources (flags, env, files etc.) and suggest to address it in a separate ticket since it's unrelated to the topic of this PR.

packages-ts/starknet-gauntlet-ledger/src/index.ts
sdrug: Jest attempts to import the package multiple times when compiling. The package only allows to be imported once (it takes ownership of the socket when imported).

packages-ts/starknet-gauntlet-starkgate/src/commands/bridge/deploy.ts
sdrug: I'll address the naming in Bridge PR. We inherited the naming with the whole gauntlet implementation.

packages-ts/starknet-gauntlet-ledger/src/index.ts
sdrug: `starkex` is used here for two reasons:
- They refer to this layer definition in all their docs
- Empirically no other values work with the provided version of the ledger app. I didn't try many options though neither did I find the definition of what should we provide there. I believe it was hardcoded to shortcut the initial app release.

shell.nix
```diff
@@ -23,6 +23,10 @@
...
+    # ledger specific packages
+    libusb
+    udev
   ];
```
archseer:
```suggestion
  ] ++ lib.optionals stdenv.isLinux [
    # ledger specific packages
    libudev-zero
    libusb1
  ];
```

krebernisak (approved):

PR #81, opened 2022-08-23
median provider returns OnchainConfigCodec

aalu1418:
depends on:
- [x] https://github.com/smartcontractkit/chainlink-relay/pull/69


relayer/pkg/chainlink/ocr2/utils.go
```diff
@@ -221,7 +221,7 @@ func parseConfigEventData(eventData []*caigotypes.Felt) (types.ContractConfig, e
 		Min: signedFelt(onchainConfigFelts[1]),
 		Max: signedFelt(onchainConfigFelts[2]),
 	}
-	onchainConfig, err := temp.Encode()
+	onchainConfig, err := median.StandardOnchainConfigCodec{}.Encode(temp)
```
aalu1418: @archseer how were you thinking about using a custom onchainConfig codec here? would that involve contract changes so version/min/max were encoded together?

aalu1418:
re-running e2e against the working branch in core: https://github.com/smartcontractkit/chainlink-starknet/actions/runs/2920514540

aalu1418:
> re-running e2e against the working branch in core: https://github.com/smartcontractkit/chainlink-starknet/actions/runs/2920514540

test successfully passed ✅ (but won't show in CI here, because separately run)

relayer/pkg/chainlink/ocr2/config_digester.go
```diff
@@ -59,15 +59,15 @@ func (d offchainConfigDigester) ConfigDigest(cfg types.ContractConfig) (types.Co
...
 		return configDigest, err
 	}
-	encodedOnchainConfig := []*big.Int{
-		big.NewInt(1), // onchainConfigVersion
-		new(big.Int).Mod(onchainConfig.Min, caigotypes.MaxFelt.Big()), // TODO: this wraps negative values correctly into felts, use a helper
-		new(big.Int).Mod(onchainConfig.Max, caigotypes.MaxFelt.Big()), // TODO: this wraps negative values correctly into felts, use a helper
+
+	encodedOnchainConfig := []*big.Int{}
+	for i := 0; i < len(onchainConfig); i++ {
+		modded := new(big.Int).Mod(onchainConfig[i], caigotypes.MaxFelt.Big()) // TODO: this wraps negative values correctly into felts, use a helper
```
archseer: This shouldn't be done here, and not for all values. Only answers, observations & min/max answer are signed values. Make this a part of the encode operation.

krebernisak: Isn't out `cfg.OnchainConfig` already encoded as bytes at this point, why do we decode then encode again to compute the digest?

relayer/pkg/chainlink/ocr2/utils.go
```diff
@@ -217,11 +218,10 @@ func parseConfigEventData(eventData []*caigotypes.Felt) (types.ContractConfig, e
 		return types.ContractConfig{}, errors.Wrapf(err, "expected onchainConfig version 1, got %v", onchainConfigVersion)
 	}
 
-	temp := median.OnchainConfig{
+	onchainConfig, err := medianreport.OnchainConfigCodec{}.Encode(median.OnchainConfig{
```
krebernisak: In this file we have another ad-hoc encoding, this time from `[]*caigotypes.Felt` to `[]byte`:

```
onchainConfigFelts := eventData[index:(index + int(onchainConfigLen))]

onchainConfigVersion := onchainConfigFelts[0].Uint64()
if onchainConfigVersion != 1 {
	return types.ContractConfig{}, errors.Wrapf(err, "expected onchainConfig version 1, got %v", onchainConfigVersion)
}

onchainConfig, err := medianreport.OnchainConfigCodec{}.Encode(median.OnchainConfig{
	Min: signedFelt(onchainConfigFelts[1]),
	Max: signedFelt(onchainConfigFelts[2]),
})
```

Could we simplify this by going directly from `[]*caigotypes.Felt` to `[]byte`?

Or we could move this logic as a helper function to `medianreport/onchainconfig.go` so we contain all these onchainconfig encoding/decoding in this one file. (notice that the version magic number is also duplicated here)

relayer/pkg/chainlink/ocr2/config_digester.go
aalu1418: @krebernisak you are correct, the main thing is that onchainconfig is hashed as individual components because of using felts - so `encodedOnchainConfig` in this sense is improperly named since it's actually the OnchainConfig decoded into `*big.int` components

aalu1418:
manually running e2e test against the corresponding branch in core: https://github.com/smartcontractkit/chainlink-starknet/actions/runs/2927711618

aalu1418:
> manually running e2e test against the corresponding branch in core: https://github.com/smartcontractkit/chainlink-starknet/actions/runs/2927711618

✅ 

also note: this PR can be squashed!

relayer/pkg/chainlink/ocr2/config_digester.go
```diff
@@ -57,18 +57,12 @@ func (d offchainConfigDigester) ConfigDigest(cfg types.ContractConfig) (types.Co
 		oracles = append(oracles, signer, transmitter)
 	}
 
-	offchainConfig := EncodeBytes(cfg.OffchainConfig)
+	offchainConfig := starknet.EncodeBytes(cfg.OffchainConfig)
```
krebernisak, line 60: Shouldn't this fn also be renamed as `DecodeBytes` considering the usual encoding/decoding process is:

```go
Encode(T) ([]byte, error)
Decode([]byte) (T, error)
```

krebernisak (approved):

relayer/pkg/chainlink/ocr2/config_digester.go
krebernisak: Also I'd consider renaming as `DecodeFelts` & `EncodeFelts` 🤔

relayer/pkg/chainlink/ocr2/medianreport/onchainconfig.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return median.OnchainConfig{}, err
+	}
+
+	// convert felts to big.Ints
+	min := starknet.FeltToBigInt(&caigotypes.Felt{Int: felts[1]})
+	max := starknet.FeltToBigInt(&caigotypes.Felt{Int: felts[2]})
+
+	if !(min.Cmp(max) <= 0) {
+		return median.OnchainConfig{}, fmt.Errorf("OnchainConfig min (%v) should not be greater than max(%v)", min, max)
```
sdrug, line 70:
```suggestion
		return median.OnchainConfig{}, fmt.Errorf("OnchainConfig min (%v) should be less than max(%v)", min, max)
```

aalu1418: the error msg just follows what is used in `libocr` so i'm more inclined to leave it as is for now: https://github.com/smartcontractkit/libocr/blob/db92a9fdaa520171ae8973e5462136402c795151/offchainreporting2/reportingplugin/median/median.go#L64

relayer/pkg/chainlink/ocr2/config_digester.go
aalu1418: I'll take a look at this in a separate PR, but yeah I agree the naming isn't very clear

PR #82, opened 2022-08-24
negative handling in report

aalu1418:
needs to be rebased on top of:
- [x] https://github.com/smartcontractkit/chainlink-starknet/pull/81 (otherwise libocr versions conflict when building node from core develop)

### DIFF
--- a/integration-tests/smoke/ocr2_test.go
+++ b/integration-tests/smoke/ocr2_test.go
@@ -109,7 +109,7 @@ var _ = Describe("StarkNET OCR suite @ocr", func() {
 		})
 
 		By("Deploying OCR2 contract", func() {
-			ocrAddress, err = sg.DeployOCR2ControllerContract(0, 100000000000, decimals, "auto", linkTokenAddress)
+			ocrAddress, err = sg.DeployOCR2ControllerContract(-100000000000, 100000000000, decimals, "auto", linkTokenAddress)
 			Expect(err).ShouldNot(HaveOccurred(), "OCR contract deployment should not fail")
 		})
 
@@ -146,8 +146,12 @@ var _ = Describe("StarkNET OCR suite @ocr", func() {
 			sum`
 
 			observationSource := `
-			sum [type="sum" values=<[900000000]>]
-			sum
+			pos [type="sum" values=<[900000000]>]
+			neg [type="sum" values=<[-900000000]>]
+			pick [type="any"]
+			
+			pos -> pick
+			neg -> pick
 			`
 
 			t.SetBridgeTypeAttrs(&client.BridgeTypeAttributes{
@@ -178,9 +182,13 @@ var _ = Describe("StarkNET OCR suite @ocr", func() {
 			stuckCount := 0
 			ctx := context.Background() // context background used because timeout handeld by requestTimeout param
 
+			// assert both positive and negative values have been seen
+			var positive bool
+			var negative bool
+
 			for start := time.Now(); time.Since(start) < roundWaitTimeout; {
-				// end condition: enough rounds have occured
-				if increasing == increasingCountMax {
+				// end condition: enough rounds have occured, and positive and negative answers have been seen
+				if increasing == increasingCountMax && positive && negative {
 					break
 				}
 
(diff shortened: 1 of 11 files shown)


relayer/pkg/chainlink/ocr2/medianreport/report.go
```diff
@@ -59,7 +64,11 @@ func (c ReportCodec) BuildReport(oo []median.ParsedAttributedObservation) (types
 	var observations []junotypes.Felt
 	for i, o := range oo {
 		observers[i] = byte(o.Observer)
-		observations = append(observations, junotypes.BigToFelt(o.Value))
+		obsBytes, err := bigbigendian.SerializeSigned(junotypes.FeltLength, o.Value)
+		if err != nil {
+			return nil, errors.Wrap(err, "error in serializing observation")
+		}
+		observations = append(observations, junotypes.BytesToFelt(obsBytes))
```
archseer: I think you just need to use the same functions I used for transmission: `parseAnswer` and `signedFelt` for parsing: https://github.com/smartcontractkit/chainlink-starknet/blob/7b9676a8e87b5f474c015ad0e4e8b7b788e950d4/relayer/pkg/chainlink/ocr2/utils.go#L75-L98 this `mod` for serializing: https://github.com/smartcontractkit/chainlink-starknet/blob/7b9676a8e87b5f474c015ad0e4e8b7b788e950d4/relayer/pkg/chainlink/ocr2/config_digester.go#L69-L70

Cairo has no notion of signed values: negative values have to be wrapped into the upper half of `PRIME` (so `0` to `PRIME/2` is positive, `PRIME/2` to `PRIME` is negative) or transmission will fail. Try removing this function in aggregator tests for example:

https://github.com/smartcontractkit/chainlink-starknet/blob/7b9676a8e87b5f474c015ad0e4e8b7b788e950d4/contracts/test/ocr2/aggregator-test.ts#L14-L18

archseer (changes requested):

integration-tests/smoke/ocr2_test.go
```diff
@@ -146,8 +146,12 @@ var _ = Describe("StarkNET OCR suite @ocr", func() {
...
 
 			observationSource := `
-			sum [type="sum" values=<[900000000]>]
-			sum
+			pos [type="sum" values=<[900000000]>]
+			neg [type="sum" values=<[-900000000]>]
+			pick [type="any"]
+			
+			pos -> pick
+			neg -> pick
```
aalu1418: this introduces some non-deterministic behavior because the job run will randomly choose a positive or negative value. the test will run will continue until both a positive and negative value have been seen onchain - because the test waits for 10 rounds, this usually occurs within 10 rounds

aalu1418:
this PR should be squash merged

integration-tests/smoke/ocr2_test.go
archseer: image

relayer/pkg/chainlink/ocr2/medianreport/report.go
```diff
@@ -59,7 +61,8 @@ func (c ReportCodec) BuildReport(oo []median.ParsedAttributedObservation) (types
 	var observations []junotypes.Felt
 	for i, o := range oo {
 		observers[i] = byte(o.Observer)
-		observations = append(observations, junotypes.BigToFelt(o.Value))
+		obs := starknet.BigIntToFelt(o.Value) // converst negative bigInts to corresponding felt in bigInt form
```
archseer: Looks good, but the functions are now very similarly named: `starknet.BigIntToFelt`  and `junotypes.BigToFelt`. Maybe `SignedBigIntToFelt`? That's getting a bit long which is why I called the util helper `signedFelt`.

relayer/pkg/chainlink/ocr2/medianreport/report.go
```diff
@@ -48,6 +49,7 @@ func (c ReportCodec) BuildReport(oo []median.ParsedAttributedObservation) (types
 		return oo[i].JuelsPerFeeCoin.Cmp(oo[j].JuelsPerFeeCoin) < 0
 	})
 	juelsPerFeeCoin := oo[num/2].JuelsPerFeeCoin
+	juelsPerFeeCoin = starknet.BigIntToFelt(juelsPerFeeCoin) // converts negative bigInts to corresponding felt in bigInt form
```
archseer: We probably don't need this since this number should never be negative

integration-tests/smoke/ocr2_test.go
aalu1418: lol fixed this - with an EA and manually changing the returned value after a few successful rounds
`900000000 => -900000000`

relayer/pkg/chainlink/ocr2/medianreport/report.go
aalu1418: i think this might still be worth handling...I'm just thinking defensively where we always want the value to be updated onchain even though the payment might be incorrect

scenario 1 (do not convert negative val): if enough operators submit negative numbers, the median will be negative. this would cause the transaction to fail because a negative big.Int doesn't fit on the starkcurve

scenario 2 (return an error if negative): again if enough operators submit negative values, the negative median would cause an error for a valid report, and reports would stop being broadcast

relayer/pkg/chainlink/ocr2/medianreport/report.go
aalu1418: renamed to `SignedBigToFelt` and `FeltToSignedBig` similar to the juno types but with explicit `Signed`

also renamed additional utilities according to what @krebernisak suggested: https://github.com/smartcontractkit/chainlink-starknet/pull/81#discussion_r955150142

relayer/pkg/starknet/utils.go
```diff
@@ -46,8 +46,8 @@ func Min[T constraints.Ordered](a, b T) T {
 	return b
 }
 
-// Encodes a byte slice as a bunch of felts. First felt indicates the total byte size.
-func EncodeBytes(data []byte) (felts []*big.Int) {
+// EncodeFelts takes a byte slice and splits as bunch of felts. First felt indicates the total byte size.
+func EncodeFelts(data []byte) (felts []*big.Int) {
```
archseer, line 50: Nit: this makes it seem like we're encoding felts, where we're actually encoding bytes as felts

archseer (approved):

PR #83, opened 2022-08-25
Update multisig to use upstream package

archseer:


packages-ts/starknet-gauntlet-multisig/src/wrapper/index.ts
```diff
@@ -109,7 +109,7 @@ export const wrapCommand = <UI, CI>(
         multisig,
         proposal: {
           id: proposalId,
-          approvers: proposal.tx.threshold,
+          confirmations: proposal.tx.confirmations,
```
archseer, line 112: This got renamed just prior to the release

krebernisak (approved):

PR #85, opened 2022-08-25
Contracts path start with chainlink.

krebernisak:


contracts/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "compile:cairo": "hardhat starknet-compile ./contracts ./vendor",
+    "compile:cairo": "hardhat starknet-compile ./src ./vendor",
```
archseer: Why break cairo conventions here?

krebernisak: I think this is mostly a hardhat convention, that's why the cairo plugin uses the same default `./contracts` folder. 

Both starkgate & OZ lib projects have their contracts source under `./src` so I choose that approach mainly to remove the duplicate `./contracts/contracts` folder structure.

krebernisak: I don't think this is actually required as we have `paths.cairoPaths` defined.

PR #88, opened 2022-08-29
Format cairo smart contracts + add to CI workflow

krebernisak:


contracts/src/chainlink/cairo/ocr2/validator.cairo
```diff
@@ -46,20 +37,13 @@ func constructor{
...
-    range_check_ptr,
-}(
-    prev_round_id: felt,
-    prev_answer: felt,
-    round_id: felt,
-    answer: felt
-) -> (valid: felt):
+func validate{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
+    prev_round_id : felt, prev_answer : felt, round_id : felt, answer : felt
+) -> (valid : felt):
```
archseer, line 42: I think `cairo-fmt` is still pretty dumb, I find this a lot less readable than the original formatting

.github/workflows/changesets.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     steps:
       # Checkout this repository
       - name: Checkout Repo
-        uses: actions/checkout@f25a3a9f25bd5f4c5d77189cab02ff357b5aedeb # v2.4.1
+        uses: actions/checkout@v3.0.2
```
archseer: All of these actions should be pinned to a specific SHA, tags/versions are unsafe since they're mutable

krebernisak: Updating....

contracts/src/chainlink/cairo/ocr2/validator.cairo
krebernisak: While perhaps I'd agree, I also appreciate the KISS approach of everybody using the same formatter and just having consistency all-round. I suggest we use it, and hope it improves with time. The next upgrade brings syntax changes to the Cairo lang so we'll see how that goes.

archseer (approved):

PR #92, opened 2022-08-31
Transmitter reimbursement calculation + aggregator.cairo cleanup 

archseer:


contracts/src/chainlink/cairo/ocr2/aggregator.cairo
```diff
@@ -759,13 +729,6 @@ func latest_transmission_details{syscall_ptr : felt*, pedersen_ptr : HashBuiltin
     let (epoch_and_round) = latest_epoch_and_round_.read()
     let (transmission : Transmission) = transmissions_.read(latest_round_id)
 
-    let round = Round(
```
archseer, line 738: This was an unused variable in some function

relayer/pkg/chainlink/ocr2/medianreport/report.go
```diff
@@ -52,6 +52,8 @@ func (c ReportCodec) BuildReport(oo []median.ParsedAttributedObservation) (types
 	juelsPerFeeCoin = starknet.SignedBigToFelt(juelsPerFeeCoin) // converts negative bigInts to corresponding felt in bigInt form
 	juelsPerFeeCoinFelt := junotypes.BigToFelt(juelsPerFeeCoin)
 
+	// TODO: source and include gasPrice
```
archseer: Note: a value needs to be added here, then the aggregator contracts package needs to be rebuilt and uploaded to npm. Tests currently pass since it's still using the old contract versions

contracts/src/chainlink/cairo/access/SimpleReadAccessController/library.cairo
```diff
@@ -17,6 +16,11 @@ namespace simple_read_access_controller:
         return ()
     end
 
+    # Gives access to:
+    # - any externally owned account (note that offchain actors can always read
+    # any contract storage regardless of onchain access control measures, so this
+    # does not weaken the access control while improving usability)
+    # - accounts explicitly added to an access list
```
krebernisak, line 22: Maybe avoid to mention EOA and we just say accounts:

```suggestion
    # - accounts (note that offchain actors can always read
    # any contract storage regardless of onchain access control measures, so this
    # does not weaken the access control while improving usability)
    # - contract addresses explicitly added to an access list
```

contracts/src/chainlink/cairo/access/ownable.cairo
```diff
@@ -22,6 +23,9 @@ end
 func Ownable_only_owner{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}():
     let (owner) = Ownable_owner.read()
     let (caller) = get_caller_address()
+    with_attr error_message("Ownable: Caller is the zero address"):
+        assert_not_zero(caller)
+    end
```
krebernisak: This seems unnecessary as the next check will catch all that we require. And now it looks like we disabled ownership transfers to the 0 account.

contracts/src/chainlink/cairo/access/ownable.cairo
```diff
@@ -3,6 +3,7 @@
 
 from starkware.cairo.common.cairo_builtins import HashBuiltin
 from starkware.starknet.common.syscalls import get_caller_address
+from starkware.cairo.common.math import assert_not_zero
 
 @storage_var
 func Ownable_owner() -> (owner_address : felt):
```
krebernisak: Maybe we additionally disable initializing the owner to 0 as it's done here: https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.7/ConfirmedOwnerWithProposal.sol#L18

contracts/src/chainlink/cairo/access/ownable.cairo
```diff
@@ -38,6 +42,9 @@ end
 func Ownable_transfer_ownership{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
     new_owner : felt
 ) -> ():
+    with_attr error_message("Ownable: Cannot transfer to zero address"):
+        assert_not_zero(new_owner)
+    end
```
krebernisak: I feel like this contract is missing native capability to burn ownership.

It's the same with our Ethereum implementation as there you can't accept ownership from a burn address (0x0) so the workflow would be to deploy a dummy contract that would accept ownership and burn it (have no capability to use it).

But then there is a problem of how does one programatically confirm that the ownership is actually burned.

In any case, maybe we should consider adding this capability in (could be 2-step to be safer).

contracts/src/chainlink/cairo/access/ownable.cairo
```diff
@@ -47,6 +54,10 @@ func Ownable_accept_ownership{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*,
     ) -> (new_owner : felt):
     let (proposed_owner) = Ownable_proposed_owner.read()
     let (caller) = get_caller_address()
+    # caller cannot be zero address to avoid overwriting owner when proposed_owner is not set
+    with_attr error_message("Ownable: caller is the zero address"):
+        assert_not_zero(caller)
+    end
```
krebernisak: We should try to write a test to see if this is even possible, maybe ask our auditors for advice here, as all these `assert_not_zero(caller)` checks are slightly confusing.

contracts/src/chainlink/cairo/emergency/SequencerUptimeFeed/library.cairo
```diff
@@ -230,7 +229,7 @@ func _record_round{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check
     _set_round(round_id, round)
 
     let (sender) = get_caller_address()
-    NewRound.emit(round)
+    NewRound.emit(round_id=round_id, started_by=sender, started_at=timestamp)
```
krebernisak, line 242: Ok, so I just noticed these events are only emitted in our L2 uptime feed implementation (implements IAggregator) but not emitted in our OCRv2 feed implementation (there we only emit 'new_transmission', also implements IAggregator).

- Ethereum OCRv2 implementation emits both (new and old legacy) events: https://github.com/smartcontractkit/libocr/blob/master/contract2/dev/OCR2Aggregator.sol#L797
- L2EP implementation for Optimism/Arbitrum emits old legacy events: https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.8/dev/OptimismSequencerUptimeFeed.sol#L151

This is totally not consistent across multiple implementations. What do we do? cc @cds95

contracts/src/chainlink/cairo/ocr2/aggregator.cairo
```diff
@@ -174,6 +172,7 @@ func constructor{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_p
     description : felt,
 ):
     Ownable_initializer(owner)
+    simple_read_access_controller.initialize(owner)
```
krebernisak: This will clash, write to the same slot, but that should be ok right?
Do we need both though, as `simple_read_access_controller` already brings/includes `ownable`?

contracts/src/chainlink/cairo/ocr2/aggregator.cairo
```diff
@@ -300,7 +264,9 @@ func set_config{
     assert_nn_le(oracles_len, MAX_ORACLES)  # oracles_len <= MAX_ORACLES
     assert_lt(3 * f, oracles_len)  # 3 * f < oracles_len
     assert_nn(f)  # f is positive
-    assert onchain_config_len = 0  # empty onchain config provided
+    with_attr error_message("onchain_config must be empty"):
```
krebernisak, line 233: Can we please add a note here why it must be empty. Thx!

contracts/src/chainlink/cairo/ocr2/aggregator.cairo
```diff
@@ -1133,10 +1137,18 @@ end
 
 # --- Transmitter Payment
 
-func calculate_reimbursement() -> (amount : felt):
-    # TODO:
-    let amount = 0
-    return (amount)
+const MARGIN = 115
+
+func calculate_reimbursement{range_check_ptr}(
```
krebernisak, line 1142: Hmm, some thinking...

1. Static gas estimates in contracts are brittle. Gas values could change unexpectedly with a network upgrade and our estimates are off at that point (would require feed redeploys).
2. This is not an issue with on-chain gas metering, but that is only available on EVM.
3. I feel we can and probably should do the estimates off-chain. The only thing I worry it would slow down the libocr's runtime report generation hot path.
4. Instead of reporting on `gas_price` we should report the `max_fee` as this is the API the StarkNet network uses. It would represent the minimum `max_fee` as estimated by the cluster, which the transmitter should use (it's agreed upon). Transmitter is disallowed to use a fee less than agreed (implicitly for now, alerting system can flag) but it is allowed to use a higher fee if conditions changed while the transaction is in/pre flight.


So my ad-hoc suggestion would be something like this:

1. We move estimates off-chain as that way they can be cheaper (no gas spent to estimate gas) and safer (not static but dynamic on any gas costs upgrades + simpler on-chain code).
2. The DON reports on "minimum fee/cost" (signed) and the transmitter reports on his "used fee/cost" (unverified). We add the `max_fee_proposed` and `max_fee_used` to the `transmit` input, `max_fee_proposed` is used as part of report, `max_fee_used` is just logged as part of the 'new_transmission' event.
3. We cap the `max_fee_used` to be at most something like 2x over `max_fee_proposed`, so if someone decides to drain they will need to do it slowly until we detect it and kick them out.
4. We reimburse the transmitter using the `max_fee_used` (optimistically trust)
5. We add the alert condition to the on-chain monitor service, which would alert on "transmitter spent less than 'max_fee_proposed'" and "transmitter reported invalid 'max_fee_used'".

cc @aalu1418 any thoughts? This is also important in the context of TXM, because for example if a cluster agrees on a static single price/fee for a transmitter to use this would mean no gas bumping without re-agreeing and reaching consensus on a new price to use. Could make the protocol a lot slower at times.

contracts/src/chainlink/cairo/access/SimpleReadAccessController/library.cairo
archseer: I copy pasted this from the solidity version

contracts/src/chainlink/cairo/emergency/SequencerUptimeFeed/library.cairo
archseer: I'd only use the new event types. We usually drop old view methods in new implementations too (Solana, Terra only implemented v3 queries)

contracts/src/chainlink/cairo/ocr2/aggregator.cairo
archseer: Hmm they will both end up writing to the same storage variable but it's the same value. I could remove `Ownable` from aggregator and just use the implicit one from `simple_read_access_controller`

contracts/src/chainlink/cairo/access/ownable.cairo
archseer: OpenZeppelin has `renounce_ownership` which explicitly sets it to `0`: https://github.com/OpenZeppelin/cairo-contracts/blob/b324794e5337ce72d65266df660e4918fde60c70/src/openzeppelin/access/ownable/library.cairo#L86-L94

contracts/src/chainlink/cairo/access/ownable.cairo
archseer: I wanted to keep it as similar as possible to the OpenZeppelin version: https://github.com/OpenZeppelin/cairo-contracts/blob/b324794e5337ce72d65266df660e4918fde60c70/src/openzeppelin/access/ownable/library.cairo#L44-L58

They deliberately split the two so that zero address has a more distinct error message: https://github.com/OpenZeppelin/cairo-contracts/pull/422

contracts/src/chainlink/cairo/access/ownable.cairo
archseer: I've done this according to https://github.com/crytic/building-secure-contracts/tree/master/not-so-smart-contracts/cairo/access_controls (org owned by Trail of Bits)

contracts/src/chainlink/cairo/emergency/SequencerUptimeFeed/library.cairo
krebernisak: I agree. 

Unfortunately, the L2EP was not designed with the new OCRv2 interface in mind. The Optimism and Arbitrum implementations both emit the old events, StarkNet also. Even if we wanted to emit the `NewTransmission` aggregator event on StarkNet L2EP, we don't have all the data (L1 validator doesn't send all, only status) and we don't have defined what that data should be for an x-domain representation of a feed.

My suggestion is we leave this as it currently is (same as our L2EP implementation on other networks, not compliant with OCRv2 aggregator interface), we acknowledge and raise the issue, then we fix it for L2EP on all networks at once.

Also, it's too close to the first audit to add a big API update like this.

ops/localenv/main.go
```diff
@@ -30,7 +30,7 @@ func main() {
 	switch strings.ToLower(os.Args[1]) {
 	// create k8s cluster + resources
 	case "create":
-		run("create registry", "k3d", "registry", "create", "registry.localhost", "--port", "12345")
+		run("create registry", "k3d", "registry", "create", "registry.localhost", "--port", "12345", "--default-network", "podman")
```
archseer: TODO: need to revert that/make it detect podman

contracts/src/chainlink/cairo/emergency/SequencerUptimeFeed/library.cairo
```diff
@@ -102,6 +102,26 @@ func l1_sender{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr
     return (address)
 end
 
+# --- Ownership ---
+
+@view
+func owner{pedersen_ptr : HashBuiltin*, syscall_ptr : felt*, range_check_ptr}() -> (owner : felt):
```
krebernisak: Looks like we are also missing these Ownership view/external fn definitions in the `proxy.cairo`, right?

Also, let's rename the `proxy.cairo` as `aggregator_proxy.cairo` as it's not a general proxy, but specifically an aggregator interface proxy.

krebernisak: We could also use an `shouldBehaveLikeOwnableContract` behaviour test we could use for multiple contracts where this behaviour is expected.

ops/localenv/main.go
krebernisak: What do you want to do here?

packages-ts/starknet-gauntlet-ocr2/test/commands/ocr2.test.ts
```diff
@@ -1,4 +1,6 @@
 import { makeProvider } from '@chainlink/starknet-gauntlet'
+// TODO: Import from package
+import deployOZCommand from '../../../starknet-gauntlet-oz/src/commands/account/deploy'
```
krebernisak, line 3: Agreed, most relative imports across packages in the monorepo should be avoided.

relayer/pkg/chainlink/ocr2/medianreport/report.go
```diff
@@ -52,6 +53,9 @@ func (c ReportCodec) BuildReport(oo []median.ParsedAttributedObservation) (types
 	juelsPerFeeCoin = starknet.SignedBigToFelt(juelsPerFeeCoin) // converts negative bigInts to corresponding felt in bigInt form
```
krebernisak, line 53: Do we expect the `juelsPerFeeCoin` to ever be a negative number? 
We should probably error here if that happens, right? And probably the same for the `gasPrice`.

relayer/pkg/chainlink/ocr2/medianreport/report.go
```diff
@@ -52,6 +53,9 @@ func (c ReportCodec) BuildReport(oo []median.ParsedAttributedObservation) (types
...
 	juelsPerFeeCoinFelt := junotypes.BigToFelt(juelsPerFeeCoin)
 
+	// TODO: source gasPrice from last block
+	gasPrice := junotypes.BigToFelt(big.NewInt(1))
```
krebernisak:
```suggestion
	// TODO: source from observations
	gasPrice := big.NewInt(1) // := oo[num/2].GasPrice
	gasPriceFelt := junotypes.BigToFelt(gasPrice)
```

contracts/src/chainlink/cairo/ocr2/IAggregator.cairo
```diff
@@ -21,6 +21,7 @@ func NewTransmission(
     config_digest : felt,
     epoch_and_round : felt,
     reimbursement : felt,
+    gas_price : felt,
```
krebernisak: nit: I'd put the `gas_price` next to `juels_per_fee_coin` as they are of the same type - input observations.

Also, as this event change our parsing of the event needs to be updated and the `TestNewTransmissionEvent_Parse` test in `ocr2/events_test.go`

contracts/test/ocr2/aggregator.test.ts
```diff
@@ -241,7 +244,7 @@ describe('aggregator.cairo', function () {
     assert.equal(e.data.round_id, 1n)
     assert.equal(e.data.observation_timestamp, 1n)
     assert.equal(e.data.epoch_and_round, 1n)
-    assert.equal(e.data.reimbursement, 0n)
+    // assert.equal(e.data.reimbursement, 0n)
```
krebernisak: Let's test a number here

contracts/src/chainlink/cairo/ocr2/aggregator.cairo
```diff
@@ -1109,10 +1079,18 @@ end
...
```
krebernisak, line 1072: We should consider making some of these coefficients configurable via `set_billing`

contracts/src/chainlink/cairo/emergency/SequencerUptimeFeed/library.cairo
krebernisak: @archseer I've made these changes.

contracts/test/ocr2/aggregator.test.ts
archseer: This is based on the gas calculation so we'd have to update any time the constant changes

packages-ts/starknet-gauntlet-ocr2/test/commands/ocr2.test.ts
archseer: I copy pasted the TODO from another test that already did this, let's fix in a follow up

relayer/pkg/chainlink/ocr2/medianreport/report.go
archseer: We don't expect it to be negative, but the EVM codec omits this check too: https://github.com/smartcontractkit/libocr/blob/cdfa6bef133ac1c69042ff04295ab65d867837c9/offchainreporting2/reportingplugin/median/evmreportcodec/reportcodec.go#L49-L53

I think it's done somewhere inside libocr maybe?

contracts/src/chainlink/cairo/ocr2/aggregator.cairo
```diff
@@ -1109,10 +1067,19 @@ end
 
 # --- Transmitter Payment
 
-func calculate_reimbursement() -> (amount : felt):
-    # TODO:
-    let amount = 0
-    return (amount)
+const MARGIN = 115
```
krebernisak, line 1070: Not sure how useful is the fixed margin as we can scale dynamically via `billing.gas_base|gas_per_signature` configuration.

contracts/src/chainlink/cairo/ocr2/aggregator.cairo
```diff
@@ -565,11 +506,11 @@ func transmit{
     let (median_idx : felt, _) = unsigned_div_rem(observations_len, 2)
     let median = observations[median_idx]
 
-    # Check abs(median) is in i192 range.
-    # NOTE: (assert_lt_felt(-i192::MAX, median) doesn't work correctly so we have to use abs!)
+    # Check abs(median) is in i128 range.
```
krebernisak, line 509: Let's add a test for this.

We could test the min/max values during the 'should transmit correctly' test

krebernisak (dismissed):

packages-ts/starknet-gauntlet-ocr2/test/commands/ocr2.test.ts
krebernisak: GH Issue: https://github.com/smartcontractkit/chainlink-starknet/issues/111

krebernisak (approved):
LGTM! 🚀

PR #93, opened 2022-08-31
Restructure tests

qperrot:
Test for starkgate bridge with our link_token


contracts/test/bridge/tokenBridge.ts
```diff
@@ -0,0 +1,189 @@
...
+    })
+
+    it('Should deposit to the L2 contract, L1 balance should be decreased by 2', async () => {
+      await testERC20.approve(newStarkNetERC20Bridge.address, 2)
+      await newStarkNetERC20Bridge.deposit(2, owner.starknetContract.address)
+
+      const balance = await testERC20.balanceOf(deployer.address)
+      expect(balance).to.equal(8)
+    })
+  })
```
ivpavici: nitpick: new line here :)

contracts/test/bridge/tokenBridge.ts
```diff
@@ -0,0 +1,189 @@
...
+      expectAddressEquality(flushL1Messages[0].args.to_address, tokenBridgeContract.address)
+      expectAddressEquality(flushL1Messages[0].address, mockStarknetMessaging.address)
+
+      let { balance: balance } = await linkTokenContract.call('balanceOf', {
+        account: owner.starknetContract.address,
+      })
+      expect(uint256.uint256ToBN(balance)).to.deep.equal(number.toBN(2))
+    })
+
+    it('Should initiate withdraw and send message to L1 ', async () => {
```
ivpavici: nitpick: remove whitespace - "L1 '

contracts/src/chainlink/solidity/mocks/TestERC20.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: Apache-2.0.
+pragma solidity ^0.6.12;
+
+import "../../../../vendor/starkware-libs/starkgate-contracts/src/starkware/solidity/tokens/ERC20/ERC20.sol";
```
ivpavici: Maybe these contracts could belong here ? :
https://github.com/smartcontractkit/chainlink-starknet/tree/develop/contracts/vendor/starkware-libs

contracts/test/bridge/tokenBridge.ts
```diff
@@ -0,0 +1,189 @@
...
+    })
+
+    let linkTokenFactory = await starknet.getContractFactory('link_token')
+    linkTokenContract = await linkTokenFactory.deploy({ owner: tokenBridgeContract.address })
+
+    const accounts = await ethers.getSigners()
+    deployer = accounts[0]
+
+    const contractFile = fs.readFileSync(
+      '../node_modules/internals-starkgate-contracts/artifacts/0.0.3/eth/StarknetERC20Bridge.json',
```
ivpavici: can this be retrieved in a similar way like with loadContract?

contracts/src/chainlink/solidity/mocks/TestERC20.sol
qperrot: The contracts under vendor look to not be automatically compile unless if they are used by a contract which is outside vendor folder.

contracts/test/utils.ts
```diff
@@ -1,4 +1,22 @@
 import { expect } from 'chai'
+import fs from 'fs'
+import { json } from 'starknet'
+
+export const loadStarkgateContract = (name: string): any => {
```
ivpavici: Is this whole folder now redundant: https://github.com/smartcontractkit/chainlink-starknet/tree/develop/packages-ts/integration-starkgate ?
There is a function with the same name in utils folder over there

contracts/vendor/starkware-libs/starkgate-contracts/src/starkware/solidity/tokens/ERC20/ERC20.sol
```diff
@@ -0,0 +1,129 @@
+// SPDX-License-Identifier: Apache-2.0.
```
ivpavici: Check if these contracts can be imported from OpenZeppelin via npm instead of copying? Check if there are any differences

krebernisak: Yeah, we can use the OZ Solidity ERC20 implementation by simply importing their NPM package. There is no need to use the StarkGate ERC20 or copy the contract over.

contracts/test/bridge/tokenBridge.ts
```diff
@@ -0,0 +1,184 @@
+import { TIMEOUT } from '../constants'
```
krebernisak: These tests should not go under our `/contracts` project, but we have a specific integration project for it already: https://github.com/smartcontractkit/chainlink-starknet/tree/develop/packages-ts/integration-starkgate

Let's move it all there.

contracts/test/bridge/artifacts-test/Proxy.json
```diff
@@ -0,0 +1,371 @@
+{
```
krebernisak: I think this should be available directly from here: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/ERC1967/ERC1967Proxy.sol

We want to avoid using static artifacts as much as we can. Instead add a NPM dependency if not available, load the artifact from there and deploy.

contracts/src/chainlink/solidity/mocks/TestERC20.sol
```diff
@@ -0,0 +1,16 @@
...
+
+/*
+  An ERC20 for testing where anyone can set the balance for everyone.
+*/
+contract TestERC20 is ERC20 {
+  function setBalance(address account, uint256 amount) external {
```
krebernisak: Not sure if this custom ERC20 is required. What's the purpose of adding this method?

I must be missing something as currently I see you added the method, and you are just testing this specific method.

contracts/test/token/starkgate/behavior/ERC20.ts
```diff
@@ -98,15 +98,19 @@ export const shouldBehaveLikeStarkGateERC20 = (beforeFn: BeforeFn) => {
           amount: uint256.bnToUint256(103),
         })
         throw new Error('This should not pass!')
-      } catch (error: any) {}
+      } catch (error: any) {
+        expect(/assert/gi.test(error.message)).to.be.true
```
krebernisak: Can we test for more here than just for the 'assert' string? 
There should be more data/info in that error message, right?

packages-ts/integration-starkgate/test/ERC20token.ts
```diff
@@ -1,404 +0,0 @@
-import { starknet } from 'hardhat'
```
krebernisak, line 1: 👍

contracts/test/bridge/tokenBridge.ts
```diff
@@ -0,0 +1,184 @@
...
+  let mockStarknetMessaging: Contract
+  let proxy: Contract
+  let testERC20: Contract
+  let newStarkNetERC20Bridge: Contract
+
+  before(async () => {
+    owner = await starknet.deployAccount('OpenZeppelin')
+
+    let tokenBridgeFactory = await starknet.getContractFactory(
+      '../../node_modules/@chainlink-dev/starkgate-contracts/artifacts/token_bridge.cairo',
```
krebernisak: Do we really need this full path here?

contracts/test/bridge/tokenBridge.ts
```diff
@@ -0,0 +1,184 @@
...
+    const mockStarknetMessagingFactory = await ethers.getContractFactory(
+      'MockStarkNetMessaging',
+      deployer,
+    )
+    mockStarknetMessaging = await mockStarknetMessagingFactory.deploy()
+    await mockStarknetMessaging.deployed()
+
+    await starknet.devnet.loadL1MessagingContract(networkUrl, mockStarknetMessaging.address)
+
+    const testERC20Factory = await ethers.getContractFactory('TestERC20', deployer)
```
krebernisak: I'd like to see just the standard OZ ERC20 token here. 
I don't think a custom ERC20 is required or desired, as we want to test with the default one.

contracts/test/bridge/tokenBridge.ts
```diff
@@ -0,0 +1,184 @@
...
+    it('Test Set and Get function for L2 token address', async () => {
+      const new_data = ethers.utils.hexConcat([
+        ethers.utils.hexZeroPad(ethers.constants.AddressZero, 32),
+        ethers.utils.hexZeroPad(testERC20.address, 32),
+        ethers.utils.hexZeroPad(mockStarknetMessaging.address, 32),
+      ])
+
+      await proxy.connect(deployer).addImplementation(starkNetERC20Bridge.address, new_data, false)
+      await proxy.connect(deployer).upgradeTo(starkNetERC20Bridge.address, new_data, false)
+    })
```
krebernisak: What are we testing here exactly? Maybe you wanted to define a before hook here (that will point the proxy to the contract we're testing) and not the test (`it`).

contracts/test/bridge/tokenBridge.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    it('Should add implementation and upgrade successfully', async () => {
```
krebernisak: This test title makes no sense, considering the body.

contracts/test/bridge/tokenBridge.ts
```diff
@@ -0,0 +1,184 @@
...
+      await owner.invoke(tokenBridgeContract, 'set_l2_token', {
+        l2_token_address: linkTokenContract.address,
+      })
+      const { res: l2_address } = await tokenBridgeContract.call('get_l2_token', {})
+      expectAddressEquality(l2_address.toString(), linkTokenContract.address)
+    })
+
+    it('Should wrap contract and set L2 TokenBridge successfully', async () => {
+      const contract = await loadStarkgateContract('StarknetERC20Bridge')
+      newStarkNetERC20Bridge = await ethers.getContractAt(contract.abi, proxy.address)
```
krebernisak: Please define this at the beginning of the test, as part of the before hook or smth. It's used in multiple tests, so it makes sense to set it at the beginning - not just randomly in one of the tests first time you need to use it.

contracts/test/bridge/tokenBridge.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    it('Should setBalance to the token bridge', async () => {
+      await testERC20.setBalance(newStarkNetERC20Bridge.address, 1000)
+      const balance = await testERC20.balanceOf(newStarkNetERC20Bridge.address)
+      expect(balance).to.equal(1000)
+
+      await testERC20.setBalance(deployer.address, 10)
+      const balance2 = await testERC20.balanceOf(deployer.address)
+      expect(balance2).to.equal(10)
+    })
```
krebernisak: Why? If you need/want to transfer some funds to the L1 bridge contract, just use the available transfer method please.

contracts/test/bridge/tokenBridge.ts
```diff
@@ -0,0 +1,184 @@
...
+      const deposit = await newStarkNetERC20Bridge.maxDeposit()
+      expect(deposit).to.equal(100)
+    })
+
+    it('Should deposit to the L2 contract, L1 balance should be decreased by 2', async () => {
+      await testERC20.approve(newStarkNetERC20Bridge.address, 2)
+      await newStarkNetERC20Bridge.deposit(2, owner.starknetContract.address)
+
+      const balance = await testERC20.balanceOf(deployer.address)
+      expect(balance).to.equal(8)
```
krebernisak: Yeah, but we should also flush the bridge messages and assert that the L2 state changed (receiver got 2 deposited).

contracts/test/bridge/tokenBridge.ts
```diff
@@ -0,0 +1,184 @@
...
+    })
+  })
+
+  describe('Test bridge from L2 to L1', function () {
+    it('Should flush the L1 messages so that they can be consumed by the L2.', async () => {
```
krebernisak: Ok, so here is the test in another scope :D

Let's please move it to the L1 -> L2 scope/suite, and then here test the L2 -> L1 in a new scope/suite set of tests.

contracts/test/bridge/tokenBridge.ts
qperrot: Right I changed the content of the function but forgot the title

contracts/test/bridge/tokenBridge.ts
qperrot: The default one doesn't allow us to mint some token during constructor. That's why Starkgate has the TestERC20.sol to be able to set balance for the bridge. I think we can use this contract from OpenZeppelin https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/presets/ERC20PresetFixedSupply.sol

contracts/package.json
```diff
@@ -32,6 +32,7 @@
   "dependencies": {
     "@chainlink/contracts": "^0.4.2",
     "@toruslabs/starkware-crypto": "^1.0.0",
-    "axios": "^0.24.0"
+    "axios": "^0.24.0",
+    "internals-starkgate-contracts": "^0.0.3"
```
ivpavici: remove this, add to packages-ts

packages-ts/integration-starkgate/utils/index.ts
```diff
@@ -1,11 +1,31 @@
 import fs from 'fs'
-import { CompiledContract, json } from 'starknet'
+import { json } from 'starknet'
 
-export const loadStarkgateContract = (name: string): CompiledContract => {
+export const loadStarkgateContract = (name: string): any => {
```
ivpavici: maybe rename to `loadInternalStarkgateContract`

contracts/vendor/starkware-libs/starkgate-contracts/src/starkware/solidity/tokens/ERC20/ERC20.sol
ivpavici: Moved

contracts/test/bridge/tokenBridge.ts
qperrot: test Removed

contracts/test/bridge/tokenBridge.ts
ivpavici: Done

contracts/test/bridge/tokenBridge.ts
qperrot: Have to check again but I think so

contracts/test/token/starkgate/behavior/ERC20.ts
```diff
@@ -119,6 +126,7 @@ export const shouldBehaveLikeStarkGateERC20 = (beforeFn: BeforeFn) => {
         })
         throw new Error('This should not pass!')
       } catch (error: any) {
+        console.log(error.message)
```
krebernisak:
```suggestion
```

packages-ts/integration-starkgate/utils/index.ts
```diff
@@ -1,11 +1,31 @@
...
+      )
+      .toString('ascii'),
+  )
+}
+
+export const loadSolidityContract = (name: string): any => {
+  return json.parse(
+    fs
+      .readFileSync(
+        `${__dirname}/../../../contracts/artifacts/src/chainlink/solidity/mocks/${name}.sol/${name}.json`,
```
krebernisak: It's slightly unbelievable that we need to load contracts like this - very brittle.

For this PR please find all instances of functions like this (ctrl + F 'readFileSync') and please rename all these load functions to start with `loadContract_` prefix so they are easily discoverable.

We'll need to go in as as a separate effort and figure out proper contract loading for both Cairo and Solidity.

https://github.com/NomicFoundation/hardhat/issues/1040

contracts/package.json
krebernisak: Yeah, should not be required here anymore.

packages-ts/integration-starkgate/utils/index.ts
krebernisak: Opened a GH Issue on this:

- https://github.com/smartcontractkit/chainlink-starknet/issues/96

contracts/test/token/starkgate/behavior/ERC20.ts
```diff
@@ -3,6 +3,7 @@ import { StarknetContract, Account } from 'hardhat/types/runtime'
 import { uint256 } from 'starknet'
 import { toBN } from 'starknet/utils/number'
 import { TIMEOUT } from '../../../constants'
+import { assertErrorMsg } from '../../../utils'
```
krebernisak: The tests are failing, check the CI: 

```
An unexpected error occurred:

test/token/starkgate/behavior/ERC20.ts(6,10): error TS2305: Module '"../../../utils"' has no exported member 'assertErrorMsg'.
```

packages-ts/integration-starkgate/test/tokenBridge.ts
```diff
@@ -1,57 +1,184 @@
-import { starknet } from 'hardhat'
-import { expect } from 'chai'
-import { StarknetContract, ArgentAccount } from 'hardhat/types/runtime'
 import { TIMEOUT } from './constants'
```
krebernisak: Please rename the file as `tokenBridge.test.ts`

contracts/test/token/starkgate/behavior/ERC20.ts
qperrot: I don't know why this happen

ivpavici: note: it doesn't fail locally!

Makefile
```diff
@@ -157,8 +159,55 @@ test-unit-go:
...
+	cd packages-ts/starknet-gauntlet-example/ && \
+		yarn test
+	cd packages-ts/starknet-gauntlet-multisig/ && \
+		yarn test
+	cd packages-ts/starknet-gauntlet-ocr2/ && \
+		yarn test
+	cd packages-ts/starknet-gauntlet-oz/ && \
+		yarn test
+	cd packages-ts/starknet-gauntlet-starkgate/ && \
+		yarn test
```
archseer, line 196: Why not rely on the toplevel `yarn test` which uses the Jest config to run all the tests? I personally only use that and ignore the `Makefile`

contracts/package.json
```diff
@@ -15,23 +15,17 @@
...
+    "@types/mocha": "^9.1.1",
     "cairo-ls": "^0.0.4",
-    "chai": "^4.3.4",
+    "chai": "^4.3.6",
     "ethereum-waffle": "^3.4.4",
     "ethers": "^5.6.8",
-    "hardhat": "^2.10.2",
-    "starknet": "3.9.0",
-    "ts-node": "^10.4.0",
-    "typescript": "^4.5.2"
```
archseer, line 30: Aren't these required for tests?

krebernisak: They are declared as part of `<root>/package.json` (just once, single version) and used by the whole yarn workspace. No need to redeclare them here, if a custom version is not required.

Makefile
krebernisak: The version before this PR just ran Jest tests via root `package.json` test script.

The problem was not all workspace packages use Jest for testing - our contracts, example project, and 2x integration projects (multisig, starkgate) use hardhat + mocha, so they wouldn't be run.

Then I updated the root `package.json` test script to iterate over all projects and run their test scripts via `"yarn workspaces run test"` - but now the problem is Hardhat setup tests expect external network to be started already, and Gauntlet Jest tests run their own network (Integrated Devnet) on `beforeAll`. This was now a conflict because Integrated Devnet wouldn't start successfully if docker container was already running on the same 5050 port.

So for now I've split the two category of tests as different target, and different CI workflows. Now there is no infra conflict, and all tests are finally ran.

This is enough for this PR I'd say, and the plan for the next one is to restructure how do we run local env in tests as dependencies. Every tests suite should start its own network programatically in test setup, and tear it down when finished. At that point we can just run all tests at once via `"yarn workspaces run test"` and they would be even more isolated.

krebernisak: tldr; I consider this a temp workaround, and will try to remove it as test runs are restructured to be idempotent.

contracts/test/token/starkgate/behavior/ERC20.ts
```diff
@@ -3,18 +3,38 @@ import { StarknetContract, Account } from 'hardhat/types/runtime'
 import { uint256 } from 'starknet'
 import { toBN } from 'starknet/utils/number'
 import { TIMEOUT } from '../../../constants'
+import { expectInvokeError } from '../../../../test/utils'
```
qperrot, line 6: Maybe We can write it back to `import { expectInvokeError } from '../../../utils'`

Makefile
krebernisak: Extracted a GH Issue:

- https://github.com/smartcontractkit/chainlink-starknet/issues/101

contracts/test/utils.ts
```diff
@@ -1,38 +1,58 @@
+import { constants, ec, encode, hash, number, uint256, stark, KeyPair } from 'starknet'
```
qperrot, line 1: Don't need all of it we can change it to `import { constants, encode, number} from 'starknet'`

krebernisak (approved):

PR #97, opened 2022-09-05
add l2 emergency protocol docs

aiagentdev-1990:


docs/emergency-protocol/README.md
```diff
@@ -0,0 +1,112 @@
...
+     undefined,
+     { maxFee: '0' },
+   )
+   ```
+
+3. As the above transaction is expected to fail, the EA will consider the Sequencer as healthy if it receives any of the expected error statuses
+   1. `StarknetErrorCode.UNINITIALIZED_CONTRACT` if the dummy contract has not been initialized
+   2. `StarknetErrorCode.OUT_OF_RANGE_FEE` if the dummy contract has been initialized by accident. As Starknet is a permissionless network, we cannot guarantee that a user deploys and initializes a contract at the dummy address. As a result, the EA will set the `maxFee` to 0 so that the transaction will fail with the `StarknetErrorCode.OUT_OF_RANGE_FEE` status code.
+
+## **Deployed Addresses**
```
krebernisak: Let's remove this section, these are our internal test deployments that are probably not too useful to be added to the docs.

docs/emergency-protocol/README.md
```diff
@@ -0,0 +1,112 @@
+# Starknet Emergency Protocol
+
+## Background
+
+The purpose of the Starknet Sequencer Emergency Protocol is to track the last known status of the Starknet Sequencer at a given point in time.
+
```
krebernisak, line 10: Let's add a `## Contracts` section and list all relevant contracts (link to their source) that are part of the protocol.

docs/emergency-protocol/README.md
```diff
@@ -0,0 +1,112 @@
...
+## Architecture
+
+The diagram above illustrates the general path of how the Sequencer’s status is relayed from L1 to L2.
+
+![L2 Emergency Protocol Diagram](./Starknet-L2EP.png)
```
krebernisak: Let's please replace these static assets with code generated graphs - Mermaid, which is supported by GitHub. That way these visualizations will be a lot easier to support and upgrade.

- Editor: https://mermaid.live/edit
- Docs: https://mermaid-js.github.io/mermaid/#/

docs/emergency-protocol/README.md
```diff
@@ -0,0 +1,112 @@
...
+![Starknet Sequencer Pending](.//Starknet-Pending.png)
+
+**After Sequencer comes back online**
+
+- `LogMessageToL2` events are picked up and added to the pending queue.
+- Transactions in the pending queue are processed chronologically so **Tx1** is processed before **Tx2.**
+- As **Tx1** happens before **Tx2, Tx2** will read the status of the Sequencer as being down
+
+### Layer2 Sequencer Health External Adapter
+
```
krebernisak, line 96: Please link to the EA source code.

docs/emergency-protocol/README.md
```diff
@@ -0,0 +1,112 @@
+# Starknet Emergency Protocol
+
+## Background
+
+The purpose of the Starknet Sequencer Emergency Protocol is to track the last known status of the Starknet Sequencer at a given point in time.
+
+## Architecture
+
```
krebernisak, line 12: Let's please add `### Bridge Fees` section, note that it's WIP (currently static) but we plan to support dynamic fees either via an L1 gas price oracle (Chainlink or StarkNet).

krebernisak: Let's also add an example how these fees could be estimated via the CLI:

```bash
starknet estimate_message_fee \
  --feeder_gateway_url=https://alpha4.starknet.io/feeder_gateway/ \
  --abi=contracts/starknet-artifacts/src/chainlink/cairo/emergency/SequencerUptimeFeed/sequencer_uptime_feed.cairo/sequencer_uptime_feed.json \
  --from_address ${FROM} \
  --address ${TO} \
  --function update_status \
  --inputs ${STATUS} ${TIMESTAMP}
```

docs/emergency-protocol/README.md
```diff
@@ -0,0 +1,112 @@
+# Starknet Emergency Protocol
+
+## Background
+
+The purpose of the Starknet Sequencer Emergency Protocol is to track the last known status of the Starknet Sequencer at a given point in time.
```
krebernisak: Maybe expand the why part, as in why do we do this, and link to official docs: https://docs.chain.link/docs/l2-sequencer-flag/

krebernisak (approved):
Thanks @cds95!

Let's please remove the .png static assets and then squash-merge the PR so these are excluded from our git history.

PR #99, opened 2022-09-05
Try to fix i192 values (currently over 2^128 fails to transmit)

archseer:


archseer:
Settled on i128 for now

PR #102, opened 2022-09-05
rename event to CamelCase and delete unused import

qperrot:


contracts/src/chainlink/cairo/ocr2/IAggregator.cairo
```diff
@@ -8,6 +8,13 @@ struct Round:
     member updated_at : felt
 end
 
+struct Transmission:
```
krebernisak: This is an internal struct (not part of the interface) that's why I wouldn't move it to the interface.

What actually makes sense here is to split the `aggregator.cairo` to the `Aggregator/library.cairo` lib + the contract implementation itself (uses the lib). Then the Transmission struct could be part of the library and reused by other contracts (Mock for example).

Don't do this as part of this PR. Let's just move it back for now.

qperrot: ok

contracts/src/chainlink/cairo/ocr2/aggregator.cairo
```diff
@@ -210,7 +203,7 @@ end
 # --- Configuration
 
 @event
-func config_set(
+func ConfigSet(
```
krebernisak, line 258: Be careful to search the whole project. There are more instances where events are used in the relayer for example:

- https://github.com/smartcontractkit/chainlink-starknet/search?q=config_set
- https://github.com/smartcontractkit/chainlink-starknet/search?q=new_transmission

archseer (approved):

PR #103, opened 2022-09-05
Monitoring

topliceanu:


relayer/ops/monitoring/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+ENV PATH /usr/local/go/bin:$PATH
+
+# Compile binary
+
+WORKDIR /starknet-monitoring/relayer
+RUN go build -o ./monitoring ./cmd/monitoring/*.go
+
+# Production image
+
+FROM ubuntu:20.04
```
pedrovsg: I would recommend alpine

relayer/pkg/chainlink/ocr2/client.go
```diff
@@ -143,27 +167,48 @@ func (c *Client) ConfigFromEventAt(ctx context.Context, address string, blockNum
...
-				if err != nil {
-					return cc, errors.Wrap(err, "couldn't parse config event")
-				}
-
-				return ContractConfig{
-					Config:      config,
-					ConfigBlock: blockNum,
-				}, nil
+			if starknet.IsEventFromContract(&decodedEvent, address, eventType) {
+				eventsAsFeltArrs = append(eventsAsFeltArrs, decodedEvent.Data)
```
sdrug, line 200: Could you please explain the purpose of this change?
I think it's guaranteed to have one or none config event per block, because otherwise we would have a design issue  since we refer to config by block number.

relayer/pkg/chainlink/ocr2/types.go
```diff
@@ -57,3 +59,35 @@ func NewBillingDetails(observationPaymentFelt junotypes.Felt, transmissionPaymen
...
+	RoundID     uint32
+	Answer      *big.Int
+	BlockNumber uint64
+	StartedAt   time.Time
+	UpdatedAt   time.Time
+}
+
+func NewRoundData(felts []junotypes.Felt) (data RoundData, err error) {
+	if len(felts) < 5 {
+		return data, fmt.Errorf("expected number of felts to be 5 but got %d", len(felts))
```
sdrug, line 73: Error description doesn't correspond to condition?

relayer/pkg/monitoring/config_chain.go
```diff
@@ -0,0 +1,112 @@
...
+	NetworkID        string
+	ChainID          string
+	ReadTimeout      time.Duration
+	PollInterval     time.Duration
+	LinkTokenAddress string
+}
+
+var _ relayMonitoring.ChainConfig = StarknetConfig{}
+
+func (s StarknetConfig) GetRPCEndpoint() string         { return s.RPCEndpoint }
```
sdrug: Do we need getters for public properties?

relayer/pkg/monitoring/source_envelope.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return config, fmt.Errorf("couldn't fetch config at block '%d' for contract '%s': %w", configDetails.Block, contractAddress, err)
+	}
+	return config, nil
+}
+
+func (s *envelopeSource) fetchLinkAvailableForPayment(ctx context.Context, contractAddress string) (*big.Int, error) {
+	results, err := s.ocr2Reader.BaseReader().CallContract(ctx, starknet.CallOps{
+		ContractAddress: contractAddress,
+		Selector:        "link_available_for_payments",
```
sdrug: Maybe add a corresponding method to base starknet or ocr2 client?

relayer/ops/monitoring/Dockerfile
topliceanu: I agree. I added a TODO to update this at a later point. It shouldn't require any other changes. I'll use this https://github.com/smartcontractkit/atlas/blob/master/evm/Dockerfile as an example. Wdyt?

relayer/pkg/chainlink/ocr2/client.go
topliceanu: Sure. Here's my thinking:

`ConfigFromEventAt` is doing three things from my pov: 
1. fetches a block
2. filters events from that block by event type (eg. `config_set`) and source contract address.
3. stops at the first config_set event found, decodes it and returns it

What I want is a private generic function to do 1. and 2. Then I can use it to parse other types of event like `new_transmission`. 

> I think it's guaranteed to have one or none config event per block

Sure, but returning an array is more generic for the utility function I have in mind and would accommodate other event types.

relayer/pkg/monitoring/config_chain.go
topliceanu: Note [line 22](https://github.com/smartcontractkit/chainlink-starknet/pull/103/files#diff-cb132ac733f7b2819c72de47842ec0baccbb536abbabaa8fe7c45210563a1d49R22): this struct needs to implement [ChainConfig from chainlink-relay/pkg/monitoring](https://github.com/smartcontractkit/chainlink-relay/blob/373aac020fc861ac3f8115b50225376a10270f42/pkg/monitoring/chain_config.go#L5) I agree the design is not ideal. I subsequently realized that using type assertions is simpler to use - despite less safety.

relayer/pkg/monitoring/source_envelope.go
topliceanu: That's a good idea! I feel it's place is in ocr2.Client. Wdyt?

relayer/ops/monitoring/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+RUN wget -c https://dl.google.com/go/go1.18.1.linux-amd64.tar.gz -O - \
+  | tar -xz -C /usr/local \
+  && mkdir -p /go/src /go/bin
+ENV PATH /usr/local/go/bin:$PATH
+
+# Compile binary
+
+WORKDIR /starknet-monitoring/relayer
+RUN go build -o ./monitoring ./cmd/monitoring/*.go
```
archseer: Since all this is doing is just compiling go code, why not use the golang build image like core does?

https://github.com/smartcontractkit/chainlink/blob/b64fe315a2e7af10244f737c9acd7bd3c75bbeef/core/chainlink.Dockerfile#L30

archseer (dismissed):

relayer/ops/monitoring/Dockerfile
archseer: I wonder if you need `alpine` and `scratch` could be enough?

As long as you

```
# copy ca certs
COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 
```

relayer/pkg/chainlink/ocr2/client.go
sdrug: Alright. Let's maybe find a better name for it, something like `fetchEventsFromBlock`?
And I would leave a comment at places where you use first array element (or even throw an error if there are more than one event) for better readability.

relayer/pkg/monitoring/config_chain.go
sdrug: Maybe then convert them all into private properties with public getters?
We have mixed coding styles across the stack (private properties with public getters vs public properties). I wonder what the idiomatic go vision on that. I'd bet on public properties since it's less code.
Nit for now, just to keep in mind for future extensions.

relayer/pkg/monitoring/source_envelope.go
sdrug: Yeah, I think it belongs there.

relayer/pkg/chainlink/ocr2/client.go
topliceanu: Excellent ideas!

topliceanu: > or even throw an error if there are more than one event) for better readability.   

This is also true for new_transmission events, right? There shouldn't be more than one from an aggregator in a block

topliceanu: Done in https://github.com/smartcontractkit/chainlink-starknet/pull/103/commits/53ab00621edbc60c4a57d000bf701057eba0133c

relayer/pkg/monitoring/config_chain.go
topliceanu: A bit more context here. This `StarknetConfig` struct contains chain-level configuration for the monitor. In chainlink-solana there is a SolanaConfig. These two share some fields but also have chain-specific fields. 

The fields that are shared are also needed/required by the [core on-chain monitoring library](https://github.com/smartcontractkit/chainlink-relay/tree/main/pkg/monitoring), that's why they are encoded as methods of an interface. 

I agree with you that making these fields public with also public getters is strange!

topliceanu: > Maybe then convert them all into private properties with public getters?

I will do this!

topliceanu: Done in https://github.com/smartcontractkit/chainlink-starknet/pull/103/commits/0a5ebae47b94e10bde3b97b9e3686d315764d139

topliceanu:
@archseer you marked your review as "requesting changes". What is exactly that you want me to change? the Dockerfile?

archseer:
Yeah it should be a small change that speeds up the build.

relayer/ops/monitoring/Dockerfile
```diff
@@ -0,0 +1,36 @@
+# Build image
```
aalu1418: just a general thought on the location of this file: would it better to be in the `chainlink-starknet/ops` folder rather than a new `relayer/ops` folder?

relayer/go.mod
```diff
@@ -7,34 +7,49 @@ require (
...
+	github.com/linkedin/goavro/v2 v2.9.7 // indirect
 	github.com/mattn/go-colorable v0.1.12 // indirect
 	github.com/mattn/go-isatty v0.0.14 // indirect
+	github.com/matttproud/golang_protobuf_extensions v1.0.1 // indirect
 	github.com/mr-tron/base58 v1.2.0 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
+	github.com/prometheus/client_model v0.2.0 // indirect
+	github.com/prometheus/common v0.35.0 // indirect
+	github.com/prometheus/procfs v0.7.3 // indirect
+	github.com/riferrei/srclient v0.4.1-0.20211229125508-8edc580da179 // indirect
```
aalu1418: just some more thinking about how to structure monitoring components: maybe it's better to have a separate monitoring module outside of `relayer`?

pros:
* reduce the number of dependencies within the relayer module

cons:
* some dependency weirdness similar to how `ops` => `relayer` => `e2e_tests`, so then it would be `relayer` is a module used in `monitoring`

relayer/pkg/chainlink/ocr2/client.go
```diff
@@ -18,7 +20,10 @@ import (
 type OCR2Reader interface {
 	LatestConfigDetails(context.Context, string) (ContractConfigDetails, error)
 	LatestTransmissionDetails(context.Context, string) (TransmissionDetails, error)
+	LatestRoundData(context.Context, string) (RoundData, error)
+	LinkAvailableForPayment(context.Context, string) (*big.Int, error)
 	ConfigFromEventAt(context.Context, string, uint64) (ContractConfig, error)
+	NewTransmissionEventAt(context.Context, string, uint64) (NewTransmissionEvent, error)
```
aalu1418: test case coverage?

relayer/pkg/monitoring/config_feed.go
```diff
@@ -0,0 +1,99 @@
...
+	Symbol         string   `json:"symbol,omitempty"`
+	HeartbeatSec   int64    `json:"heartbeat,omitempty"`
+	ContractType   string   `json:"contract_type,omitempty"`
+	ContractStatus string   `json:"status,omitempty"`
+	MultiplyRaw    string   `json:"multiply,omitempty"`
+	Multiply       *big.Int `json:"-"`
+
+	ContractAddress string `json:"contract_address,omitempty"`
+	ProxyAddress    string `json:"proxy_address,omitempty"`
+}
```
aalu1418: just wondering, this feels pretty generic to most feed structures. is it?
if so, can something like this be moved to `chainlink-relay`?

relayer/go.mod
topliceanu: This can be done. It would be easy to do here, but it would have to be done in chainlink-solana for consistency. Over there, it's harder because the main go module is in the root of the project.

relayer/pkg/monitoring/config_feed.go
topliceanu: It is generic! But also there are chain-specific field. 
The "generic" part is behind an interface - see line 26 below

relayer/go.mod
aalu1418: makes sense! my thinking is that we use starknet implementation as a way to pilot the best layout of files/components that we would want to see in the other existing integrations like solana. we have some future planning to refactor the solana (and terra?) repos so we would propagate the best practices from starknet to the other chains

relayer/ops/monitoring/Dockerfile
topliceanu: done

relayer/go.mod
topliceanu: I agree. I moved everything in `/monitoring`. See https://github.com/smartcontractkit/chainlink-starknet/pull/103/commits/7c822bc7787c845f9270648f525843ab457bd9f6

relayer/pkg/chainlink/ocr2/client.go
topliceanu: added!

topliceanu:
@archseer and @pedrovsg I tried to update the build image to golang-buster and the prod image to alpine, but I didn't get it to work. Let's merge this PR without it and I'll update the image in a followup PR. Wdyt?

relayer/ops/monitoring/Dockerfile
topliceanu: I'll do this in a separate PR. Wdyt?

relayer/ops/monitoring/Dockerfile
topliceanu: I'll do this in a separate PR. Wdyt?

monitoring/go.mod
```diff
@@ -0,0 +1,61 @@
...
+	golang.org/x/crypto v0.0.0-20220525230936-793ad666bf5e // indirect
+	golang.org/x/exp v0.0.0-20220608143224-64259d1afd70 // indirect
+	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c // indirect
+	golang.org/x/sys v0.0.0-20220520151302-bc2c85ada10a // indirect
+	google.golang.org/protobuf v1.28.0 // indirect
+	gopkg.in/natefinch/npipe.v2 v2.0.0-20160621034901-c1b8fa8bdcce // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+)
+
+replace github.com/smartcontractkit/chainlink-starknet/relayer => ../relayer
```
aalu1418: we've tried to avoid using `replace` in the `go.mod` files

instead directly importing specific versions (just means we need to be careful about squash merges)
like the integration-tests import `ops` and `relayer` but without `replace`

relayer/pkg/chainlink/ocr2/types_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"math/big"
+	"testing"
+	"time"
+
+	junotypes "github.com/NethermindEth/juno/pkg/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewRoundData(t *testing.T) {
+	// TODO (dru) add a snapshot of the response.
```
aalu1418: is this TODO resolved?

relayer/pkg/chainlink/ocr2/types_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+// Helpers
+
+func rawToFelts(raw []string) []junotypes.Felt {
+	out := make([]junotypes.Felt, len(raw))
+	for i := 0; i < len(raw); i++ {
+		out[i] = junotypes.HexToFelt(raw[i])
+	}
+	return out
+}
```
aalu1418: this feels very similar? https://github.com/smartcontractkit/chainlink-starknet/blob/develop/relayer/pkg/starknet/utils.go#L123

or maybe this would be better in the utils file

monitoring/ops/Dockerfile
```diff
@@ -0,0 +1,36 @@
+# Build image
+
+FROM ubuntu:20.04 AS build
```
aalu1418, line 3: I think you mentioned that this would be handled in a follow-up PR - could you make an issue to capture that?

aalu1418 (commented):
looking good! just small things :)

monitoring/go.mod
topliceanu: I'm not too sure how this would work. I depend on a version of the relayer from this branch. Maybe after I merge this to master I can update to a commit hash from master?

relayer/pkg/chainlink/ocr2/types_test.go
topliceanu: Yes

relayer/pkg/chainlink/ocr2/types_test.go
topliceanu: Done in b90403e

monitoring/ops/Dockerfile
topliceanu: Yeah, I'll let o11y know

monitoring/go.mod
topliceanu: Done in 9126230

monitoring/ops/Dockerfile
aalu1418: added a github issue to track this: https://github.com/smartcontractkit/chainlink-starknet/issues/145

aalu1418 (approved):
looks good to me! happy to approve knowing there will be follow up PRs :)

**NOTE: this PR should not be squashed - commit history needs to be preserved**

PR #107, opened 2022-09-06
Set fee when sending cross chain validator message

aiagentdev-1990:


contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
```diff
@@ -82,7 +100,18 @@ contract StarkNetValidator is TypeAndVersionInterface, AggregatorValidatorInterf
...
+    );
     return true;
   }
+
+  /**
+   * @notice makes this contract payable
+   * @dev receives funds:
+   *  - funds are used to send cross chain messages to L2
+   */
+  receive() external payable {}
```
krebernisak, line 260: We also need the ability for the owner to withdraw funds from this contract, for example when we migrate to a different contract for whatever reason.

Also we need to investigate how message cancelations work, and do we require any additional logic in this contract except the `receive` fn to be refunded: https://docs.starknet.io/docs/L1-L2%20Communication/messaging-mechanism#l1--l2-message-fees

> Throughtout the lifecycle of the message, the assoicated fee is locked in the Core contract. If the message was not processed for any reason, then [cancelling](https://docs.starknet.io/docs/L1-L2%20Communication/messaging-mechanism#l1--l2-message-cancellation) the message will also return the fee to the sender.

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
```diff
@@ -63,6 +73,14 @@ contract StarkNetValidator is TypeAndVersionInterface, AggregatorValidatorInterf
     return "StarkNetValidator 0.1.0";
   }
 
+  function setFee(uint256 newFee) external onlyOwner {
```
krebernisak: https://docs.starknet.io/docs/L1-L2%20Communication/messaging-mechanism#l1--l2-message-fees

>  The sequencer charges the fee in full upon updating the L1 state with the consumption of this message.

Static fees will not work for us I'm afraid as it looks like the fee will be taken in full. On Arbitrum the static fee is an option and it currently works that way, as we can overpay for the message send and get the excess refunded.

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
krebernisak: I've added the withdrawal functionality.

We still need to investigate message cancelations.

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
```diff
@@ -81,8 +96,39 @@ contract StarkNetValidator is TypeAndVersionInterface, AggregatorValidatorInterf
     payload[0] = toUInt256(status);
     payload[1] = block.timestamp;
     // Make the StarkNet x-domain call.
-    // NOTICE: we ignore the output of this call (msgHash, nonce), and we don't raise any events as the event LogMessageToL2 will be emitted from the messaging contract
-    STARKNET_CROSS_DOMAIN_MESSENGER.sendMessageToL2(L2_UPTIME_FEED_ADDR, STARK_SELECTOR_UPDATE_STATUS, payload);
+    // NOTICE: we ignore the output of this call (msgHash, nonce).
+    // We also don't raise any events as the 'LogMessageToL2' event will be emitted from the messaging contract.
+    STARKNET_CROSS_DOMAIN_MESSENGER.sendMessageToL2{value: s_fee}(
```
krebernisak: Let's add a specific internal function that computes the required fee, based on the proposed workflow commented above.

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
krebernisak: Ok, here is a strategy I think we should try.

Let's try to express the bridge fee as a function of the L1 gas price: https://data.chain.link/ethereum/mainnet/gas/fast-gas-gwei

We should be able to set and configure:

- L1 gas price feed address
- L2 tx gas estimate, which we can figure out empirically via https://starknet.io/docs/hello_starknet/l1l2.html#estimate-message-fee

When configuring the L2 tx gas estimate, we can add some buffer on top to be able to mitigate more volatility.

Additionally, I think we should add the **ability for the owner to retry** the last transmit/validate/send, if for whatever reason the L1 -> L2 message doesn't go through. To implement this I think we should store the L1 feed that we validate, and read the last round from there on retry (onlyOwner).

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
```diff
@@ -6,34 +6,39 @@ import "@chainlink/contracts/src/v0.8/interfaces/TypeAndVersionInterface.sol";
...
-/**
- * @title StarkNetValidator - makes cross chain call to update the Sequencer Uptime Feed on L2
- */
+/// @title StarkNetValidator - makes cross chain calls to update the Sequencer Uptime Feed on L2
 contract StarkNetValidator is TypeAndVersionInterface, AggregatorValidatorInterface, SimpleWriteAccessController {
   int256 private constant ANSWER_SEQ_OFFLINE = 1;
-  // Selector hardcoded because StarkNet hash function is not available in this environment
+  // The selector is the starknet_keccak hash of the function name 'update_status'.
+  // Notice: hardcoded b/c starknet_keccak is not available in this environment.
   uint256 constant STARK_SELECTOR_UPDATE_STATUS =
```
krebernisak: We could actually compute this in-place (easier to understand and update) by using `keccak256` and cleaning the high 6 bits: https://docs.starknet.io/docs/Hashing/hash-functions#starknet-keccak

> StarkNet keccak, usually denoted by sn\_keccaksn_keccak, is defined as the first 250 bits of the Keccak256 hash (this is just Keccak256 augmented in order to fit into a field element).

aiagentdev-1990: Wait I'm a bit confused.  I had left a comment in an earlier PR about hardcoding this and was told that we have to hardcode it here.  Is this no longer the case?

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
aiagentdev-1990: We can cancel messages by calling the `cancelL1ToL2Message` on the `Starknet` contract.

https://goerli.etherscan.io/address/0xde29d060D45901Fb19ED6C6e959EB22d8626708e#writeProxyContract.

The tricky thing about this is that we'd have to store the nonce and status in our contract before we send the cross chain message.  This might get expensive.

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
krebernisak: We can either hard code, or compute it by slightly modifying the output of Solidity `keccak256`.

Hardcode is fine, computing it is better - some small cost on construction, but more readable core easier to support. Not a big priority, just an improvement proposal for when we solve the main open items.

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
```diff
@@ -74,15 +125,94 @@ contract StarkNetValidator is TypeAndVersionInterface, AggregatorValidatorInterf
...
+    }
+    _;
+  }
+
+  /**
+   * @notice makes this contract payable
+   * @dev receives funds:
+   *  - funds are used to send cross chain messages to L2
+   */
+  receive() external payable {}
```
krebernisak, line 260: Where did the withdrawal functionality I added [here](https://github.com/smartcontractkit/chainlink-starknet/pull/107/commits/9620ecb6308435c88e0711b7667a5a945099ee14) go?

krebernisak: I've added it back.

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
```diff
@@ -74,15 +122,152 @@ contract StarkNetValidator is TypeAndVersionInterface, AggregatorValidatorInterf
...
+      SELECTOR_STARK_UPDATE_STATUS,
+      payload
+    );
     return true;
   }
+
+  /// @notice L1 oracle is asked for a fast L1 gas price, and the price multiplied by the configured gas estimate
+  function _approximateFee() internal view returns (uint256) {
+    (, int256 fastGasPriceInWei, , , ) = AggregatorV3Interface(s_gasConfig.gasPriceL1Feed).latestRoundData();
+    return (uint256(fastGasPriceInWei) * s_gasConfig.gasUsed * s_gasConfig.buffer);
```
krebernisak: @cds95 I've removed the `/ 1e18` multiplier here. Shouldn't we send the fee in wei?

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
```diff
@@ -6,44 +6,82 @@ import "@chainlink/contracts/src/v0.8/interfaces/TypeAndVersionInterface.sol";
...
    */
-  constructor(address starkNetMessaging, uint256 l2UptimeFeedAddr) {
+  constructor(
+    address starkNetMessaging,
+    address configAC,
+    address gasPriceL1Feed,
+    address source,
+    uint256 l2Feed,
+    uint256 gasUsed,
+    uint256 buffer
```
krebernisak: I'd simplify this and just use a single value `s_gasConfig.gasEstimate`, which can include the buffer so no need for an extra config:

```suggestion
    uint256 gasEstimate,
```

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
krebernisak: If we ever need to do that we can read that data from the raised bridge events, and invoke it manually.

I assume that invocation will refund the funds to the contract, that's the only thing that makes sense to do.

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
```diff
@@ -74,15 +122,152 @@ contract StarkNetValidator is TypeAndVersionInterface, AggregatorValidatorInterf
...
-    bool status = currentAnswer == ANSWER_SEQ_OFFLINE;
-    uint256[] memory payload = new uint256[](2);
+    return _sendUpdateMessageToL2(currentAnswer);
+  }
+
+  /**
+   * @notice retries to send the latest answer as update message to L2
+   * @dev only with access, useful in cases where a previous x-domain message was handeled unsuccessfully.
+   */
+  function retry() external checkAccess returns (bool) {
```
krebernisak, line 130: @cds95 I've changed this to `checkAccess` modifier here instead of `onlyOwner

I think this is more correct, as the owner could already add access to another party who could then retry by calling the `validate` fn manually. Here I suggest we lower the restriction to retry.

Let's say there is a use case to automate retries, this is also a required change to support those use cases.

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
aiagentdev-1990: ah this must have gone away when I was resolving conflicts.

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
```diff
@@ -74,15 +120,144 @@ contract StarkNetValidator is TypeAndVersionInterface, AggregatorValidatorInterf
...
+   * @notice sets the source aggregator AggregatorInterface contract
+   * @dev only owner can call this
+   * @param source the source aggregator that we'll read data from (on retries)
+   */
+  function setSourceAggregator(address source) external onlyOwner {
+    _setSourceAggregator(source);
+  }
+
+  /// @notice Internal method that sets the source aggregator AggregatorInterface contract
+  function _setSourceAggregator(address source) internal {
```
aiagentdev-1990, line 238: @krebernisak I wonder if we should follow the `propose` and `upgrade` pattern here.

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
```diff
@@ -74,15 +120,137 @@ contract StarkNetValidator is TypeAndVersionInterface, AggregatorValidatorInterf
...
+    address previousSource = address(s_source);
+    if (source != previousSource) {
+      s_source = AggregatorV3Interface(source);
```
aiagentdev-1990, line 244: I removed the part about adding and removing access as the aggregator never writes to the Validator so it doesn't need any access.  Calling `this.addAccess` and `this.removeAccess` also does not work when we try to initialize the aggregator in the constructor.

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
```diff
@@ -74,15 +116,158 @@ contract StarkNetValidator is TypeAndVersionInterface, AggregatorValidatorInterf
...
+      SELECTOR_STARK_UPDATE_STATUS,
+      payload
+    );
     return true;
   }
+
+  /// @notice L1 oracle is asked for a fast L1 gas price, and the price multiplied by the configured gas estimate
+  function _approximateFee() internal view returns (uint256) {
+    (, int256 fastGasPriceInWei, , , ) = AggregatorV3Interface(s_gasConfig.gasPriceL1Feed).latestRoundData();
+    return (uint256(fastGasPriceInWei) * s_gasConfig.gasEstimate);
```
krebernisak:
```suggestion
    return uint256(fastGasPriceInWei) * s_gasConfig.gasEstimate;
```

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
krebernisak: > as the aggregator never writes to the Validator so it doesn't need any access

Not exactly true, as `ValidatorProxy` infra is optional. But we do use it for most deployments. This would potentially be useful for simple test deployments.

Ok, let's leave it like this.

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
krebernisak: If we don't do it with config AC or the config itself, no need to do it for this operation.

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
```diff
@@ -6,44 +6,76 @@ import "@chainlink/contracts/src/v0.8/interfaces/TypeAndVersionInterface.sol";
...
+  error InvalidGasPriceL1FeedAddress();
+  /// @notice Error thrown when caller is not the owner and does not have access
+  error AccessForbidden();
+
+  /// @notice This event is emitted when the gas config is set.
+  event GasConfigSet(uint256 gasEstimate, address indexed gasPriceL1Feed);
+  /// @notice emitted when a new gas access-control contract is set
+  event ConfigACSet(address indexed previous, address indexed current);
+  /// @notice emitted when a new source aggregator contract is set
+  event SourceAggregatorSet(address indexed previous, address indexed current);
```
krebernisak: This one is unused?

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
```diff
@@ -6,44 +6,76 @@ import "@chainlink/contracts/src/v0.8/interfaces/TypeAndVersionInterface.sol";
...
 
-  IStarknetMessaging public immutable STARKNET_CROSS_DOMAIN_MESSENGER;
+  uint256 public immutable SELECTOR_STARK_UPDATE_STATUS = _selectorStarkNet("update_status");
   uint256 public immutable L2_UPTIME_FEED_ADDR;
 
-  /// @notice StarkNet messaging contract address - the address is 0.
-  error InvalidStarkNetMessaging();
+  IStarknetMessaging public immutable STARKNET_CROSS_DOMAIN_MESSENGER;
 
+  AggregatorV3Interface private s_source;
```
krebernisak, line 27: We should add a getter for this.

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
krebernisak: Please add the in-line docs back, somewhere maybe here:

```
// NOTICE: we don't give access to the new source aggregator
// It is not always the case that the source aggregator is also the sender for the 'validate' invocation
// as we usually deploy an additional proxy in between (owner can give access):
//   https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.8/ValidatorProxy.sol
```

It's helpful to document these things for anyone reading the code, and us once we'll need to go back and work on access control improvements for these contracts.

contracts/test/emergency/StarkNetValidator.test.ts
```diff
@@ -52,24 +75,281 @@ describe('StarkNetValidator', function () {
...
+      await expect(
+        factory.deploy(
+          mockStarkNetMessaging.address,
+          mockAccessController.address,
+          mockGasPriceFeed.address,
+          mockAggregator.address,
+          0,
+          0,
+        ),
+      ).to.be.revertedWith('InvalidL2FeedAddress()')
```
krebernisak: How come we're not using this custom matcher for custom errors?
https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-chai-matchers/src/internal/reverted/revertedWithCustomError.ts

contracts/test/emergency/StarkNetValidator.test.ts
```diff
@@ -25,13 +25,36 @@ describe('StarkNetValidator', function () {
...
+      if (!buildInfo) {
+        throw Error('Cannot find build info')
+      }
+      const { solcVersion, input, output } = buildInfo
+      console.log('Sending compilation result for StarkNetValidator test')
+      await network.provider.request({
+        method: 'hardhat_addCompilationResult',
+        params: [solcVersion, input, output],
+      })
+      console.log('Successfully sent compilation result for StarkNetValidator test')
```
krebernisak: Let's extract this to an hardhat specific util method that prepares this for us.

contracts/test/emergency/StarkNetValidator.test.ts
```diff
@@ -52,24 +75,281 @@ describe('StarkNetValidator', function () {
...
+
+  beforeEach(async () => {
     // Deploy the L1 StarkNetValidator
     const starknetValidatorFactory = await ethers.getContractFactory('StarkNetValidator', deployer)
-    starkNetValidator = await starknetValidatorFactory.deploy(
-      mockStarkNetMessaging.address,
-      l2Contract.address,
-    )
+    starkNetValidator = await starknetValidatorFactory
+      .connect(deployer)
```
krebernisak: Deployed should be connected already as you used it in the constructor line before.

contracts/test/emergency/StarkNetValidator.test.ts
aiagentdev-1990: I didn't use it as it was not part of the list of matchers that were available.  I just updated the package so that we can use this.

aiagentdev-1990:
@krebernisak  I'm not sure what you mean here https://github.com/smartcontractkit/chainlink-starknet/pull/107#discussion_r969474516.  Are you saying that we should remove the function?

krebernisak (approved):
LGTM! Love the test coverage 🙌

PR #108, opened 2022-09-06
add some tests

qperrot:
Some more tests

### DIFF
--- a/contracts/test/ocr2/proxy.test.ts
+++ b/contracts/test/ocr2/proxy.test.ts
@@ -1,5 +1,6 @@
 import { assert } from 'chai'
 import { starknet } from 'hardhat'
+import { number } from 'starknet'
 import { Account, StarknetContract, StarknetContractFactory } from 'hardhat/types/runtime'
 import { TIMEOUT } from '../constants'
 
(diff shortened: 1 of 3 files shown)


contracts/test/token/ERC677/ERC677.test.ts
```diff
@@ -122,5 +122,20 @@ describe('ERC677', function () {
         expect(uint256.uint256ToBN(balance1)).to.deep.equal(toBN(0))
       }
     })
+    it('throws an error when sending to 0 address', async () => {
```
ivpavici, line 126: newline :)

contracts/test/ocr2/proxy.test.ts
```diff
@@ -28,8 +29,6 @@ describe('proxy.cairo', function () {
     })
 
     console.log(proxy.address)
```
ivpavici, line 31: is this console.log necessary still?

qperrot: It was already here

ivpavici: ok, then let's leave these comment for Kristijan to see!

archseer (approved):

PR #109, opened 2022-09-09
Upgrade to cairo 0.10 and starknet-devnet 0.3.0

archseer:
Blocked on https://github.com/OpenZeppelin/cairo-contracts/pull/449


archseer:
Methods are no longer implicitly imported, so I think we need to explicitly import `transfer_ownership` etc.

archseer:
`starknet-hardhat-plugin` also needs `0.6.6.alpha`+

packages-ts/starknet-gauntlet/src/commands/base/executeCommand.ts
```diff
@@ -95,7 +95,7 @@ export const makeExecuteCommand = <UI, CI>(config: ExecuteCommandConfig<UI, CI>)
         id: makeCommandId(config.category, config.action, config.suffixes),
         contractAddress: c.contractAddress,
         flags: flags,
-        contract: new Contract(c.contract.abi, c.contractAddress, c.provider.provider),
+        contract: new Contract(c.contract.abi, c.contractAddress || '', c.provider.provider),
```
archseer: Should revert once https://github.com/0xs34n/starknet.js/commit/be50c90ee6af29ab0ceddb679b5d33d5193b9a14 is in a release

.github/workflows/e2e_custom_cl.yml
```diff
@@ -28,8 +28,8 @@ jobs:
         with:
           cl_repo: smartcontractkit/chainlink
           # By default we are integrating with CL develop
-          cl_ref: ${{ github.event.inputs.cl_branch_ref }}
-          #cl_ref: 22c6ac0237a85010b91154277533dfd6acc8912c # https://github.com/smartcontractkit/chainlink/pull/7366
+          # cl_ref: ${{ github.event.inputs.cl_branch_ref }}
+          cl_ref: 6b1552850eb190ff1de094dd619e74e2a4e198ff # https://github.com/smartcontractkit/chainlink/pull/7366
```
aalu1418, line 32: I would suggest running the CI manually pointed to a commit in core rather than updating the reference each time (and then having to change it back)

example in `chainlink-solana`: https://github.com/smartcontractkit/chainlink-solana/pull/387#issuecomment-1224232963

integration-tests/go.sum
```diff
@@ -142,6 +142,8 @@ github.com/docker/distribution v2.8.1+incompatible/go.mod h1:J2gT2udsDAN96Uj4Kfc
 github.com/docopt/docopt-go v0.0.0-20180111231733-ee0de3bc6815/go.mod h1:WwZ+bS3ebgob9U8Nd0kOddGdZWjyMGR8Wziv+TBNwSE=
 github.com/dontpanicdao/caigo v0.3.1-0.20220812122711-b855f2b57bb5 h1:37YQPO1xKfciFA/4y371DOmTaxmn8QD3P1nAPOBpySI=
 github.com/dontpanicdao/caigo v0.3.1-0.20220812122711-b855f2b57bb5/go.mod h1:6i+wtZJN1nvTbh2i55OgMuEHq3wYxe6dvRTHBVWZEZ8=
+github.com/dontpanicdao/caigo v0.3.1-0.20221017220428-d0cf10bd3c1e h1:pLswtYCwQ60gz9UV9EexcJ6KjRvW8lODJMMUPMj/42w=
+github.com/dontpanicdao/caigo v0.3.1-0.20221017220428-d0cf10bd3c1e/go.mod h1:xLt3Z79hOOzkAnMF3P4FuumV6fWHjdyYnrYS9qNkzLY=
```
aalu1418, line 146: `go mod tidy` ?

ops/go.mod
```diff
@@ -24,25 +24,31 @@ require (
 	github.com/cdk8s-team/cdk8s-core-go/cdk8s/v2 v2.3.41 // indirect
 	github.com/chai2010/gettext-go v0.0.0-20160711120539-c6fed771bfd5 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/deckarep/golang-set v1.8.0 // indirect
 	github.com/emicklei/go-restful v2.9.5+incompatible // indirect
+	github.com/ethereum/go-ethereum v1.10.21 // indirect
```
aalu1418, line 29: 😬 we've tried so hard to not import geth...but it potentially seems unavoidable now?

we've discussed moving any relayer integration tests out of the relayer module into the integration-tests folder

aalu1418: oh! but it looks like it isn't needed in `relayer` 🙏🏼

relayer/go.sum
```diff
@@ -21,6 +21,8 @@ github.com/decred/dcrd/dcrec/secp256k1/v4 v4.0.1 h1:YLtO71vCjJRCBcrPMtQ9nqBsqpA1
 github.com/decred/dcrd/dcrec/secp256k1/v4 v4.0.1/go.mod h1:hyedUtir6IdtD/7lIxGeCxkaw7y45JueMRL4DIyJDKs=
 github.com/dontpanicdao/caigo v0.3.1-0.20220812122711-b855f2b57bb5 h1:37YQPO1xKfciFA/4y371DOmTaxmn8QD3P1nAPOBpySI=
 github.com/dontpanicdao/caigo v0.3.1-0.20220812122711-b855f2b57bb5/go.mod h1:6i+wtZJN1nvTbh2i55OgMuEHq3wYxe6dvRTHBVWZEZ8=
+github.com/dontpanicdao/caigo v0.3.1-0.20221017220428-d0cf10bd3c1e h1:pLswtYCwQ60gz9UV9EexcJ6KjRvW8lODJMMUPMj/42w=
+github.com/dontpanicdao/caigo v0.3.1-0.20221017220428-d0cf10bd3c1e/go.mod h1:xLt3Z79hOOzkAnMF3P4FuumV6fWHjdyYnrYS9qNkzLY=
```
aalu1418, line 25: `go mod tidy` ?

relayer/pkg/chainlink/ocr2/contract_reader.go
```diff
@@ -20,14 +21,14 @@ type Reader interface {
...
-	address string
+	address caigotypes.Hash
 	reader  OCR2Reader
 	lggr    logger.Logger
 }
 
 func NewContractReader(address string, reader OCR2Reader, lggr logger.Logger) Reader {
 	return &contractReader{
-		address: address,
+		address: caigotypes.HexToHash(address), // TODO: propagate type everywhere
```
aalu1418, line 31: is the `TODO` resolved?

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -193,9 +200,9 @@ func (txm *starktxm) Ready() error {
 	return txm.starter.Ready()
 }
 
-func (txm *starktxm) Enqueue(tx types.Transaction) error {
+func (txm *starktxm) Enqueue(sender caigotypes.Hash, tx caigotypes.FunctionCall) error {
 	select {
-	case txm.queue <- tx:
+	case txm.queue <- Tx{sender: sender, call: tx}:
```
aalu1418, line 206: just a thought: would it be easier to just pass the `Tx` struct rather than passing two parameters?

i'm thinking in a future case where we might add more parameters, then we can just define them in the struct rather than changing the function signature

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -35,19 +42,19 @@ type starktxm struct {
...
 	stop    chan struct{}
-	queue   chan types.Transaction
+	queue   chan Tx
 	ks      keys.Keystore
 	cfg     Config
 
 	// TODO: use lazy loaded client
-	client    types.Provider
-	getClient func() (types.Provider, error)
+	client    *starknet.Client
```
aalu1418, line 50: can this be an interface instead?

that way we can plug in mocked clients, etc

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -141,23 +148,23 @@ func (txm *starktxm) run() {
...
 	}
 
-	details := caigo.ExecuteDetails{
-		MaxFee: &types.Felt{
-			Int: new(big.Int).SetUint64((fee.OverallFee * FEE_MARGIN) / 100),
-		},
+	fee, _ := big.NewInt(0).SetString(string(feeEstimate.OverallFee), 0)
+	expandedFee := big.NewInt(0).Mul(fee, big.NewInt(int64(FEE_MARGIN)))
+	details := caigotypes.ExecuteDetails{
+		MaxFee: expandedFee,
```
aalu1418: no need to divide by 100 anymore?

relayer/pkg/chainlink/txm/txm.go
aalu1418: ohhhh I see...the `Account` interface now requires the gateway client not an interface anymore

relayer/pkg/starknet/client.go
```diff
@@ -49,18 +49,18 @@ type ReaderWriter interface {
 
 var _ ReaderWriter = (*Client)(nil)
 
-var _ caigotypes.Provider = (*Client)(nil)
+// var _ caigotypes.Provider = (*Client)(nil)
```
aalu1418, line 45: if this doesn't need to conform to the `Provider` type anymore, we can remove the `Unimplemented` interface + related functions

relayer/pkg/starknet/client.go
```diff
@@ -49,18 +49,18 @@ type ReaderWriter interface {
...
 
 type Client struct {
-	gw             *caigogw.GatewayProvider
+	Gw             *caigogw.GatewayProvider
```
aalu1418, line 48: just a thought, maybe rather than exporting it directly, you can have a method to export it? no big deal tho

aalu1418 (changes requested):
looks like we probably also need to update the monitoring side?

.github/workflows/e2e_custom_cl.yml
archseer: But that way the code is merged with the CI failing and will continue to fail on develop until the core PR is merged?

ops/go.mod
archseer: `caigo` now imports `geth` for some utility functions

relayer/pkg/chainlink/ocr2/contract_reader.go
archseer: No, I'd prefer the `NewContractReader` and all parent methods to also take a `Hash` (we could alias it as the `starknet.Address` type)

ops/go.mod
archseer: Also they use `geth`'s RPC package to wrap the JSON RPC interface

relayer/pkg/chainlink/txm/txm.go
archseer: Good catch

PR #95, opened 2022-09-01
[SC-51602] Adding Soak support

smickovskid:
- Adding Soak support to our tests

PR #120, opened 2022-09-15
[feature/sc-51602] Adding soak test support

smickovskid:
- Added a remote runner that runs the soak in k8
- Added logging to soak tests instead of panic
- Default value is 720h for the k8 env and 10m for the soak
- Adjusted soak logic to switch between positive and negative answers


package.json
```diff
@@ -15,6 +15,7 @@
   "main": "packages-ts/starknet-gauntlet-cli/dist/index.js",
   "bin": "packages-ts/starknet-gauntlet-cli/dist/index.js",
   "scripts": {
+    "pregauntlet": "command -v tsc >/dev/null 2>&1 || yarn install",
```
smickovskid: This checks if `tsc` is present and does yarn install if it is not. Needed for the remote runner

archseer:
Should #95 be closed now?

integration-tests/common/test_common.go
```diff
@@ -34,6 +35,7 @@ var (
 	// These are one of the default addresses based on the seed we pass to devnet which is 0
 	defaultWalletPrivKey = ops.PrivateKeys0Seed[0]
 	defaultWalletAddress string // derived in init()
+	soakNetwork          *blockchain.EVMNetwork
```
krebernisak: I'd love to hear more about how this is currently used and how do we replace it with an interface in a network agnostic way. 

1. This configuration is set as env variables to the node environment?
2. `soakNetwork` looks unused? We load it but we don't use it?

cc @kylesmartin @RodrigoAD ping in the context of test configuration and the CLI surfacing or sourcing it. Every test should have some kind of configuration. Here we discuss the network part of that configuration.

integration-tests/go.mod
```diff
@@ -90,7 +90,7 @@ require (
 	github.com/smartcontractkit/libocr v0.0.0-20220726132443-ef1f5a4b63d0 // indirect
 	github.com/spf13/cobra v1.5.0 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
-	github.com/stretchr/testify v1.8.0 // indirect
+	github.com/stretchr/testify v1.8.0
```
krebernisak: go mod tidy? (should go to direct block)

integration-tests/soak/soak_runner_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/smartcontractkit/chainlink-testing-framework/logging"
+	networks "github.com/smartcontractkit/chainlink/integration-tests"
+)
+
+func init() {
+	logging.Init()
+}
+
+var baseEnvironmentConfig = &environment.Config{
+	TTL: time.Hour * 720, // 30 days,
```
krebernisak: We'd want this to be configurable.

Let's add it to the CI and figure out when. We don't want to run this for every commit of every PR, but maybe just when we merge to develop or when we tag a release candidate. 

So we have an open question of when to run it and how long to run it? Please give it some thought and we can sync to discuss our options.

integration-tests/soak/soak_runner_test.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// Run the OCR soak test defined in ./tests/ocr_test.go
+func TestOCRSoak(t *testing.T) {
+	activeEVMNetwork := networks.GeneralEVM() // Environment currently being used to soak test on
```
krebernisak: This introduced a dependency to `github.com/smartcontractkit/chainlink/integration-tests` while that function is just a proxy to a dependency we already have and use. Looks like this:

```go
// GeneralEVM loads general EVM settings from env vars
func GeneralEVM() *blockchain.EVMNetwork {
    return blockchain.LoadNetworkFromEnvironment()
}
```

We should be able to remove this dependency.

integration-tests/common/test_common.go
```diff
@@ -185,6 +193,11 @@ func (t *Test) GetStarkNetAddress() string {
 	return t.Env.URLs[t.Common.ServiceKeyL2][0]
 }
 
+// GetStarkNetAddressRemote Returns the remote StarkNET address
+func (t *Test) GetStarkNetAddressRemote() string {
+	return t.Env.URLs[t.Common.ServiceKeyL2][1]
```
krebernisak: cc @kylesmartin @RodrigoAD ping in the context of how do we write and read configuration for a specific tests. We have an opportunity to replace this with a nicer and more useful chain-agnostic API which should be defined on this QA/DevEx boundary.

integration-tests/soak/soak_runner_test.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// builds tests, launches environment, and triggers the soak test to run
+func soakTestHelper(
+	t *testing.T,
+	testTag string,
+	testEnvironment *environment.Environment,
+	activeEVMNetwork *blockchain.EVMNetwork,
+) {
+	exeFile, exeFileSize, err := actions.BuildGoTests("./", "../soak/tests", "../../")
```
krebernisak: 1. Arn't these already compiled when this file is executed?
2. How do I run these tests without the remote runner?
3. What is the output of this command? When is it copied to the pod?

I'm having trouble understanding the flow here.

integration-tests/soak/soak_runner_test.go
```diff
@@ -0,0 +1,137 @@
...
+		AddHelm(devnet.New(nil)).
+		AddHelm(mockservercfg.New(nil)).
+		AddHelm(mockserver.New(nil)).
+		AddHelm(chainlink.New(0, clConfig)).
+		Run()
+
+	testEnvironment.Client.CopyToPod(
+		testEnvironment.Cfg.Namespace,
+		"../../ops",
+		fmt.Sprintf("%s/%s:/root/", testEnvironment.Cfg.Namespace, "remote-test-runner"),
```
krebernisak: Let's define this `"remote-test-runner"` string somewhere and consider should it be configurable.

integration-tests/soak/tests/ocr_test.go
```diff
@@ -0,0 +1,263 @@
...
+
+		By("Deploying the environment", func() {
+			cfg = &common.Common{
+				ChainName:           chainName,
+				ChainId:             chainId,
+				ServiceKeyChainlink: serviceKeyChainlink,
+				ServiceKeyL1:        serviceKeyL1,
+				ServiceKeyL2:        serviceKeyL2,
+			}
+			//soakNetwork = blockchain.LoadNetworkFromEnvironment()
```
krebernisak: ?

integration-tests/soak/tests/ocr_test.go
```diff
@@ -0,0 +1,263 @@
...
+			t = &common.Test{}
+			t.DeployCluster(5, cfg)
```
krebernisak: Make the magic number 5 configurable

integration-tests/soak/tests/ocr_test.go
```diff
@@ -0,0 +1,263 @@
...
+			})
+			t.SetMockServerValue("", mockServerVal)
+			err = t.Common.CreateJobsForContract(t.GetChainlinkClient(), observationSource, juelsPerFeeCoinSource, ocrAddress)
+			Expect(err).ShouldNot(HaveOccurred(), "Creating jobs should not fail")
+		})
+
+	})
+
+	Describe("with OCRv2 job @soak", func() {
+		It("works", func() {
```
krebernisak: Can we implement this as a plugin (test strategy) and hook it in so it's easy to replace (configure) with a different plugin (i.e., assertion, reporting, other)

krebernisak: We could reuse the same plugin for our smoke test, but for example in our smoke category we'd run this test just until we see 7 rounds (quick smoke test).

package.json
krebernisak: Where is this `"pregauntlet"` script defined? How do we know it needs to be added here?

Not a fan of this type of integration, very implicit and brittle. 
We'd need to explicitly define what a remote runner is, why is it required, how do we set it up.

integration-tests/soak/soak_runner_test.go
```diff
@@ -0,0 +1,137 @@
+package soak_test
```
krebernisak, line 1: Let's please start writing documentation about our integration-tests under `./docs`. Please start with some getting started overview (what is it, how to run it).

package.json
smickovskid: This is a npm check basically, if we prepend `pre` to any existing script it will execute prior to the script which in our case is `gauntlet`, so each time we invoke `gauntlet` the `pregauntlet` will run. This checks if tsc is installed since by default we don't install, we only build so it causes some packages to be missing, if it is not installed it performs `yarn install` which installs all the deps only once

integration-tests/common/test_common.go
kylesmartin: The only fields of `blockchain.EVMNetwork` that get set on the CL nodes are `PrivateKeys` and `URLs`, specifically URLs[0]. Functions like [infra.AddNode](https://github.com/smartcontractkit/blockchain-integrations-framework/blob/c7c59da1b7e6c4d0b184a570252596c5070b324d/integration/infra/infra.go#L9) call [blockchain.EVMNetwork.ChainlinkValuesMap](https://github.com/smartcontractkit/chainlink-testing-framework/blob/cee4064bf95f88615a78eea4351fde413e196425/blockchain/config.go#L86) to create an environment mapping that gets set on the node. The other fields are used to configure the Ethereum client. Here are some relevant links:
- [newEVMClient called in soak test](https://github.com/smartcontractkit/chainlink/blob/2552bf22bba3171b0e6e15f17941c20bad791665/integration-tests/soak/tests/ocr_test.go#L54)
- [newEVMClient function creates an EthereumClient, which implements the EVMClient interface](https://github.com/smartcontractkit/chainlink-testing-framework/blob/cee4064bf95f88615a78eea4351fde413e196425/blockchain/ethereum.go#L49)
- [EVMClient interface](https://github.com/smartcontractkit/chainlink-testing-framework/blob/cee4064bf95f88615a78eea4351fde413e196425/blockchain/blockchain.go#L17)

Ideally, we could further abstract `EVMClient` into a `NetworkClient` interface that developers can use to create a client for any chain or chain family.

To configure the CL node itself in a more network agnostic way, we could encourage developers to use the new [TOML config](https://github.com/smartcontractkit/chainlink/tree/develop/internal/config). Developers could pass their own TOML config file into BIF, which would then convert the TOML into a string and set it on all CL nodes in a test via the [CL_CONFIG](https://github.com/smartcontractkit/chainlink/blob/9a3c836eeb1948d7015909468b42a95c9337ad93/core/cmd/app.go#L71) environment variable.

You can configure test settings in BIF using [soak.TestSettings](https://github.com/smartcontractkit/blockchain-integrations-framework/blob/c7c59da1b7e6c4d0b184a570252596c5070b324d/integration/soak/config.go#L12). This struct can be expanded to manage more test settings.

Currently, we use TOML to configure test and network settings (see [integration README](https://github.com/smartcontractkit/blockchain-integrations-framework/tree/develop/docs/integration#configuration)). I think we could still use this method to configure network clients and test settings, but we should use the core node TOML config to configure nodes in a test.

integration-tests/common/test_common.go
kylesmartin: I think we have two options here. If we assume that developers will hard-code configurations into their tests, we should define a set of test interfaces that they completely implement themselves. If we write the base test, we could use TOML files to allow users to input the necessary configs.

integration-tests/soak/soak_runner_test.go
```diff
@@ -0,0 +1,137 @@
...
+			"p2p_listen_port":             "0",
+		},
+	}
+	err = testEnvironment.
+		AddHelm(remotetestrunner.New(remoteRunnerWrapper)).
+		AddHelm(devnet.New(nil)).
+		AddHelm(mockservercfg.New(nil)).
+		AddHelm(mockserver.New(nil)).
+		AddHelm(chainlink.New(0, clConfig)).
+		Run()
```
aalu1418: can this be moved somewhere more general? especially the env vars, they seem to be duplicates of what is in `test_common`

integration-tests/soak/tests/ocr_test.go
```diff
@@ -0,0 +1,263 @@
+package soak_test
```
aalu1418: there seems to be a lot of duplicated code from `smoke/ocr2_test.go`. is there any way to reduce the code duplication?

aalu1418 (changes requested):
any way that we could reduce duplicate code would be great :)

integration-tests/soak/tests/ocr_test.go
smickovskid: Wrapped gauntlet contract deployments as well as env deployments to reduce duplication

integration-tests/soak/soak_runner_test.go
smickovskid: Refactored a little bit to avoid repetition and exported as variable

integration-tests/soak/soak_runner_test.go
smickovskid: Made configurable with env var `TTL`

integration-tests/soak/soak_runner_test.go
smickovskid: Adjusted the existing documentation in `docs/integration-tests/README.md`

integration-tests/soak/tests/ocr_test.go
smickovskid: This will be handled with the PoC we are working on, my suggestion is to remain with our current setup now so we have running tests while we define the new process

Makefile
```diff
@@ -166,13 +166,25 @@ test-unit-go:
...
 .PHONY: test-integration
 test-integration: test-integration-smoke test-integration-contracts test-integration-gauntlet
 
 .PHONY: test-integration-smoke
-test-integration-smoke: build-ts-contracts
+test-integration-smoke: test-integration-prep
 	ginkgo -v -r --junit-report=tests-smoke-report.xml --keep-going --trace integration-tests/smoke
 
+.PHONY: test-integration-soak
+test-integration-soak: test-integration-prep
```
skudasov, line 184: Can we have the same `make` command in `README.md`?

integration-tests/soak/tests/ocr_test.go
```diff
@@ -0,0 +1,225 @@
...
+			nodeCountSet, nodeCountDefined := os.LookupEnv("NODE_COUNT")
+			if nodeCountDefined == true {
+				nodeCount, err = strconv.Atoi(nodeCountSet)
+				if err != nil {
+					panic(fmt.Sprintf("Please define a proper node count for the test: %v", err))
+				}
+			}
+
+			// Checking if TTL env var is set to adjust duration to custom value
+			ttlValue, ttlDefined := os.LookupEnv("TTL")
```
skudasov: Keep in mind that not all valid `Go` duration strings are valid for `janitor` we are using, please use `ShortDur` from `chainlink-env`

skudasov (dismissed):

skudasov (commented):
The test is pretty complex, it'd be good if we refactor all env vars/configuration and make `StarkNETConfig` that includes all the other configurations and provide `.Defaults()` and `.Validate()` methods to be more readable.
Let's merge what we have for now to have regular CI runs.

.github/workflows/integration-tests-soak.yml
```diff
@@ -12,6 +12,15 @@ on:
         required: true
         default: develop
         type: string
```
aalu1418, line 14: this work flow looks like it's running on every push to `main` and `develop`

is that what we want? isn't that just for smoke tests?

docs/integration-tests/README.md
```diff
@@ -2,23 +2,51 @@
...
+
+##### Env vars
+
+`TTL=72h` - duration of soak (72h default)
+
+`NODE_COUNT` - number of OCR nodes (5 default)
 
 ### Structure
 
 [Commons](../../integration-tests/common/common.go) - Common EVM based methods to generate chains, nodes, key bundles
```
aalu1418:
```suggestion
[Commons](../../integration-tests/common/common.go) - Common Chainlink methods to generate chains, nodes, key bundles
```

integration-tests/common/test_common.go
```diff
@@ -4,42 +4,52 @@ import (
...
-)
-
 var (
-	err       error
-	clImage   string
-	clVersion string
+	err error
+	// Default version to run on
+	clImage   = "public.ecr.aws/chainlink/chainlink"
+	clVersion = "1.9.0"
```
aalu1418: just curious, why are we setting defaults here? it feels like an easy mistake to forget to set it when running locally and then it it will use the defaults

maybe it's just better to throw an error if the image/version is not passed via ENV vars or CLI flags?

integration-tests/go.mod
```diff
@@ -1,76 +1,305 @@
...
 	github.com/satori/go.uuid v1.2.0
-	github.com/smartcontractkit/chainlink-starknet/ops v0.0.0-20221024053349-209ee0d97d21
-	github.com/smartcontractkit/chainlink-starknet/relayer v0.0.0-20221024054022-3c77b546dc04
-	github.com/smartcontractkit/chainlink-testing-framework v1.5.8
-	github.com/smartcontractkit/chainlink/integration-tests v0.0.0-20220822181847-8f75df66a5f1
+	github.com/smartcontractkit/chainlink v1.9.0
+	github.com/smartcontractkit/chainlink-starknet/ops v0.0.0-20221103134312-d40949720ad3
+	github.com/smartcontractkit/chainlink-starknet/relayer v0.0.0-20221103134312-d40949720ad3
+	github.com/smartcontractkit/chainlink-testing-framework v1.7.9-0.20221103145957-d40b562b818b
+	github.com/terra-money/core v0.5.20 // indirect
```
aalu1418: can this be moved with the rest of the `// indirect` references?

integration-tests/go.mod
```diff
@@ -1,76 +1,305 @@
...
+)
+
+require github.com/smartcontractkit/chainlink/integration-tests v0.0.0-20221102233518-1e61a9b970a2
```
aalu1418: can this be moved up into the above `require` (keep them together)

integration-tests/go.mod
```diff
@@ -1,76 +1,305 @@
...
+	gonum.org/v1/gonum v0.11.0 // indirect
+	google.golang.org/genproto v0.0.0-20220712132514-bdd2acd4974d // indirect
+	google.golang.org/grpc v1.49.0 // indirect
+	gopkg.in/ini.v1 v1.66.4 // indirect
+	gopkg.in/natefinch/lumberjack.v2 v2.0.0 // indirect
+	k8s.io/apiextensions-apiserver v0.25.0 // indirect
+	sigs.k8s.io/controller-runtime v0.13.0 // indirect
 )
 
 require (
```
aalu1418: not sure why there are two `// indirect` sections?

integration-tests/soak/soak_runner_test.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/smartcontractkit/chainlink-testing-framework/logging"
+)
+
+func init() {
+	logging.Init()
+}
+
+var (
+	clImage             = "public.ecr.aws/chainlink/chainlink"
+	clVersion           = "1.9.0"
```
aalu1418: mentioned this above, but having hard-coded defaults like this might cause confusion when deploying - compared to ensuring that the CI or user passes it in

aalu1418: another reason is that each time a new release is out, we would have to bump the version in the code - otherwise the default would be outdated...

integration-tests/soak/tests/ocr_test.go
aalu1418: there's still quite a few duplicate parts:
- env var look ups
- job spec creation and node setup
- the entire section for inspecting the operating feed

integration-tests/soak/tests/ocr_test.go
```diff
@@ -0,0 +1,225 @@
+package soak_test
```
aalu1418: can we standardize the name to `ocr2_test.go` like in the smoke test?

aalu1418 (changes requested):

.github/workflows/integration-tests-soak.yml
smickovskid: Idea was to run it on every merge, @krebernisak do we want to just leave it on demand or as it is now ?

integration-tests/common/test_common.go
smickovskid: The custom images we build have a TTL of around 1 week and we cant set them since tests will start failing so I just set this to the latest release candidate. Maybe just forcing an image version is the best option, or develop

shell.nix
```diff
@@ -13,7 +13,7 @@
     nodePackages.typescript-language-server
     nodePackages.npm
 
-    go_1_19
+    go_1_18
```
archseer: This reverts the go upgrade?

.github/workflows/integration-tests-soak.yml
```diff
@@ -59,14 +79,11 @@ jobs:
     steps:
       - name: Checkout the repo
         uses: actions/checkout@v3.0.2
-      - name: Install Nix
-        uses: cachix/install-nix-action@d64e0553100205688c0fb2fa16edb0fc8663c590 # v17
-        with:
-          nix_path: nixpkgs=channel:nixos-unstable
       - name: Run Tests
-        uses: smartcontractkit/chainlink-github-actions/chainlink-testing-framework/run-tests@cd9e407a6e0aa8ef7f315067644c0333846ffd40 # v2.0.24
+        uses: smartcontractkit/chainlink-github-actions/chainlink-testing-framework/run-tests@13437f3be1c82243a8c44eabc043a4fa58d72d73 # v2.0.20
```
aalu1418: wondering: why is this going from v2.0.20 => v2.0.24

integration-tests/soak/tests/ocr_test.go
smickovskid: Decoupled except for "the entire section for inspecting the operating feed" as discussed will be done afterwards

.github/workflows/integration-tests-soak.yml
smickovskid: Leaving as is now

shell.nix
smickovskid: No, I was testing out with go 1.19 but then I just removed nix from the integration tests since it was not needed and then reverted this file as it is on develop

.github/workflows/integration-tests-soak.yml
smickovskid: Bumped accordingly to latest

aalu1418 (approved):
Approving with follow up items to be addressed in a subsequent PR: https://github.com/smartcontractkit/chainlink-starknet/pull/120#discussion_r1016766343

skudasov (approved):

PR #121, opened 2022-09-21
golangci-lint

archseer:
Copied the latest config from https://github.com/smartcontractkit/chainlink


.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,38 @@
...
+        with:
+          go-version: ${{ steps.tool-versions.outputs.golang_version }}
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
+          version: v${{ steps.tool-versions.outputs.golangci-lint_version }}
+
+          # Optional: working directory, useful for monorepos
+          working-directory: relayer
```
krebernisak, line 26: We should include all Go modules in the monorepo: https://github.com/golangci/golangci-lint/issues/828#issuecomment-658207652

krebernisak (dismissed):

.github/workflows/golangci-lint.yml
krebernisak: Extracted GH Issue: https://github.com/smartcontractkit/chainlink-starknet/issues/122 (not blocking)

relayer/pkg/chainlink/ocr2/transmissions_cache.go
```diff
@@ -20,7 +20,6 @@ var _ median.MedianContract = (*transmissionsCache)(nil)
 type transmissionsCache struct {
 	transmissionDetails TransmissionDetails
 	tdLock              sync.RWMutex
-	tdTime              time.Time
```
aalu1418, line 23: Interesting that this wasn't used...there should be a TTL check here similar to the contract_cache implementation

relayer/pkg/starknet/utils.go
```diff
@@ -74,12 +74,12 @@ func DecodeFelts(felts []*big.Int) ([]byte, error) {
...
-		buf := buf[:Min(chunkSize, length)]
+		bytes := buf[:Min(chunkSize, length)]
 
-		felt.FillBytes(buf)
-		data = append(data, buf...)
+		felt.FillBytes(bytes)
+		data = append(data, bytes...)
 
-		length -= len(buf)
+		length -= len(bytes)
```
aalu1418: wouldn't naming the slice `bytes` here conflict with the `bytes` module that is also used?
(i guess even if it doesn't, naming it `bytes` is a bit confusing to the reader since the module is also used)

aalu1418 (changes requested):
ideally we use a different name for the slice naming
(the other issue can be addressed later)

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,38 @@
...
+
+          # Optional: golangci-lint command line arguments.
+          args: --enable=gofmt --tests=false --exclude-use-default --timeout=5m0s
```
jmank88: Could drop this CI timeout since the `.golangci.yml` specifies one now.
```suggestion
          args: --enable=gofmt --tests=false --exclude-use-default
```

.github/workflows/golangci-lint.yml
jmank88, line 26: Is it possible to use matrix args or something even more automated to repeat for each module?

jmank88: Is it possible to use matrix args or something even more automated to repeat for each module?

relayer/pkg/starknet/utils.go
archseer: It's just scoped to that small block, can use a different name though, what should we name it?

aalu1418: maybe something simple like `byteBuffer` or `buffer` ?

aalu1418 (approved):
fixed the one renaming change I requested
`bytes` => `bytesBuffer`

PR #128, opened 2022-09-27
add events in StarknetValidator.sol and ownable.cairo and add test 

qperrot:
Add events and some tests

### DIFF
--- a/contracts/src/chainlink/cairo/access/ownable.cairo
+++ b/contracts/src/chainlink/cairo/access/ownable.cairo
@@ -13,6 +13,10 @@ from starkware.cairo.common.math import assert_not_zero
 func OwnershipTransferred(previousOwner : felt, newOwner : felt):
 end
 
+@event
+func OwnershipTransferRequested(from_address : felt, to : felt):
+end
+
 #
 # Storage
 #
@@ -35,7 +39,7 @@ namespace Ownable:
         with_attr error_message("Ownable: Cannot transfer to zero address"):
             assert_not_zero(owner)
         end
-        _transfer_ownership(owner)
+        _accept_ownership_transfer(owner)
         return ()
     end
 
@@ -73,7 +77,7 @@ namespace Ownable:
         assert_only_owner()
         Ownable_proposed_owner.write(new_owner)
         let (previous_owner : felt) = Ownable_owner.read()
-        OwnershipTransferred.emit(previous_owner, new_owner)
+        OwnershipTransferRequested.emit(previous_owner, new_owner)
         return ()
     end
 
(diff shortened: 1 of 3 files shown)


contracts/src/chainlink/cairo/access/ownable.cairo
```diff
@@ -72,6 +72,8 @@ namespace Ownable:
         end
         assert_only_owner()
         Ownable_proposed_owner.write(new_owner)
+        let (previous_owner : felt) = Ownable_owner.read()
+        OwnershipTransferred.emit(previous_owner, new_owner)
```
krebernisak: Nice catch! 

Reading this contract, looks like we have some issues with `transfer_ownership` fn which also emits `OwnershipTransferred` event while it shouldn't! Check out these as a guideline:

- https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.8/ConfirmedOwnerWithProposal.sol
- https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/Ownable2Step.sol

Let's actually add the new type of event `event OwnershipTransferRequested(address indexed from, address indexed to);` and only emit that one in external `transfer_ownership` fn.

krebernisak: Also I feel we should rename the internal `_transfer_ownership` as `_accept_ownership` or `_accept_ownership_transfer`.

The current implementation of `_transfer_ownership` doesn't match what an external `transfer_ownership` does (just proposes, 2-step) so it's confusing.

contracts/src/chainlink/solidity/emergency/StarkNetValidator.sol
```diff
@@ -43,6 +43,10 @@ contract StarkNetValidator is TypeAndVersionInterface, AggregatorValidatorInterf
   event GasConfigSet(uint256 gasEstimate, address indexed gasPriceL1Feed);
   /// @notice emitted when a new gas access-control contract is set
   event ConfigACSet(address indexed previous, address indexed current);
+  /// @notice emitted when a new source aggregator contract is set
+  event SourceAggregatorSet(address indexed previous, address indexed current);
+  /// @notice emitted when withdrawFunds or withdrawFundsTo is call
+  event WithdrawFunds(address indexed recipient, uint256 amount);
```
krebernisak:
```suggestion
  event FundsWithdrawn(address indexed recipient, uint256 amount);
```

contracts/test/emergency/StarkNetValidator.test.ts
```diff
@@ -219,6 +219,38 @@ describe('StarkNetValidator', () => {
     })
   })
 
+  describe('#setSourceAggregator', () => {
+    describe('when called by non owner', () => {
+      it('reverts', async () => {
+        await expect(
+          starkNetValidator.connect(eoaValidator).setSourceAggregator(ethers.constants.AddressZero),
```
krebernisak: nit: It's weird to connect a contract like this. 

I'd prefer if we'd explicitly deploy a new test account (once, in beforeAll), call it `alice` and connect `alice` in tests like this.

ivpavici: not done yet (with last push)

krebernisak (dismissed):

PR #129, opened 2022-09-27
fixes: cache TTL + key generation

aalu1418:


relayer/pkg/chainlink/keys/utils.go
```diff
@@ -40,10 +40,10 @@ func PubKeyToStarkKey(pubkey PublicKey) []byte {
 	return starknet.PadBytes(pubkey.X.Bytes(), byteLen)
 }
 
-// reimplements: https://github.com/dontpanicdao/caigo/blob/main/utils.go#L85
+// reimplements parts of https://github.com/dontpanicdao/caigo/blob/main/utils.go#L85
 // using io.Reader, and Key struct
 func GenerateKey(material io.Reader) (k Key, err error) {
```
krebernisak, line 46: Let's properly document this function and add a comment explaining that "we generate the PK as a pseudo-random number in the interval [1, CurveOrder - 1]`.

krebernisak (dismissed):

krebernisak (approved):

PR #134, opened 2022-09-29
Fixe validation accept negative

qperrot:


contracts/src/chainlink/cairo/ocr2/aggregator.cairo
```diff
@@ -165,8 +165,8 @@ func constructor{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_p
     let range : Range = (min=min_answer, max=max_answer)
     Aggregator_answer_range.write(range)
 
-    with_attr error_message("decimals exceed 2^8"):
-        assert_lt(decimals, UINT8_MAX)
+    with_attr error_message("decimals is negative or exceed 2^8"):
```
ivpavici:
```suggestion
    with_attr error_message("decimals are negative or exceed 2^8"):
```

contracts/src/chainlink/cairo/emergency/SequencerUptimeFeed/library.cairo
```diff
@@ -62,7 +62,7 @@ func require_valid_round_id{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, ra
     with_attr error_message("invalid round_id"):
         assert_not_zero(round_id)
         # TODO: do we need to check if uint80 is overflown?
```
krebernisak: @cds95 any ideas what's this about?

We should consider removing this TODO from here.

contracts/src/chainlink/cairo/ocr2/aggregator.cairo
```diff
@@ -165,8 +165,8 @@ func constructor{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_p
     let range : Range = (min=min_answer, max=max_answer)
     Aggregator_answer_range.write(range)
 
-    with_attr error_message("decimals exceed 2^8"):
-        assert_lt(decimals, UINT8_MAX)
+    with_attr error_message("decimals are negative or exceed 2^8"):
+        assert_nn_lt(decimals, UINT8_MAX)
```
krebernisak: This constant is defined as 255: https://github.com/OpenZeppelin/cairo-contracts/blob/b764660da224ee757b831ea62bf7a4c63ee898fb/src/openzeppelin/utils/constants/library.cairo#L10

So we can just use the standard lib `assert_nn_le` instead of our aux `assert_nn_lt`.

These two are the same:
- `assert_nn_le(decimals, UINT8_MAX)`
- `assert_nn_lt(decimals, 2ˆ8)`

cc @archseer

qperrot: @krebernisak `assert_nn_lt `also check if decimals is not equal to 2^8, `which assert_nn_le`does not. Is it ok if decimals == 2^8 ?

krebernisak: My suggestion is to just use `assert_nn_le(decimals, UINT8_MAX)` as that will assert that `0 >= decimals <= 255`, which is what we should be checking here.

This means `assert_nn_lt` is not required anymore.

contracts/src/chainlink/cairo/utils.cairo
```diff
@@ -1,4 +1,9 @@
-from starkware.cairo.common.math import assert_in_range, assert_lt_felt
+from starkware.cairo.common.math import (
```
krebernisak: We have some unused imports now

contracts/src/chainlink/cairo/emergency/SequencerUptimeFeed/library.cairo
krebernisak: @qperrot let's remove this comment as it doesn't make sense in the current context

krebernisak (approved):

PR #140, opened 2022-09-30
Fixed issues 6 & 7

taco-paco:


relayer/pkg/starknet/utils.go
```diff
@@ -71,15 +71,22 @@ func DecodeFelts(felts []*big.Int) ([]byte, error) {
...
-	length := int(felts[0].Int64())
+	length := felts[0].Uint64()
 
 	for _, felt := range felts[1:] {
-		bytesBuffer := buf[:Min(chunkSize, length)]
+		bytesLen := Min(chunkSize, length)
+		bytesBuffer := buf[:bytesLen]
+
+		// TODO: clarify if only last felt can be != chunkSize
+		// If no then Decoding and then Encoding same array will yield different result
```
archseer: It's only last felt: 65 bytes will be split into `31 + 31 + 3` bytes so that last felt will have a bunch of padding.

relayer/pkg/chainlink/keys/ocr2key.go
```diff
@@ -89,6 +89,10 @@ func (sk *OCR2Key) Verify(publicKey ocrtypes.OnchainPublicKey, reportCtx ocrtype
 	var keys [2]PublicKey
 	keys[0].X = new(big.Int).SetBytes(publicKey)
 	keys[0].Y = caigo.Curve.GetYCoordinate(keys[0].X)
+	if keys[0].Y == nil {
+		return false
```
krebernisak, line 95:
```suggestion
                // When there is no point with the provided x-coordinate, the GetYCoordinate function returns the nil value.
		return false
```

relayer/pkg/starknet/utils.go
krebernisak: @PacoEdwin let's add a test that decodes then encodes again and checks the final result.

relayer/pkg/starknet/utils.go
```diff
@@ -71,15 +71,22 @@ func DecodeFelts(felts []*big.Int) ([]byte, error) {
 
 	data := []byte{}
 	buf := make([]byte, chunkSize)
-	length := int(felts[0].Int64())
+	length := felts[0].Uint64()
 
 	for _, felt := range felts[1:] {
```
krebernisak, line 76: @archseer looks like we have one extra felt that we can remove from the ocr config if we want to optimise.

When we encode a byte array as felts, first felt represents the length. But when we set the contract config we prepend the length one more time so it basically [looks like this](https://github.com/smartcontractkit/chainlink-starknet/blob/3df46ca75401919aee6f9b65675ec48792a614c1/relayer/pkg/chainlink/ocr2/config_digester.go#L85):  
```
...[version][length][length][off_chain_config_raw]...
```

archseer: It's two different lengths. The first is a number of felts, used by cairo/starknet to specify an array. The second is a byte length of the data after decoding, that tells us if there's any padding on the last felt. So a 65 byte array is encoded as 

```
[version] 3 65 x y z
```

Where felts = 3, byte length = 65 and `z` will contain three bytes and the rest is padding.

krebernisak: Got it, thx!

relayer/pkg/starknet/utils.go
taco-paco: Added tests Decode -> Encode, Encode -> Decode

relayer/pkg/chainlink/keys/ocr2key.go
taco-paco: Done

PR #144, opened 2022-10-03
Implement account funder

qperrot:


contracts/test/account/fundAccount.test.ts
```diff
@@ -0,0 +1,75 @@
+import { assert } from 'chai'
+import { starknet } from 'hardhat'
+import { Account } from 'hardhat/types/runtime'
+import { TIMEOUT } from '../constants'
+import { AccountFunder } from '@chainlink/starknet/src/utils'
```
krebernisak: Move this test to the source `@chainlink/starknet`

packages-ts/starknet/src/utils/index.ts
```diff
@@ -1,4 +1,4 @@
-import { constants, ec, encode, hash, number, uint256, stark, KeyPair } from 'starknet'
+import { constants, encode, number } from 'starknet'
```
krebernisak: 1. Move this file to `packages-ts/starknet/src/utils.ts`
2. Define `packages-ts/starknet/src/index.ts` as:

```ts
export * from './utils';
```

3. Import it as:

```ts
import { AccountFunder } from '@chainlink/starknet'
```

packages-ts/starknet/src/utils/index.ts
```diff
@@ -79,3 +79,42 @@ export const toFelt = (int: number | BigNumberish): BigNumberish => {
...
+
+      try {
+        await fetch(`${gateway_url}/mint`, {
+          method: 'post',
+          body: JSON.stringify(body),
+          headers: { 'Content-Type': 'application/json' },
+        })
+      } catch (error: any) {
+        throw Error(error?.message)
+      }
```
krebernisak: Remove this catch clause, don't see a purpose for it

packages-ts/starknet/src/utils/index.ts
```diff
@@ -79,3 +79,42 @@ export const toFelt = (int: number | BigNumberish): BigNumberish => {
...
+  private opts: any
+
+  constructor(opts: any) {
+    this.opts = opts
+  }
+
+  public async fund(accounts: FundAccounts[]) {
+    let gateway_url: string
+    if (this.opts.network == 'devnet') {
+      gateway_url = process.env.NODE_URL || 'http://127.0.0.1:5050'
```
krebernisak: The url should also be an input option

contracts/test/account/fundAccount.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    bob = await starknet.deployAccount('OpenZeppelin')
+    await funder.fund([
+      { account: alice.address, amount: 5000 },
+      { account: bob.address, amount: 8000 },
+    ])
+  })
+
+  it('should have fund alice', async () => {
+    let gateway_url = process.env.NODE_URL || 'http://127.0.0.1:5050'
+    const response = await fetch(`${gateway_url}/account_balance?address=${alice.address}`, {
```
krebernisak: Let's use starknet.js instead of doing manual calls like this

packages-ts/starknet/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "format": "yarn prettier --write ./src",
+    "format:check": "yarn prettier --check ./src",
+    "clean": "rm -rf ./dist/ ./bin/",
+    "build": "yarn clean && tsc -b",
+    "bundle": "yarn build && pkg ."
+  },
+  "dependencies": {
+    "@chainlink/gauntlet-contracts-example": "*",
+    "@chainlink/gauntlet-core": "0.3.1",
+    "@chainlink/starknet-gauntlet": "^0.0.1"
```
krebernisak: We shouldn't be depending on those. Why are they declared here?

packages-ts/starknet/package.json
```diff
@@ -0,0 +1,27 @@
...
+  ],
+  "main": "./dist/index.js",
+  "types": "dist/index.d.ts",
+  "files": [
+    "dist/**/*",
+    "!dist/**/*.test.js"
+  ],
+  "scripts": {
+    "format": "yarn prettier --write ./src",
+    "format:check": "yarn prettier --check ./src",
```
krebernisak, line 20: Check out other packages and their definitions. The test script definition is missing.

packages-ts/starknet/src/utils/index.ts
krebernisak: Let's split this into an `AccountFunder` that accepts `opts: AccountFunderOptions` and a function that loads these options from the environment. The function should return options for Devnet as default when nothing is configured in the environment.

Then we need to implement and choose between different funding strategies per network: `DevnetFundingStrategy` and `AllowanceFundingStrategy` (operator acc funds test accounts by sending them his private funds from his account - applicable on live networks)

I suggest we export those as an account object:

```typescript
import { account } from '@chainlink/starknet'

const opts = account.makeFunderOptsFromEnv()
const funder = new account.Funder(opts)
await funder.fund([{ account: owner.address, amount: 5000 }])
```

This suggestion supersedes the previous suggestion here: https://github.com/smartcontractkit/chainlink-starknet/pull/144/files#r986174863

Try to design this as a nice API for consumers to use in tests.

packages-ts/starknet/test/fundAccount.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  let alice: Account
+  let bob: Account
+  let provider: SequencerProvider
+  let erc20_address: string
+
+  const opts = account.makeFunderOptsFromEnv()
+  const funder = account.Funder(opts)
+
+  before(async function () {
+    const gateway_url = process.env.NODE_URL || 'http://127.0.0.1:5050'
```
ivpavici: nit: maybe extract 'http://127.0.0.1:5050' to a const, as well as below where used

packages-ts/starknet/test/fundAccount.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    ])
+  })
+
+  it('should have fund alice', async () => {
+    const balance = await alice.callContract({
+      contractAddress: erc20_address,
+      entrypoint: 'balanceOf',
+      calldata: [BigInt(alice.address).toString(10)],
+    })
+    // const balance = await erc20.balanceOf(alice.address)
```
ivpavici: leftover comment?

packages-ts/starknet/src/utils.ts
```diff
@@ -0,0 +1,189 @@
...
+
+    accounts.forEach(async (account) => {
+      const hash = await accountFunder.execute(
+        {
+          contractAddress: ERC20_ADDRESS_TESTNET,
+          entrypoint: 'transfer',
+          calldata: [account.account, account.amount],
+        },
+        undefined,
+        { maxFee: '32703804275172' },
```
ivpavici: I think we could call estimateFee before and the send it here instead of hardcoded value

packages-ts/starknet/src/utils.ts
```diff
@@ -0,0 +1,189 @@
...
+  constructor(opts: AccountFunderOptions) {
+    this.opts = opts
+    if (this.opts.network === 'devnet') {
+      this.strategy = new DevnetFundingStrategy()
+      return
+    }
+    this.strategy = new AllowanceFundingStrategy()
+  }
+
+  //This function add some funds to predeploy account that we are using in our test.
```
ivpavici:
```suggestion
  //This function adds some funds to pre-deployed account that we are using in our test.
```

packages-ts/starknet/src/utils.ts
```diff
@@ -0,0 +1,189 @@
...
+
+interface FunderOpts {
+  makeFunderOptsFromEnv: () => AccountFunderOptions
+  Funder: (opts: AccountFunderOptions) => AccountFunder
+}
+
+// This function loads options from the environment.
+// It returns options for Devnet as default when nothing is configured in the environment.
+const makeFunderOptsFromEnv = (): AccountFunderOptions => {
+  const network = process.env.NETWORK || 'devnet'
```
ivpavici: I would maybe extract `devnet` string here and in one place below (line 128) to a constant

contracts/tsconfig.json
```diff
@@ -1,12 +1,18 @@
...
+    "module": "commonjs",
+    "strict": true,
+    "esModuleInterop": true,
+    "outDir": "dist",
+    "resolveJsonModule": true
+  },
+  "include": [
+    "./scripts",
+    "./test",
+    "../packages-ts/starknet/test/fundAccount.test.ts"
```
krebernisak: Hmm, why the change here?

packages-ts/starknet/test/fundAccount.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    const gateway_url = process.env.NODE_URL || DEVNET_URL
+    provider = new SequencerProvider({ baseUrl: gateway_url })
+
+    if (gateway_url === DEVNET_URL) {
+      erc20_address = ERC20_ADDRESS_DEVNET
+    } else {
+      erc20_address = ERC20_ADDRESS_TESTNET
+    }
+
+    const alice_starkKeyPair = ec.genKeyPair()
```
krebernisak: Please use camelCase instead of snake_case

```suggestion
    const aliceStarkKeyPair = ec.genKeyPair()
```

packages-ts/starknet/src/utils.ts
```diff
@@ -0,0 +1,207 @@
...
+  return number.toBN(int).umod(prime)
+}
+
+// NOTICE: Leading zeros are trimmed for an encoded felt (number).
+//   To decode, the raw felt needs to be start padded up to max felt size (252 bits or < 32 bytes).
+export const hexPadStart = (data: number | bigint, len: number) => {
+  return `0x${data.toString(16).padStart(len, '0')}`
+}
+
+interface FunderOpts {
```
krebernisak: We have a full package now to work with, so let's extract the funding functionality into a separate file like `funder.ts`

packages-ts/starknet/src/utils.ts
```diff
@@ -0,0 +1,207 @@
...
+// This function loads options from the environment.
+// It returns options for Devnet as default when nothing is configured in the environment.
+const makeFunderOptsFromEnv = (): AccountFunderOptions => {
+  const network = process.env.NETWORK || DEVNET_NAME
+  const gateway_url = process.env.NODE_URL || DEVNET_URL
+
+  return { network, gateway_url }
+}
+
+const Funder = (opts: AccountFunderOptions): AccountFunder => {
```
krebernisak: Feels redundant, what is this function for? 
It just calls the constructor. Why call this fn instead of the constructor directly?

krebernisak: interface FunderOpts also feels redundant in this context.

We should just export the things we want public, we don't need an interface for the things we export.

packages-ts/starknet/src/utils.ts
```diff
@@ -0,0 +1,207 @@
...
+        lite: true,
+      }
+
+      try {
+        await fetch(`${opts.gateway_url}/mint`, {
+          method: 'post',
+          body: JSON.stringify(body),
+          headers: { 'Content-Type': 'application/json' },
+        })
+      } catch (error: any) {}
```
krebernisak: How do we know if this fails? Swallowing errors like this is just a foot gun that will go off in the worst moment. If the funding strategy fails, the test should error.

I'd suggest we just remove the `try..catch` from here.

packages-ts/starknet/src/utils.ts
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// Fund the Account on Testnet
+class AllowanceFundingStrategy implements IFundingStrategy {
+  public async fund(accounts: FundAccounts[], opts: AccountFunderOptions) {
+    const provider = new SequencerProvider({
+      baseUrl: 'https://alpha4.starknet.io',
+    })
+
+    const keyPair = ec.getKeyPair(process.env.ACCOUNT_PRIVATE_KEY)
```
krebernisak: We have an input `opts` an function that is called `makeFunderOptsFromEnv` and we still read from the env directly in this implementation :)

Please move this to `makeFunderOptsFromEnv` and read from prepared `opts` here.

packages-ts/starknet/src/utils.ts
```diff
@@ -0,0 +1,207 @@
...
+      const hash = await accountFunder.execute(
+        {
+          contractAddress: ERC20_ADDRESS_TESTNET,
+          entrypoint: 'transfer',
+          calldata: [
+            account.account,
+            bnToUint256(account.amount).low.toString(),
+            bnToUint256(account.amount).high.toString(),
+          ],
+        },
```
krebernisak: Extract this payload as a variable, instead of c/p here as is also defined couple of lines above.

krebernisak (changes requested):

packages-ts/starknet/src/utils.ts
qperrot: Yes If this fail the account will doesn't have any fund and the test will fail that's all. I remove it.

packages-ts/starknet/src/utils.ts
qperrot: Yes I was doing it in a wrong way I fixed it

contracts/tsconfig.json
qperrot: Removed

packages-ts/starknet/src/utils.ts
qperrot: Done

packages-ts/starknet/src/utils.ts
qperrot: Done

packages-ts/starknet/src/utils.ts
qperrot: Done

krebernisak (approved):

packages-ts/integration-eqlabs-multisig/test/Multisig.test.ts
```diff
@@ -3,10 +3,14 @@ import { starknet } from 'hardhat'
 import { StarknetContract, Account } from 'hardhat/types/runtime'
 import { getSelectorFromName } from 'starknet/dist/utils/hash'
 import { number } from 'starknet'
+import { account } from '@chainlink/starknet'
 
 describe('Multisig integration tests', function () {
   this.timeout(300_000)
 
+  const opts = account.makeFunderOptsFromEnv()
+  const funder = account.Funder(opts)
```
krebernisak: Looks like the test is failing b/c of this:

```suggestion
  const funder = new account.Funder(opts)
```

packages-ts/integration-starkgate/test/tokenBridge.test.ts
```diff
@@ -5,19 +5,21 @@ import { uint256, number } from 'starknet'
...
+import { account, hexPadStart } from '@chainlink/starknet'
 
 const NAME = 'ChainLink Token'
 const SYMBOL = 'LINK'
 
 describe('Test StarkGate token bridge + link_token.cairo', function () {
   this.timeout(TIMEOUT)
 
+  const opts = account.makeFunderOptsFromEnv()
+  const funder = account.Funder(opts)
```
krebernisak:
```suggestion
  const funder = new account.Funder(opts)
```

packages-ts/starknet/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './utils'
+export * as account from './funder'
```
krebernisak: We shouldn't alias like this. I suggest just rename the file to `account.ts`

```suggestion
export * as account from './account'
```

packages-ts/starknet/src/funder.ts
```diff
@@ -0,0 +1,110 @@
...
+
+interface AccountFunderOptions {
+  network?: string
+  gateway_url?: string
+  keyPair?: KeyPair
+}
+
+// Define the Strategy to use depending on the network.
+export class Funder {
+  private opts: AccountFunderOptions
```
krebernisak:
```suggestion
export class Funder {
  private opts: FunderOptions
```

packages-ts/starknet/src/funder.ts
```diff
@@ -0,0 +1,110 @@
...
+}
+
+interface FundAccounts {
+  account: string
+  amount: number
+}
+
+interface AccountFunderOptions {
+  network?: string
+  gateway_url?: string
```
krebernisak:
```suggestion
  gateway?: string
```

packages-ts/starknet/src/funder.ts
```diff
@@ -0,0 +1,110 @@
...
+
+// Fund the Account on Testnet
+class AllowanceFundingStrategy implements IFundingStrategy {
+  public async fund(accounts: FundAccounts[], opts: AccountFunderOptions) {
+    const provider = new SequencerProvider({
+      baseUrl: 'https://alpha4.starknet.io',
+    })
+
+    const keyPair = ec.getKeyPair(process.env.ACCOUNT_PRIVATE_KEY)
+    const accountFunder = new Account(provider, process.env.ACCOUNT.toLowerCase(), keyPair)
```
krebernisak: Please use `process.env.ACCOUNT_ADDRESS` for this option:

```suggestion
    const operator = new Account(provider, opts.accountAddr, opts.keyPair)
```

krebernisak (changes requested):

krebernisak (approved):

PR #146, opened 2022-10-04
Fixe underflow uint256 sub

qperrot:


archseer:
`link_available_for_payment` should also be fixed?

qperrot:
>

contracts/src/chainlink/cairo/ocr2/aggregator.cairo
```diff
@@ -1001,6 +1002,11 @@ func withdraw_funds{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_chec
     )
 
     let (link_due_uint256 : Uint256) = felt_to_uint256(link_due)
+    let (res) = uint256_le(link_due_uint256, balance)
+    with_attr error_message("Total amount due exceeds the balance"):
```
krebernisak, line 1013: We should prefix all error messages with contract/namespace like so:

```suggestion
    with_attr error_message("Aggregator: Total amount due exceeds the balance"):
```

krebernisak (changes requested):
For the `link_available_for_payment ` fn, if the result is expected to be negative number in situations of contract debt, then there is no issue but this should be properly documented:
> However, since the felt type in Cairo is not signed, whoever calls this function will have to interpret the result as the correct negative value.

@qperrot please add in-line docs to the function, and also the line where the subtraction happens so we explain what's expected here.

contracts/src/chainlink/cairo/ocr2/aggregator.cairo
krebernisak: No need to do in this PR. I've created a separate issue for it: https://github.com/smartcontractkit/chainlink-starknet/issues/156

krebernisak (approved):

PR #152, opened 2022-10-05
Add some check to accept the smallest 128-bit

qperrot:


contracts/test/ocr2/aggregator.test.ts
```diff
@@ -62,8 +64,8 @@ describe('aggregator.cairo', function () {
   let token: StarknetContract
   let aggregator: StarknetContract
 
-  let minAnswer = -10
-  let maxAnswer = 1000000000
```
archseer: Hmm I like the original `min/max` values since that tests that those two config options work correctly. Maybe keep the original bounds and then check in your tests that the error returned isn't `"value not in int128 range"`?

archseer (dismissed):

PR #153, opened 2022-10-05
Added state reload after pod restart

smickovskid:
- Added `postStart` script to load dump from devnet if it exists, this solves when devnet crashes / restarts and state is lost
- DIdn't put as additional entrypoint since it needs to run after the liveliness probe is kicked off

### DIFF
--- a/ops/charts/devnet/templates/deployment.yaml
+++ b/ops/charts/devnet/templates/deployment.yaml
@@ -17,13 +17,21 @@ spec:
         release: {{ .Release.Name }}
         app: starknet-dev
     spec:
+      volumes:
+        - name: vol
+          hostPath:
+            path: /dumps
       containers:
       - name: {{ .Chart.Name }}
+        volumeMounts:
+          - name: vol
+            mountPath: /dumps
 {{- if eq .Values.real_node true }}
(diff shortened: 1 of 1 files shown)


ops/charts/devnet/templates/deployment.yaml
```diff
@@ -56,6 +64,10 @@ spec:
             port: {{ .Values.service.externalPort }}
           initialDelaySeconds: 2
           periodSeconds: 1
+        lifecycle:
+          postStart:
+            exec:
+              command: ["/bin/sh", "-c", "ENDPOINT=http://localhost:{{ .Values.service.internalPort }}; for i in `seq 1 5`; do wget --spider $ENDPOINT/is_alive >> /dumps/log.txt 2>&1; if [ $? -eq 0 ]; then test -f /dumps/dump.pkl; if [ $? -eq 0 ]; then wget -O- --post-data='{\"path\": \"/dumps/dump.pkl\"}' --header='Content-Type:application/json' $ENDPOINT/load >> /dumps/log.txt 2>&1; if [ $? -eq 0 ]; then break; fi; fi; fi; sleep 5; done"]
```
krebernisak, line 70: Could we extract this to a script `devnet_reload.sh` (or smth) that accepts arguments?

PR #154, opened 2022-10-05
fix: Avoid panic in DecodeFelts

### LINKED ISSUE #151: Audit issue 6
After merge of [this PR](https://github.com/smartcontractkit/chainlink-starknet/pull/140) the following if statement was removed in this [commit](https://github.com/smartcontractkit/chainlink-starknet/commit/3417b2736902f23c1063cafcb49190f37b6042e0). 
```
if bytesLen != uint64(len(felt.Bytes())) {
 	return nil, errors.New("invalid: felt array can't be decoded")
 }
```

I added it for the couple of reasons:
1. There're no guarantees that every element in `felts []*big.Int` is 31 bytes long(except the last one). Assume the length >= 31 bytes. Then the [FillBytes on this line will panic](https://github.com/smartcontractkit/chainlink-starknet/blob/3417b2736902f23c1063cafcb49190f37b6042e0/relayer/pkg/starknet/utils.go#L80). From FillBytes source: `If the absolute value of x doesn't fit in buf, FillBytes will panic.` That happens because `bytesBuffer` is at most 31.
2. Since there is no check that felt len is 31 bytes the following test will as well pass
```
func TestIncorrectDecode(t *testing.T) {
	felts := make([]*big.Int, 3)
	felts[0] = big.NewInt(35)
	// bytes length can be less than 31
	felts[1] = new(big.Int).SetBytes([]byte{0x1, 0x2, 0x3, 0x4})
	felts[2] = new(big.Int).SetBytes([]byte{0x11, 0x21})

	res, err := DecodeFelts(felts)
	fmt.Println(res)

	assert.Equal(t, err, nil)
}
```
Now any element can be less than 31 bytes long. In addition `felts[2]` contains values from `felt[1]` because on second iteration `bytesLen` is 4 but `felts[2]` len is 2 so now `bytesBuffer` has 2 bytes from `felt[1]` and 2 bytes from `felt[2]`
That means that Decode -> Encode will not yield the same result anymore since Encode allows only last element to be less than 31 bytes long

archseer:
Fixes #151

### DIFF
--- a/relayer/pkg/starknet/utils.go
+++ b/relayer/pkg/starknet/utils.go
@@ -77,6 +77,12 @@ func DecodeFelts(felts []*big.Int) ([]byte, error) {
 		bytesLen := Min(chunkSize, length)
 		bytesBuffer := buf[:bytesLen]
 
+		// TODO: this is inefficient because Bytes() and FillBytes() duplicate work
+		// reuse felt.Bytes()
+		if bytesLen < uint64(len(felt.Bytes())) {
+			return nil, errors.New("invalid: felt array can't be decoded")
+		}
+
 		felt.FillBytes(bytesBuffer)
 		data = append(data, bytesBuffer...)
 
--- a/relayer/pkg/starknet/utils_test.go
+++ b/relayer/pkg/starknet/utils_test.go
@@ -130,3 +130,34 @@ func TestDecodeThenEncode(t *testing.T) {
 	encodedFelts := EncodeFelts(decodedFelts)
 	assert.Equal(t, encodedFelts, felts)
 }
+
+func TestIncorrectDecode(t *testing.T) {
+	felts := make([]*big.Int, 3)
+	felts[0] = big.NewInt(35)
+	// bytes length can be less than 31
+	felts[1] = new(big.Int).SetBytes([]byte{0x1, 0x2, 0x3, 0x4})
+	felts[2] = new(big.Int).SetBytes([]byte{0x11, 0x21})
+
+	_, err := DecodeFelts(felts)
+
+	require.NoError(t, err)
+}
+
+func TestDecodeFelts(t *testing.T) {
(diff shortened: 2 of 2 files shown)


relayer/pkg/starknet/utils.go
```diff
@@ -77,6 +77,12 @@ func DecodeFelts(felts []*big.Int) ([]byte, error) {
 		bytesLen := Min(chunkSize, length)
 		bytesBuffer := buf[:bytesLen]
 
+		// TODO: this is inefficient because Bytes() and FillBytes() duplicate work
+		// reuse felt.Bytes()
+		if bytesLen < uint64(len(felt.Bytes())) {
+			return nil, errors.New("invalid: felt array can't be decoded")
+		}
```
archseer, line 84: I tried dropping `FillBytes()` and using something like `data = append(data, bytes...)` but that fails some of the fuzz tests, need to use an intermediate buffer to copy into

taco-paco: The issue here is in the fact that `0` are dropped when creating `big.Int`. For example:
```
data := []byte{0, 0, 0, 0, 0, 1, 2, 3}
felt := new(big.Int).SetBytes(data)
fmt.Println(len(felt.Bytes()))
```

Will output 3 while we assume that output shall be 8

taco-paco: Since this is the case in 2nd point from issue there's no way differentiate
`felts[1] = new(big.Int).SetBytes([]byte{0x1, 0x2, 0x3, 0x4})`
from
`felts[1] = new(big.Int).SetBytes([]byte{0x0, 0x0, ..., 0x1, 0x2, 0x3, 0x4})`

So I think those changes are ok. Or the other way to resolve this all would be enforce EncodeFelts to take byte array with size is `n*31`,

taco-paco (approved):

PR #155, opened 2022-10-06
Remove assert_not_zero which check if caller address is zero

### LINKED ISSUE #116: Add a test that verifies fees enforcement - can't invoke without using a funded account
In the context of issue https://github.com/smartcontractkit/chainlink-starknet/issues/115 we should add a test that verifies mandatory network fees. This will confirm an invoke can't be executed without using a funded account, so we can clean up `assert_not_zero` sender address checks in our contracts.

### LINKED ISSUE #118: Clean up 'assert_not_zero' sender address checks from contracts
In the context of issue https://github.com/smartcontractkit/chainlink-starknet/issues/116, which confirms an invoke can't be executed without using a funded account, we should clean up `assert_not_zero` sender address checks in our contracts which should be unnecessary.

qperrot:
Since invoke functions will need to be call by account with funds they can't be zero address.

### DIFF
--- a/contracts/src/chainlink/cairo/access/ownable.cairo
+++ b/contracts/src/chainlink/cairo/access/ownable.cairo
@@ -46,6 +46,7 @@ namespace Ownable:
     func assert_only_owner{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}():
         let (owner) = Ownable_owner.read()
         let (caller) = get_caller_address()
+        # caller is the zero address should not be possible anymore with introduction of fees
         with_attr error_message("Ownable: Caller is the zero address"):
             assert_not_zero(caller)
         end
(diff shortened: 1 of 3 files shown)


contracts/src/chainlink/cairo/access/ownable.cairo
```diff
@@ -46,9 +46,6 @@ namespace Ownable:
     func assert_only_owner{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}():
         let (owner) = Ownable_owner.read()
         let (caller) = get_caller_address()
-        with_attr error_message("Ownable: Caller is the zero address"):
```
krebernisak, line 49: cc @archseer we removed the zero address check.

@qperrot let's please a single test into `contracts/test/access/behavior/ownable.ts` that demonstrates that this is not possible with latest devnet: https://github.com/smartcontractkit/chainlink-starknet/issues/116

archseer: Do we need to remove these checks though? I'd rather keep the contract in sync with the upstream version: https://github.com/OpenZeppelin/cairo-contracts/blob/main/src/openzeppelin/access/ownable/library.cairo

krebernisak: Ok, if you have a strong preference here.

@qperrot let's keep the `assert_not_zero`, only in `ownable` contract, and add a comment hinting that this should not be possible anymore with introduction of fees.

Then we also add a test to demonstrate it's not possible.

archseer (approved):

PR #169, opened 2022-10-17
Check if observations sorted before taking median

krebernisak:


relayer/pkg/chainlink/ocr2/medianreport/report.go
```diff
@@ -102,22 +103,34 @@ func (c ReportCodec) MedianFromReport(report types.Report) (*big.Int, error) {
...
+		oo = append(oo, o)
+	}
+
+	// Check if the report contains sorted observations
+	_less := func(i, j int) bool {
+		return oo[i].Cmp(oo[j]) < 0
+	}
+	sorted := sort.SliceIsSorted(oo, _less)
+	if !sorted {
+		return nil, errors.New("observations not sorted")
```
aalu1418, line 130: test case? otherwise, this looks great!

krebernisak: Yeah, but it's more involved than adding a simple test case (BuildReport always sorts, custom fuzzer could be built) so I'll defer to this GitHub Issue for now: https://github.com/smartcontractkit/chainlink-starknet/issues/171

PR #172, opened 2022-10-20
Migrate code from Gateway to RPC API

archseer:


relayer/pkg/starknet/client.go
```diff
@@ -65,21 +70,14 @@ func NewClient(chainID string, baseURL string, lggr logger.Logger, timeout *time
...
-	client.set(baseURL, chainID) // hack: change the base URL & chainID (not supported in caigo)
-
-	return client, nil
-}
-
-func (c *Client) set(baseURL, chainID string) {
-	if chainID != "" {
-		c.Gw.Gateway.ChainId = chainID // note: gateway API in caigo does not query endpoint, uses what is set
+	// cache chainID on the provider to avoid repeated calls
+	chainID, err := client.Provider.ChainID(context.TODO())
```
archseer: Note: I did this to avoid `ChainID` calls, but I noticed that any time `NewAccount` is called, this happens internally in caigo too. Since the TXM creates accounts all the time, this would be called on any submit:

https://github.com/smartcontractkit/caigo/blob/56b86bbbc0c4a78bba811a5c25088c5d4184c031/account.go#L110-L114

We should patch this in caigo and cache on the provider instead

relayer/pkg/chainlink/ocr2/types_test.go
```diff
@@ -12,7 +12,8 @@ import (
 func StringsToCaigoFelts(in []string) []*caigotypes.Felt {
```
cfal: could this just be a `[]caigotypes.Felt` now?

cfal (dismissed):

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -204,21 +205,20 @@ func (txm *starktxm) confirmLoop() {
...
 						txm.lggr.Errorw("failed to fetch transaction status", "hash", hash, "error", err)
 						continue
 					}
-
-					if status == nil {
-						txm.lggr.Errorw("status was nil", "hash", hash)
-						continue
+					receipt, ok := response.(caigorpc.InvokeTransactionReceipt)
+					if !ok {
+						panic(fmt.Sprintf("wrong receipt type: %v", reflect.TypeOf(response)))
```
archseer: Note: should just use Errorw + continue

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -204,21 +205,20 @@ func (txm *starktxm) confirmLoop() {
...
 					}
 
-					if status.TxStatus == "ACCEPTED_ON_L1" || status.TxStatus == "ACCEPTED_ON_L2" || status.TxStatus == "REJECTED" {
-						txm.lggr.Debugw(fmt.Sprintf("tx confirmed: %s", status.TxStatus), "hash", hash, "status", status)
+					status := receipt.Status
+
+					if status == caigotypes.TransactionAcceptedOnL1 || status == caigotypes.TransactionAcceptedOnL2 || status == caigotypes.TransactionRejected {
```
archseer, line 226: @aalu1418 it seems weird to me that rejected is just treated as confirmed?

cfal: i think Aaron mentioned handling rejections would eventually be implemented: https://github.com/smartcontractkit/chainlink-starknet/pull/230#discussion_r1194061089

aalu1418: rejection is still technically an inclusion of the tx in the blockchain - if i understood correctly, rejection == reverts

.github/workflows/integration-tests-smoke.yml
```diff
@@ -62,7 +62,8 @@ jobs:
         uses: smartcontractkit/chainlink-github-actions/chainlink-testing-framework/build-image@2c9f401149f6c25fb632067b7e6626aebeee5d69
         with:
           cl_repo: smartcontractkit/chainlink
-          cl_ref: ${{ github.event.inputs.cl_branch_ref }}
+          # cl_ref: ${{ github.event.inputs.cl_branch_ref }}
+          cl_ref: 894c741feff0d6141d7dae9b49c60a9219feabf7
```
archseer: Needs https://github.com/smartcontractkit/chainlink/pull/9137 in core

.github/workflows/integration-tests-smoke.yml
```diff
@@ -62,7 +62,8 @@ jobs:
         uses: smartcontractkit/chainlink-github-actions/chainlink-testing-framework/build-image@2c9f401149f6c25fb632067b7e6626aebeee5d69
         with:
           cl_repo: smartcontractkit/chainlink
-          cl_ref: ${{ github.event.inputs.cl_branch_ref }}
+          # cl_ref: ${{ github.event.inputs.cl_branch_ref }}
+          cl_ref: f9a01f5a78de2fff2cbb63caddc550205b9391b8
```
cfal: does this need to be hardcoded?

contracts/requirements.txt
```diff
@@ -2,4 +2,4 @@ ecdsa
 fastecdsa
 sympy
 cairo-lang>=0.11.1.1
-starknet-devnet>=0.5.2
+starknet-devnet>=0.5.3
```
cfal, line 5: not sure if helpful but - 0.5.4 available now with JSONRPC updates https://github.com/0xSpaceShard/starknet-devnet/releases/tag/v0.5.4

cfal (dismissed):

.github/workflows/integration-tests-smoke.yml
archseer: I want to revert this but https://github.com/smartcontractkit/chainlink/pull/9137 needs merging first

PR #175, opened 2022-10-24
add start network in test and start hardhat node function

qperrot:


contracts/package.json
```diff
@@ -27,6 +27,7 @@
   },
   "dependencies": {
     "@chainlink/contracts": "^0.4.2",
+    "@chainlink/starknet-gauntlet": "*",
```
krebernisak: We don't want the contracts project to depend on Gauntlet packages at this point.

Please move the Network start/stop programatic capability to `@chainlink/starknet` package that you started with the account funder.

krebernisak (commented):
@qperrot let's merge the [account funder PR](https://github.com/smartcontractkit/chainlink-starknet/pull/144) first then rebase this one on top so it doesn't include those changes and will be easier to review.

packages-ts/starknet/src/account.ts
```diff
@@ -59,6 +60,7 @@ interface IFundingStrategy {
 // Fund the Account on Devnet
 class DevnetFundingStrategy implements IFundingStrategy {
   public async fund(accounts: FundAccounts[], opts: FunderOptions) {
+    const network = await startNetwork()
```
krebernisak: The `DevnetFundingStrategy` should not be controlling the network, but just deal with the funding and assume the network is up already.

packages-ts/starknet/src/network.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  protected async spawnChildProcess(): Promise<ChildProcess> {
+    let args = ['--port', this.port, '--gas-price', '1', '--lite-mode', '--timeout', '30000']
+    if (this.opts?.seed) {
+      args.push('--seed', this.opts.seed.toString())
+    }
+    return spawn(this.command, args)
+  }
+
+  protected cleanup(): void {
```
krebernisak: Make this a default and put it in the parent class which has access to the `childProcess`

packages-ts/starknet/src/network.ts
```diff
@@ -0,0 +1,122 @@
+import { ChildProcess, spawn } from 'child_process'
+import axios from 'axios'
+
+export abstract class IntegratedDevnet {
```
krebernisak: `IntegratedDevnet` is not the appropriate name for this class. Maybe something like `ChildProcessManager`?

packages-ts/starknet/src/network.ts
```diff
@@ -0,0 +1,122 @@
...
+  public stop() {
+    if (!this.childProcess) {
+      return
+    }
+
+    this.cleanup()
+  }
+}
+
+class VenvDevnet extends IntegratedDevnet {
```
krebernisak: There is no need for these `Venv` prefixes, as we don't interact with venvs in any way here. 
Using it to set up our env is just an implementation detail these classes should not know about.

```suggestion
class StarknetDevnet extends ChildProcessManager {
```

packages-ts/starknet/src/network.ts
```diff
@@ -0,0 +1,122 @@
...
+  protected async spawnChildProcess(): Promise<ChildProcess> {
+    return spawn('npx', this.command)
+  }
+
+  protected cleanup(): void {
+    this.childProcess?.kill()
+  }
+}
+
+export const startNetwork = async (opts?: {}): Promise<IntegratedDevnet> => {
```
krebernisak: Export the class and make this a static method there as just `HardHatNode.start(opts)`

packages-ts/integration-starkgate/test/tokenBridge.test.ts
```diff
@@ -33,6 +42,9 @@ describe('Test StarkGate token bridge + link_token.cairo', function () {
   let starknetERC20Bridge: Contract
 
   before(async () => {
+    devnetNetwork = await startNetwork()
```
krebernisak: We should do this similar as the funder, have some network starter class that gets the configuration for the test env and starts the appropriate networks. If a testnet is specified we won't start anything as the starter only knows to start StarkNetDevnet and HardHat.

So try to follow the Funder example and create a `NetworkManager` that reads in config/opts (we can read config from env as with Funder), and then we just do something like `manager.start(opts)` + `manager.stop()`.

krebernisak (dismissed):

packages-ts/starknet/src/network.ts
```diff
@@ -1,11 +1,51 @@
...
+    const network = process.env.NETWORK
+    const hardhatNetwork = process.env.NETWORK_ETHEREUM
+    if (network === DEVNET_NAME) {
+      this.strategyDevnet = new StarknetDevnet('5050', this.opts)
+    }
+    if (hardhatNetwork === HARDHAT_NAME) {
+      this.strategyHardhat = new HardHatNode('8545', this.opts)
+    }
+  }
+  // This function adds some funds to pre-deployed account that we are using in our test.
```
ivpavici: leftover comment I guess :)

packages-ts/starknet/src/network.ts
```diff
@@ -1,11 +1,51 @@
...
+      await this.strategyDevnet.startNetwork()
+    }
+    if (this.strategyHardhat) {
+      await this.strategyHardhat.startNode()
+    }
+  }
+
+  public stop() {
+    if (this.strategyDevnet) {
+      this.strategyDevnet.stop()
```
ivpavici: should this set `strategyDevnet` to null also, after stopping? Same question for `strategyHardhat`... Just a random thought, maybe not needed

contracts/test/emergency/StarkNetValidator.test.ts
```diff
@@ -16,19 +16,20 @@ import { deployMockContract, MockContract } from '@ethereum-waffle/mock-contract
...
-  IntegratedDevnet,
+  loadConfig,
+  NetworkManager,
+  FunderOptions,
+  Funder,
 } from '@chainlink/starknet'
 
 describe('StarkNetValidator', () => {
+  const config = loadConfig()
+  const optsConf = { config, required: ['devnet', 'hardhat'] }
```
krebernisak: This test shouldn't know much about the environment, the specifics of the networks it requires.

Instead, we should request network/s of specific blockchain type, and then the test gets what's configured in the env (depends how we run the test). It can be `devnet + hardhat` or for example `goerli + goerli_starknet', but we request for:
```suggestion
  const optsConf = { config, required: ['starknet', 'ethereum'] }
```

contracts/test/ocr2/aggregator_proxy.test.ts
```diff
@@ -4,20 +4,28 @@ import { number } from 'starknet'
...
+  const optsConf = { config, required: ['starknet'] }
+  const manager = new NetworkManager(optsConf)
+
   let aggregatorContractFactory: StarknetContractFactory
   let proxyContractFactory: StarknetContractFactory
 
   let owner: Account
   let aggregator: StarknetContract
   let proxy: StarknetContract
+  let opts: FunderOptions
```
krebernisak: We don't need these opts here as they are not used.

contracts/test/ocr2/aggregator.test.ts
```diff
@@ -57,8 +66,12 @@ function decodeBytes(felts: BN[]): Uint8Array {
 
 describe('aggregator.cairo', function () {
   this.timeout(TIMEOUT)
-  const opts = account.makeFunderOptsFromEnv()
-  const funder = new account.Funder(opts)
+  const config = loadConfig()
+  const optsConf = { config, required: ['starknet'] }
+  const manager = new NetworkManager(optsConf)
+
+  let opts: FunderOptions
```
krebernisak, line 73: We don't need these opts here as they are not used.

contracts/test/emergency/sequencer_uptime_feed.test.ts
```diff
@@ -3,17 +3,30 @@ import { starknet } from 'hardhat'
...
   this.timeout(300_000)
+  const config = loadConfig()
+  const optsConf = { config, required: ['starknet'] }
+  const manager = new NetworkManager(optsConf)
 
   let owner: Account
   let nonOwner: Account
-  const opts = account.makeFunderOptsFromEnv()
-  const funder = new account.Funder(opts)
+  let opts: FunderOptions
```
krebernisak: We don't need these opts here as they are not used.

contracts/test/token/ERC677/ERC677.test.ts
```diff
@@ -4,12 +4,16 @@ import { starknet } from 'hardhat'
...
 
 describe('ERC677', function () {
   this.timeout(TIMEOUT)
-  const opts = account.makeFunderOptsFromEnv()
-  const funder = new account.Funder(opts)
+  const config = loadConfig()
+  const optsConf = { config, required: ['starknet'] }
+  const manager = new NetworkManager(optsConf)
+
+  let opts: FunderOptions
```
krebernisak, line 15: We don't need these opts here as they are not used.

contracts/test/token/ERC677/link-receiver.test.ts
```diff
@@ -5,12 +5,16 @@ import { uint256 } from 'starknet'
...
 
 describe('LinkToken', function () {
   this.timeout(TIMEOUT)
-  const opts = account.makeFunderOptsFromEnv()
-  const funder = new account.Funder(opts)
+  const config = loadConfig()
+  const optsConf = { config, required: ['starknet'] }
+  const manager = new NetworkManager(optsConf)
+
+  let opts: FunderOptions
```
krebernisak, line 16: We don't need these opts here as they are not used.

examples/contracts/aggregator-consumer/test/mock/Aggregator_consumer.test.ts
```diff
@@ -1,18 +1,25 @@
...
 
 describe('ContractTestsMock', function () {
   this.timeout(600_000)
-  const opts = account.makeFunderOptsFromEnv()
-  const funder = new account.Funder(opts)
+  const config = loadConfig()
+  const optsConf = { config, required: ['starknet'] }
+  const manager = new NetworkManager(optsConf)
+
+  let opts: FunderOptions
```
krebernisak: We don't need these opts here as they are not used.

contracts/test/token/starkgate/link_token.test.ts
```diff
@@ -1,12 +1,22 @@
...
 
 describe('link_token', function () {
   this.timeout(TIMEOUT)
-  const opts = account.makeFunderOptsFromEnv()
-  const funder = new account.Funder(opts)
+  const config = loadConfig()
+  const optsConf = { config, required: ['starknet'] }
+  const manager = new NetworkManager(optsConf)
+
+  let opts: FunderOptions
```
krebernisak: We don't need these opts here as they are not used.

packages-ts/integration-eqlabs-multisig/test/Multisig.test.ts
```diff
@@ -3,13 +3,17 @@ import { starknet } from 'hardhat'
...
 describe('Multisig integration tests', function () {
   this.timeout(300_000)
 
-  const opts = account.makeFunderOptsFromEnv()
-  const funder = new account.Funder(opts)
+  const config = loadConfig()
+  const optsConf = { config, required: ['starknet'] }
+  const manager = new NetworkManager(optsConf)
+
+  let opts: FunderOptions
```
krebernisak: We don't need these opts here as they are not used.

packages-ts/integration-starkgate/test/tokenBridge.test.ts
```diff
@@ -10,16 +10,27 @@ import {
...
 const NAME = 'ChainLink Token'
 const SYMBOL = 'LINK'
 
 describe('Test StarkGate token bridge + link_token.cairo', function () {
   this.timeout(TIMEOUT)
+  const config = loadConfig()
+  const optsConf = { config, required: ['starknet', 'ethereum'] }
+  const manager = new NetworkManager(optsConf)
+
+  let opts: FunderOptions
```
krebernisak: We don't need these opts here as they are not used.

packages-ts/starknet/src/network.ts
```diff
@@ -0,0 +1,159 @@
...
+    if (this.opts?.required[0] === 'starknet') {
+      if (this.opts.config.starknet === DEVNET_NAME) {
+        this.strategyDevnet = new StarknetDevnet('5050', this.opts)
+      }
+    }
+    if (this.opts?.required[1] === 'ethereum') {
+      if (this.opts.config.ethereum === HARDHAT_NAME) {
+        this.strategyHardhat = new HardHatNode('8545', this.opts)
+      }
+    }
```
krebernisak: This implementation is prone to bugs...

We should make sure that all required networks are managed, or throw an error.
We should make sure that a network is declared only once, or throw an error.
We don't care about the order at which they are declared.

packages-ts/starknet/src/network.ts
```diff
@@ -0,0 +1,159 @@
+import { ChildProcess, spawn } from 'child_process'
+import net from 'net'
+import { DEVNET_NAME, HARDHAT_NAME } from './utils'
+
+export class NetworkManager {
+  private opts?: any
+  private strategyDevnet: StarknetDevnet
+  private strategyHardhat: HardHatNode
```
krebernisak: nit: don't need the 'strategy' prefix here

packages-ts/starknet/src/network.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  constructor(opts?: any) {
+    this.opts = opts
+    if (this.opts?.required[0] === 'starknet') {
```
krebernisak: Order should not be hardcoded here.

packages-ts/starknet/src/network.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    this.cleanup()
+  }
+}
+
+class StarknetDevnet extends ChildProcessManager {
+  private command: string
+  private opts?: any
+
+  constructor(port: string, opts?: any) {
```
krebernisak: These opts are not used here - we should put the port and any other options in there.

packages-ts/starknet/src/network.ts
```diff
@@ -0,0 +1,159 @@
...
+    // Starting to poll devnet too soon can result in ENOENT
+    await new Promise((f) => setTimeout(f, 4000))
+  }
+}
+
+class HardHatNode extends ChildProcessManager {
+  private command: string[]
+  private opts?: any
+
+  constructor(port: string, opts?: any) {
```
krebernisak: These opts are not used here - we should put the port and any other options in there.

packages-ts/starknet/src/network.ts
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  public async startNode(): Promise<void> {
+    if (!(await this.isFreePort(parseInt(this.port)))) {
+      return
+    }
+    await this.start()
+
+    // Starting to poll hardhatNode too soon can result in ENOENT
+    await new Promise((f) => setTimeout(f, 10000))
```
krebernisak: What's the minimum safe value we can get those at?

packages-ts/starknet/src/network.ts
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  public async startNetwork(): Promise<void> {
+    if (!(await this.isFreePort(parseInt(this.port)))) {
+      return
+    }
+    await this.start()
+
+    // Starting to poll devnet too soon can result in ENOENT
+    await new Promise((f) => setTimeout(f, 4000))
```
krebernisak: What's the minimum safe value we can get those at?

packages-ts/starknet/src/network.ts
krebernisak: It should be ok to keep them like this, as we constructed them in the constructor and here we just manage their lifecycle. We should be able to start them again, after stopping, using the same manager.

PR #177, opened 2022-11-01
Cairo 0.10.1 + devnet 0.3.5

archseer:


packages-ts/starknet-gauntlet-ledger/src/index.ts
```diff
@@ -76,6 +77,16 @@ class LedgerSigner implements SignerInterface {
...
 
+  async signDeployAccountTransaction(transaction: DeployAccountSignerDetails): Promise<Signature> {
+    // TODO: implement this
+    return []
+  }
+
+  async signDeclareTransaction(transaction: DeclareSignerDetails): Promise<Signature> {
+    // TODO: implement this
+    return []
+  }
```
archseer, line 88: @sdrug A couple of TODOs here since I couldn't find an upstream implementation in the ledger repo. Perhaps we can base it on `starknet.js` internals

sdrug: Noted 👍

packages-ts/starknet-gauntlet-ledger/src/index.ts
```diff
@@ -13,6 +13,7 @@ import { Stark, LedgerError } from '@ledgerhq/hw-app-starknet'
 export const DEFAULT_LEDGER_PATH = "m/2645'/579218131'/894929996'/0'"
 export const LEDGER_PATH_REGEX = /^\s*m\s*\/\s*2645\s*\'\s*\/\s*579218131\s*\'\s*\/\s*(\d+)\s*\'\s*\/\s*(\d+)\s*\'$/
 
+// TODO: why are we copying this rather than using https://github.com/yogh333/starknetjs-signer-ledger/blob/main/src/ledger-signer.ts
```
sdrug, line 16: Good point! 
We implemented the signer interface before they got it upstream.
It diverges a little with what they have now. 
I'll test it out with the real device and their new c++ app and switch if ok.

sdrug (commented):
OZ Account test still fails it seems

archseer:
Yeah I need to update the `class_hash` again

sdrug (approved):

PR #178, opened 2022-11-08
Monitoring followups: docs, relay bump, dockerfile, docker-compose

topliceanu:


monitoring/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Useful links
+
+- Starknet on-chain monitor [generated docs](https://pkg.go.dev/github.com/smartcontractkit/chainlink-starknet/monitoring/pkg/monitoring).
+- On-chain monitoring (OM) framework architecture docs in [blueprints](https://github.com/smartcontractkit/chainlink-blueprints/blob/master/monitoring/README.md).
+- OM framework [generated docs](https://pkg.go.dev/github.com/smartcontractkit/chainlink-relay/pkg/monitoring).
+
+## Local development
+
+- Start the monitor's third party dependencies using [docker-compose](https://docs.docker.com/compose/).
```
krebernisak, line 11: Ok for this PR, but we're diverging our infra env between using Docker directly and k8s via Helm.

We'll need to streamline our infra setup to make it less complex.

krebernisak (approved):

monitoring/README.md
topliceanu: I agree! ideally the monitor would be integrated in the local development environment which uses k8s and helm. Though I'm not sure this is going to be "less complex"!

krebernisak: We'll work on that next!

Less complex as in consistent and slightly hidden. Just so we focus on a single env for all.

PR #181, opened 2022-11-17
Added range checks in event parsers. Issue #180

### LINKED ISSUE #180: Handle overflows in event parsers
Range of `big.Int`s isn't checked. So [following cases](https://github.com/smartcontractkit/chainlink-starknet/blob/a132881f93a6806705e9b53c438cc10e260f0c19/relayer/pkg/chainlink/ocr2/events.go#L61) may overflow

taco-paco:


relayer/pkg/chainlink/ocr2/events.go
```diff
@@ -40,6 +50,14 @@ type NewTransmissionEvent struct {
 	Reimbursement   *big.Int
 }
 
+func rangeCheck(felt *caigotypes.Felt, lowerBound *big.Int, upperBound *big.Int) error {
+	if !(felt.Int.Cmp(lowerBound) >= 0 && felt.Int.Cmp(upperBound) <= 0) {
+		return errors.New("invalid: value is out of range")
+	}
+
+	return nil
+}
```
krebernisak: nit: let's structure this utility as an object we can reuse:
```go
type Range struct {
	Lower *big.Int
	Upper *big.Int
}

// Notice: we don't pass in Felt here and we keep this fn more generic
func (r *Range) Check(n *big.Int) (error) {
        if !(n.Int.Cmp(r.Lower) >= 0 && n.Int.Cmp(r.Upper) <= 0) {
		return fmt.Errorf("invalid: value %v is out of range [%v,%v]", n, r.Lower, r.Upper)
	}

	return nil
}
```

Which will allow for less boilerplate:
```go
err := uint64Range.Check(eventData[observationsLenIndex])
if err != nil {
    return NewTransmissionEvent{}, err
}
```

krebernisak (dismissed):

relayer/pkg/chainlink/ocr2/events.go
taco-paco: Done

PR #183, opened 2022-11-21
Run gauntlet binary release and relayer code release on workflow_dispatch

kylesmartin:
Currently, the changesets action creates tags that trigger the gauntlet and relayer release workflows. However, since `secrets.GITHUB_TOKEN` can only trigger other workflows by making explicit dispatch calls, we would need to configure a custom personal access token or a deployment token for the release workflows to run on tag creation.

This PR makes the changesets action simply dispatch the release workflows, removing the requirement for a custom token.

NOTE: We still need to configure `secrets.NPM_TOKEN` if we have not already.

### DIFF
--- a/.github/workflows/release/starknet-relayer.yml
+++ b/.github/workflows/release/starknet-relayer.yml
@@ -16,7 +16,7 @@ jobs:
         run: echo "STARKNET_RELAYER=$(npm info @chainlink/starknet-relayer version)" >> $GITHUB_ENV
       # Check if release tag exists
       - name: Check release tag
-        uses: mukunku/tag-exists-action@v1.0.0
+        uses: mukunku/tag-exists-action@5dfe2bf779fe5259360bb10b2041676713dcc8a3 # v1.1.0
         id: checkTag
         with:
           tag: relayer/v${{ env.STARKNET_RELAYER }}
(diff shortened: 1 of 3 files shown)


.github/workflows/release/starknet-relayer.yml
```diff
@@ -16,8 +14,17 @@ jobs:
       # Store starknet version
       - name: Set Env Variables
         run: echo "STARKNET_RELAYER=$(npm info @chainlink/starknet-relayer version)" >> $GITHUB_ENV
+      # Check if release tag exists
+      - name: Check release tag
+        uses: mukunku/tag-exists-action@v1.0.0
```
chainchad:
```suggestion
        uses: mukunku/tag-exists-action@5dfe2bf779fe5259360bb10b2041676713dcc8a3 # v1.1.0
```

.github/workflows/release/starknet-relayer.yml
```diff
@@ -16,8 +14,17 @@ jobs:
...
+        id: checkTag
+        with:
+          tag: relayer/v${{ env.STARKNET_RELAYER }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       # Release code under vX.X.X
       - name: Release Code
+        if: steps.checkTag.outputs.exists == 'false'
         uses: softprops/action-gh-release@v1
```
chainchad:
```suggestion
        uses: softprops/action-gh-release@de2c0eb89ae2a093876385947365aca7b0e5f844 # v1
```

chainchad (dismissed):
Looks good, just have two suggestions to pin to the GHA using a SHA.

chainchad (approved):

Issue #122, opened 2022-09-21
Add golangci-lint support for all Go modules in the monorepo


krebernisak:
We should include all Go modules in the monorepo: https://github.com/golangci/golangci-lint/issues/828#issuecomment-658207652

_Originally posted by @krebernisak in https://github.com/smartcontractkit/chainlink-starknet/pull/121#discussion_r976162543_

aalu1418:
completed in: https://github.com/smartcontractkit/chainlink-starknet/pull/189

PR #189, opened 2022-12-07
add go linting to ci

tateexon:


archseer:
We already have linting here: https://github.com/smartcontractkit/chainlink-starknet/blob/54d7e784fe0f1cc0a258aeee38b554c9e682048b/.github/workflows/golangci-lint.yml

archseer:
But looks like it was only configured to run under `relayer/`

aalu1418:
> But looks like it was only configured to run under `relayer/`

I think this line also might be causing the different results in the original linting set up and the linting that tate added
https://github.com/smartcontractkit/chainlink-starknet/blob/54d7e784fe0f1cc0a258aeee38b554c9e682048b/.github/workflows/golangci-lint.yml#L32

.golangci.yml
```diff
@@ -5,10 +5,10 @@ linters:
     - exhaustive
     - exportloopref
     - revive
-    - goimports
+    # - goimports
     - gosec
     - misspell
-    - rowserrcheck
```
aalu1418, line 11: just curious, why are we not running these two anymore?

tateexon: rowserrcheck is not viable under go 1.19
goimports was just causing golangci-lint to error out and not return results because it doesn't like how the project is setup.

This draft pr is not meant to be pushed, was just using it to show we are missing stuff. As Blaz pointed out we do have a golangci-lint workflow already but it must just be ignoring the errors that exist in relayer or is just not setup correctly.

aalu1418: i think it's still worth pushing this PR! it addresses an open issue - https://github.com/smartcontractkit/chainlink-starknet/issues/122

.github/workflows/golangci-lint.yml
```diff
@@ -4,35 +4,83 @@ on:
...
+      - name: Install Nix
+        uses: cachix/install-nix-action@daddc62a2e67d1decb56e028c9fa68344b9b7c2a # v18
+        with:
+          nix_path: nixpkgs=channel:nixos-unstable
+      - name: Lint integration-tests
+        run: nix develop -c make lint-go-test
+      # Note: I could not figure out why the golangci-lint-action would not work even though it is technically running the same as above, error message is this:
+      # Running [/home/runner/golangci-lint-1.50.1-linux-amd64/golangci-lint run --out-format=github-actions --path-prefix=integration-tests --exclude=dot-imports] in [/home/runner/work/chainlink-starknet/chainlink-starknet/integration-tests] ...
+      # level=warning msg="[runner] Can't run linter goanalysis_metalinter: inspect: failed to load package client: could not load export data: no export data for \"github.com/smartcontractkit/chainlink-testing-framework/client\""
+      # level=error msg="Running error: 1 error occurred:\n\t* can't run linter goanalysis_metalinter: inspect: failed to load package client: could not load export data: no export data for \"github.com/smartcontractkit/chainlink-testing-framework/client\"\n\n"
```
aalu1418, line 78: I have a guess on this...do you think the reference in `go.mod` to the chainlink-testing-framework was dropped? 

if you check: https://github.com/smartcontractkit/chainlink-testing-framework/tree/d40b562b818b
it shows
>This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.

so maybe it's not findable anymore 🤷🏼

aalu1418: let me see if i can fix it

aalu1418: nevermind - my idea to fix didn't work

aalu1418 (approved):
🎉 - closes #122

PR #198, opened 2023-01-07
Add examples for docs.chain.link

dwightjl:
- Create a `proxy_consumer` example with simple on-chain deployment examples that require minimal configuration.
- Create an off-chain read example that uses StarkNet.js
- (Optional) Add an ABI for aggregator_proxy.cairo to simplify examples for off-chain reads using `starknet call`.


archseer:
Looks like the lint check is failing, you should `yarn format` the code

examples/contracts/proxy-consumer/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+### Requirements
+
+- Install [Python 3.9](https://www.python.org/downloads/).
+- Install [Yarn](https://classic.yarnpkg.com/lang/en/docs/install/).
+- [Setup your local StarkNet environment](https://starknet.io/docs/quickstart.html).
+- [Set up a StarkNet account](https://starknet.io/docs/hello_starknet/account_setup.html) on Starknet's `alpha-goerli` network and fund it with [testnet ETH](https://faucet.goerli.starknet.io/).
+
+### Running the on-chain example
+
```
archseer, line 16: Could we use `yarn build` to match the other packages?

archseer (commented):
Is the proxy ABI needed for the examples? I'd move it under the examples folder for now since we'll switch to publishing build artifacts via releases later.

You can also remove the built json artifacts since the user will compile them as a part of the tutorial

examples/contracts/proxy-consumer/README.md
dwightjl: - Moved the ABI to the contracts folder in the new example. Docs really only need it for simplicity. It's not a requirement.
- Removed the `./starknet-artifacts/`.
- Changed `yarn compile` to `yarn build`.
- Updated the README.md

dwightjl:
> Looks like the lint check is failing, you should `yarn format` the code

Formatting fixed. Do the integration tests also require a fix? It seems to be outside the scope of this PR.

archseer (approved):
Thanks! Looks good to me, integration tests are currently failing on develop too

PR #207, opened 2023-02-07
TT-69 Isomorphic Integration Test Deployments

tateexon:
Updates smoke and soak tests to use the isomorphic deployment strategy for CI.
Reuses nix for the test image build.
In CI build the contracts in parallel to the chainlink image to speed up the pipeline.
Use larger runners to reduce build time.
Update nix to add missing pip and helm chart dependencies.
Add make calls to run nix in a container if it is not behaving on mac.


shell.nix
```diff
@@ -35,7 +36,9 @@
   venvDir = "./.venv";
 
   postShellHook = ''
-    pip install --upgrade pip
     pip install -r ${./contracts/requirements.txt} -c ${./contracts/constraints.txt}
+    helm repo add chainlink-qa https://raw.githubusercontent.com/smartcontractkit/qa-charts/gh-pages/
+    helm repo add bitnami https://charts.bitnami.com/bitnami
+    helm repo update
```
archseer: An idea: add the `~/.config/helm/repositories.yaml` file to the project, then set `HELM_REPOSITORY_CONFIG` in this `shell.nix` to point to the file. That way it's part of the project and doesn't require running shell hooks

integration-tests/test.Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM nixos/nix:latest
```
archseer, line 1: ❤️

tateexon:
Well got the e2e tests to pass but somehow broke the contracts tests that I didn't think I touched in any way? Gonna have to figure that out tomorrow.

shell.nix
tateexon: Still had to leave in helm repo update or we wouldn't pull in the cache.

archseer:
Contract tests seem to have broken on their own with a devnet update I think, it's now broken on other branches/master

kalverra (approved):

PR #217, opened 2023-03-17
bump chainlink-relay; update relayer.Health()

jmank88:
`chainSet.Healthy()` is no longer available on latest chainlink-relay

Smoke tests depend on -solana, but fortunately they are not required b/c it depends on this also:
- https://github.com/smartcontractkit/chainlink-solana/pull/484


essamhassan (approved):
thanks Jordan!

relayer/pkg/chainlink/relay.go
```diff
@@ -51,9 +51,7 @@ func (r *relayer) Ready() error {
 	return r.chainSet.Ready()
 }
 
-func (r *relayer) Healthy() error {
-	return r.chainSet.Healthy()
-}
+func (r *relayer) Healthy() error { return nil }
```
essamhassan, line 54: Any reason why we still need this and not remove all of it?

essamhassan: same question for solana

jmank88: I was just trying to do the minimum to fix things and thought that might create more of the same problem :shrug:

PR #218, opened 2023-03-17
Update starknet.js/starknet-devnet/starknet-hardhat-plugin

archseer:
BCI-986


contracts/requirements.txt
```diff
@@ -2,5 +2,5 @@ ecdsa
 fastecdsa
 sympy
 cairo-lang==0.10.3
-starknet-devnet>=0.4.2
+starknet-devnet>=0.4.6
```
augustbleeds: Starknet just released https://github.com/Shard-Labs/starknet-devnet/releases 0.5.0a0 which supports starknet 0.11 and cairo 1.0 but I believe right now they only support x86

packages-ts/starknet-gauntlet-ocr2/src/lib/encoding.ts
```diff
@@ -20,26 +20,28 @@ export function bytesToFelts(data: Uint8Array): string[] {
...
+  //   let chunk = felt.toArray('be', Math.min(CHUNK_SIZE, length))
+  //   data.push(...chunk)
 
-    length -= chunk.length
-  }
+  //   length -= chunk.length
+  // }
 
   return Buffer.from(data)
 }
```
archseer: Note: Still need to fix this

contracts/src/chainlink/solidity/emergency/StarknetValidator.sol
```diff
@@ -123,15 +133,15 @@ contract StarknetValidator is TypeAndVersionInterface, AggregatorValidatorInterf
...
+  ) external payable override checkAccess returns (bool) {
     return _sendUpdateMessageToL2(currentAnswer);
   }
 
   /**
    * @notice retries to send the latest answer as update message to L2
    * @dev only with access, useful in cases where a previous x-domain message was handeled unsuccessfully.
    */
-  function retry() external checkAccess returns (bool) {
+  function retry() external payable checkAccess returns (bool) {
```
augustbleeds, line 144: Based on the other L2 validators (and the deposit/withdraw function defined here) I don't think the Aggregator is supposed to fund the fees. The validator is meant to withdraw the fees from it's own balance. So I think to get the tests to pass, just deposit some eth to this contract (it has a payable receive() method)

augustbleeds: Example on Arbitrum https://github.com/smartcontractkit/chainlink/blob/c8d7632bd817ca993e40862d53be7910ff38d34c/contracts/src/v0.8/dev/ArbitrumValidator.sol

archseer: Hmm could you take care of this in a follow-up PR? I just want to unblock merging the dependency bumps

archseer: I can also exclude these changes and let it fail on that case for now

PR #219, opened 2023-03-18
pkg/chainlink: remove unused config update

jmank88:
This is no longer used with static TOML config.


aalu1418 (approved):
:killitwithfire:

PR #221, opened 2023-03-24
relayer/pkg/chainlink: remove legacy config

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-2151

For compatability, merge after:
- https://github.com/smartcontractkit/chainlink/pull/8830

### DIFF
--- a/relayer/pkg/chainlink/config/config.go
+++ b/relayer/pkg/chainlink/config/config.go
@@ -1,13 +1,10 @@
 package config
 
 import (
-	"net/url"
 	"time"
 
-	"github.com/smartcontractkit/chainlink-relay/pkg/logger"
 	"github.com/smartcontractkit/chainlink-relay/pkg/utils"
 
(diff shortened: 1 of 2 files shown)


relayer/pkg/chainlink/config/config.go
```diff
@@ -44,70 +41,6 @@ type Config interface {
...
-
-func NewConfig(dbCfg db.ChainCfg, lggr logger.Logger) *config {
-	return &config{
-		defaults: DefaultConfigSet,
-		dbCfg:    dbCfg,
-		lggr:     lggr,
-	}
-}
-
-func (c *config) OCR2CachePollPeriod() time.Duration {
```
aalu1418, line 63: just wondering, where do these get implemented now?

jmank88: The same types that are used to decode the TOML

jmank88: They should be moved in to these repos eventually.

archseer (approved):

PR #224, opened 2023-04-03
Cairo 1.0

archseer:
Staging branch


contracts/src/confirmed_owner.cairo
```diff
@@ -0,0 +1,31 @@
+#[contract]
```
augustbleeds: I created this file assuming we would run eventually tests against contracts that are not deployed (testing the specific functionality of each parent contract by itself). It doesn't seem like we'll be doing that so we can probably delete this file.

contracts/src/ocr2/aggregator_proxy.cairo
```diff
@@ -0,0 +1,286 @@
...
+    #[view]
+    fn proposed_round_data(round_id: felt252) -> Round {
+        _require_access();
+        let address = _proposed_aggregator::read();
+        // TODO: zero address check?
+        let round_id128: u128 = round_id.try_into().unwrap();
+        let aggregator = IAggregatorDispatcher { contract_address: address };
+        aggregator.round_data(round_id128)
+    }
+
```
augustbleeds, line 271: We're missing the `round_data` and `latest_round_data` view functions definitions

contracts/src/ocr2/aggregator_proxy.cairo
```diff
@@ -0,0 +1,286 @@
...
+    #[event]
+    fn AggregatorConfirmed(previous: ContractAddress, latest: ContractAddress) {}
+
+    impl AggregatorProxy of IAggregatorProxy {
+        fn latest_round_data() -> Round {
+            let phase = _current_phase::read();
+            let aggregator = IAggregatorDispatcher { contract_address: phase.aggregator };
+            let round = aggregator.latest_round_data();
+            Round {
+                round_id: (phase.id.into() * SHIFT) + round.round_id,
```
augustbleeds, line 110: so higher 128 bits are for phase id and lower 124 bits for round id?

contracts/src/ocr2/aggregator_proxy.cairo
```diff
@@ -0,0 +1,286 @@
...
+                answer: round.answer,
+                block_num: round.block_num,
+                started_at: round.started_at,
+                updated_at: round.updated_at,
+            }
+        }
+        fn round_data(round_id: felt252) -> Round {
+            // TODO: panic if Narrow since phase_id should never be 0
+            let (phase_id, round_id) = match u128s_from_felt252(round_id) {
```
augustbleeds: nice, this is neat!

augustbleeds (commented):
some comments

contracts/src/ocr2/aggregator_proxy.cairo
archseer: The other way around actually, upper 124 are phase_id (since we will never need that many) and round_id is a full u128

contracts/src/multisig.cairo
```diff
@@ -0,0 +1,501 @@
...
+impl TDrop: Drop::<T>,
+impl TPartialEq: PartialEq::<T>,
+>(
+    a: @Array::<T>, len: usize, j: usize, k: usize
+) {
+    if j >= len {
+        return ();
+    }
+    if k >= len {
+        gas::withdraw_gas_all(get_builtin_costs()).expect('Out of gas');
```
augustbleeds, line 26: What's the advantage of calling the gas::withdraw_gas_all twice? I thought we only need to call to withdraw_gas_all once in the beginning of the function. (ref: https://medium.com/nethermind-eth/under-the-hood-of-cairo-1-0-exploring-sierra-7f32808421f5). So then it calculates the gas cost of the entire function at once and subtracts it from the local gas wallet.

contracts/src/multisig.cairo
```diff
@@ -0,0 +1,501 @@
...
+        assert(nonce < next_nonce, 'transaction does not exist');
+    }
+
+    fn _require_not_executed(nonce: u128) {
+        let transaction = _transactions::read(nonce);
+        assert(!transaction.executed, 'transaction already executed');
+    }
+
+    fn _require_not_confirmed(nonce: u128) {
+        let caller = get_caller_address();
```
augustbleeds, line 475: nit: save some code duplication by having one method 

fn _caller_confirmed() -> bool {
      let caller = get_caller_address();
       _is_confirmed::read((nonce, caller));
}

and then no need to define `_require_not_confirmed` and `_require_confirmed` because you can just check the assert value in the other functions

contracts/src/multisig.cairo
```diff
@@ -0,0 +1,501 @@
+use array::ArrayTrait;
```
augustbleeds, line 1: finally, got around to reviewing this. was fun to read! great job alex

contracts/src/multisig.cairo
archseer: Hmm I've always put it at the end of the function (just before the recursive call) since that seemed to be what they do in the examples. We can soon remove this because `alpha.7` will automatically add these

cfal: yeah, i think there used to be sample code in the cairo repo that was directly in front of the recursive call, so i did the same here. withdraw_gas_all is only called once before each recursive call, per iteration

contracts/src/ocr2/aggregator_proxy.cairo
cfal: thanks!

contracts/src/access_control/access_controller.cairo
```diff
@@ -0,0 +1,88 @@
...
+        Ownable::renounce_ownership();
+    }
+
+    ///
+    /// Upgradeable
+    ///
+
+    #[view]
+    fn type_and_version() -> felt252 {
+        'WriteAccessController 1.0.0'
```
archseer:
```suggestion
        'AccessController 1.0.0'
```

Makefile
```diff
@@ -138,16 +140,17 @@ format-go-mod-tidy:
...
 format-cairo:
 	find ./contracts/src -name "*.cairo" -type f \
-		-exec cairo-format -i --one_item_per_line {} +
+		-exec cairo-format -i {} +
 	find ./examples -name "*.cairo" -type f \
-		-exec cairo-format -i --one_item_per_line {} +
+		-exec cairo-format -i {} +
 
 .PHONY: format-cairo-check
 format-cairo-check:
```
augustbleeds, line 148: fyi: for me this check fails as is
```
find ./contracts/src -name "*.cairo" -type f \
                -exec cairo-format -c {} +
./contracts/src/token/mock/invalid_erc667_receiver.cairo:1:1: Unexpected character "#".
#[contract]
^
make: *** [format-cairo-check] Error 1
```

augustbleeds: not sure why

contracts/src/account/erc165.cairo
```diff
@@ -0,0 +1,45 @@
+// copied from https://github.com/OpenZeppelin/cairo-contracts/pull/583
```
augustbleeds, line 1: @cfal can you pin to the specific commit hash? the PR may change/update

contracts/src/account/account.cairo
```diff
@@ -0,0 +1,195 @@
+// copied from https://github.com/OpenZeppelin/cairo-contracts/pull/616
```
augustbleeds, line 1: @cfal can you pin to the specific commit hash? the PR may change/update

Makefile
archseer: Could be a different `cairo-format` version?

contracts/src/account/erc165.cairo
archseer: I attempted to get a commit for both this + account but the PR has since been changed. I think we can wait for it to get merged into master. (BTW account is actually from https://github.com/OpenZeppelin/cairo-contracts/pull/616 now)

Makefile
augustbleeds: when i do `which cairo-format` it points to the gitsubmodule so I think it should be correct 🤔

contracts/src/tests/test_aggregator.cairo
```diff
@@ -0,0 +1,90 @@
...
+use traits::Into;
+use traits::TryInto;
+use option::OptionTrait;
+use core::result::ResultTrait;
+
+use chainlink::ocr2::aggregator::pow;
+use chainlink::ocr2::aggregator::Aggregator;
+use chainlink::tests::test_ownable::should_implement_ownable;
+
+// TODO: aggregator tests
```
augustbleeds, line 21: can we remove this todo? @cfal cuz we decided we're keeping hardhat tests for this contract since they are hard to reproduce in cairo?

contracts/src/tests/test_erc677.cairo
```diff
@@ -0,0 +1,115 @@
...
+}
+
+#[abi]
+trait MockInvalidReceiver {
+    fn supports_interface(interface_id: u32) -> bool;
+    fn set_supports(value: bool);
+}
+
+
+// Ignored tests are dependent on upgrading our version of cairo to include this PR https://github.com/starkware-libs/cairo/pull/2912/files
```
augustbleeds, line 32: I think we can remove this comment because there are no more ignored tests in this file

contracts/src/tests/test_multisig.cairo
```diff
@@ -0,0 +1,291 @@
+use starknet::contract_address_const;
+use starknet::testing::set_caller_address;
+use starknet::class_hash::class_hash_const;
+
+use array::ArrayTrait;
+
+use chainlink::multisig::assert_unique_values;
+use chainlink::multisig::Multisig;
+
+// TODO: test execute_confirmation happy path with mocked
```
augustbleeds, line 10: What should we do about these todos?

Makefile
cfal: that's the error returned by python cairo-lang's cairo-format:

```
$ ~/.asdf/shims/cairo-format ./contracts/src/token/mock/invalid_erc667_receiver.cairo
./contracts/src/token/mock/invalid_erc667_receiver.cairo:1:1: Unexpected character "#".
#[contract]
^
```

PR #228, opened 2023-04-12
BCI-990: Update er protocol to Cairo 1.0

augustbleeds:
Please see inline comments for more info.

This PR updates ER Protocol to version cairo 1.0

Note I also added the external functions for SimpleWriteAccessController in this PR

### DIFF
--- a/contracts/src/emergency/sequencer_uptime_feed.cairo
+++ b/contracts/src/emergency/sequencer_uptime_feed.cairo
@@ -12,7 +12,7 @@ impl Felt252TryIntoBool of TryInto::<felt252, bool> {
 
 impl BoolIntoFelt252 of Into::<bool, felt252> {
     fn into(self: bool) -> felt252 {
-        if (self) {
+        if self {
             return 1;
         }
         return 0;
@@ -41,11 +41,11 @@ mod SequencerUptimeFeed {
     use chainlink::libraries::simple_read_access_controller::SimpleReadAccessController;
     use chainlink::libraries::simple_write_access_controller::SimpleWriteAccessController;
     use chainlink::ocr2::aggregator::Round;
-    use chainlink::ocr2::aggregator::IAggregator; 
+    use chainlink::ocr2::aggregator::IAggregator;
     use chainlink::ocr2::aggregator::Aggregator::Transmission;
     use chainlink::ocr2::aggregator::Aggregator::TransmissionStorageAccess;
 
-    const ETH_ADDRESS_BOUND: felt252  = 0x10000000000000000000000000000000000000000; // 2**160
+    const ETH_ADDRESS_BOUND: felt252 = 0x10000000000000000000000000000000000000000; // 2**160
 
     struct Storage {
         // l1 sender is an ethereum address
@@ -65,7 +65,9 @@ mod SequencerUptimeFeed {
     fn AnswerUpdated(current: u128, round_id: u128, timestamp: u64) {}
 
     #[event]
-    fn UpdateIgnored(latest_status: u128, latest_timestamp: u64, incoming_status: u128, incoming_timestamp: u64) {}
+    fn UpdateIgnored(
+        latest_status: u128, latest_timestamp: u64, incoming_status: u128, incoming_timestamp: u64
+    ) {}
 
     #[event]
     fn L1SenderTransferred(from_address: felt252, to_address: felt252) {}
@@ -80,7 +82,7 @@ mod SequencerUptimeFeed {
                 answer: round_transmission.answer,
                 block_num: round_transmission.block_num,
                 started_at: round_transmission.observation_timestamp,
-                updated_at: round_transmission.transmission_timestamp,   
+                updated_at: round_transmission.transmission_timestamp,
             }
         }
 
@@ -93,7 +95,7 @@ mod SequencerUptimeFeed {
                 answer: round_transmission.answer,
                 block_num: round_transmission.block_num,
                 started_at: round_transmission.observation_timestamp,
-                updated_at: round_transmission.transmission_timestamp,   
+                updated_at: round_transmission.transmission_timestamp,
             }
         }
 
@@ -122,12 +124,14 @@ mod SequencerUptimeFeed {
         let latest_round_id = _latest_round_id::read();
         let latest_round = _round_transmissions::read(latest_round_id);
 
-        if (timestamp <= latest_round.observation_timestamp) {
-            UpdateIgnored(latest_round.answer, latest_round.transmission_timestamp, status, timestamp);
+        if timestamp <= latest_round.observation_timestamp {
+            UpdateIgnored(
+                latest_round.answer, latest_round.transmission_timestamp, status, timestamp
+            );
             return ();
         }
 
-        if (latest_round.answer == status) {
+        if latest_round.answer == status {
             _update_round(latest_round_id, latest_round);
         } else {
             // only increment round when status flips
@@ -142,12 +146,12 @@ mod SequencerUptimeFeed {
 
         // convert both to u256 (felts don't implement PartialOrd)
         assert(address.into() < ETH_ADDRESS_BOUND.into(), 'invalid eth address');
-    
+
         assert(address != 0, '0 address not allowed');
 
         let old_address = _l1_sender::read();
 
-        if (old_address != address) {
+        if old_address != address {
             _l1_sender::write(address);
             L1SenderTransferred(old_address, address);
         }
@@ -226,7 +230,7 @@ mod SequencerUptimeFeed {
         SimpleReadAccessController::has_access(user, data)
     }
 
-     #[external]
+    #[external]
     fn check_access(user: ContractAddress) {
         SimpleReadAccessController::check_access(user)
     }
@@ -245,7 +249,7 @@ mod SequencerUptimeFeed {
         SimpleWriteAccessController::remove_access(user)
     }
 
-     #[external]
+    #[external]
     fn enable_access_check() {
         SimpleWriteAccessController::enable_access_check()
     }
@@ -262,7 +266,7 @@ mod SequencerUptimeFeed {
 
     fn require_access() {
         let sender = starknet::info::get_caller_address();
-         SimpleReadAccessController::check_access(sender);
+        SimpleReadAccessController::check_access(sender);
     }
 
     fn initializer(initial_status: u128, owner_address: ContractAddress) {
--- a/contracts/src/emergency/sequencer_uptime_feed.cairo
+++ b/contracts/src/emergency/sequencer_uptime_feed.cairo
@@ -1,28 +1,5 @@
-impl Felt252TryIntoBool of TryInto::<felt252, bool> {
-    fn try_into(self: felt252) -> Option<bool> {
-        if self == 0 {
-            Option::Some(false);
-        }
-        if self == 1 {
-            Option::Some(true);
-        }
-        Option::None(())
-    }
-}
-
-impl BoolIntoFelt252 of Into::<bool, felt252> {
-    fn into(self: bool) -> felt252 {
-        if self {
-            return 1;
-        }
-        return 0;
-    }
-}
-
 #[contract]
 mod SequencerUptimeFeed {
-    use super::Felt252TryIntoBool;
-    use super::BoolIntoFelt252;
     use starknet::ContractAddress;
     use starknet::StorageAccess;
     use starknet::StorageBaseAddress;
--- a/contracts/src/emergency/sequencer_uptime_feed.cairo
+++ b/contracts/src/emergency/sequencer_uptime_feed.cairo
@@ -1,10 +1,10 @@
 impl Felt252TryIntoBool of TryInto::<felt252, bool> {
     fn try_into(self: felt252) -> Option<bool> {
         if self == 0 {
-            return Option::Some(false);
+            Option::Some(false);
         }
         if self == 1 {
-            return Option::Some(true);
+            Option::Some(true);
         }
         Option::None(())
     }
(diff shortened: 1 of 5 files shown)


contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -0,0 +1,358 @@
+impl Felt252TryIntoBool of TryInto::<felt252, bool> {
+    fn try_into(self: felt252) -> Option<bool> {
+        if self == 0 {
+            return Option::Some(false);
```
augustbleeds: apparently, only match zero is supported, hence why we use if/else

`error: Only match zero (match ... { 0 => ..., _ => ... }) is currently supported`

contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -0,0 +1,358 @@
+impl Felt252TryIntoBool of TryInto::<felt252, bool> {
```
augustbleeds: implementing TryInto and Into traits are required so we can read and write boolean value into storage (where it's stored as felt252) as a part of implementing the `StorageAccess` trait

contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -0,0 +1,358 @@
...
+
+    use chainlink::libraries::ownable::Ownable;
+    use chainlink::libraries::access_controller::AccessController;
+    use chainlink::libraries::simple_read_access_controller::SimpleReadAccessController;
+    use chainlink::libraries::simple_write_access_controller::SimpleWriteAccessController;
+    use chainlink::ocr2::aggregator::Round;
+    use chainlink::ocr2::aggregator::IAggregator;
+
+    #[derive(Copy, Drop, Serde, PartialEq)]
+    struct RoundTransmission {
```
augustbleeds: Note: I would've reused chainlink::ocr2::aggregator::Aggregator::Transmission but the answer type is a bool here

contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -0,0 +1,358 @@
...
+        updated_at: u64        
+    }
+
+    impl RoundTransmissionAccess of StorageAccess::<RoundTransmission> {
+        fn read(address_domain: u32, base: StorageBaseAddress) -> SyscallResult::<RoundTransmission> {
+            Result::Ok(
+                RoundTransmission {
+                    answer: storage_read_syscall(
+                        address_domain, storage_address_from_base_and_offset(base, 0_u8)
+                    )?.try_into().unwrap(),
```
augustbleeds: using impl defined above

contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -0,0 +1,358 @@
...
+            )
+        }
+
+        fn write(
+            address_domain: u32, base: StorageBaseAddress, value: RoundTransmission
+        ) -> SyscallResult::<()> {
+            storage_write_syscall(
+                address_domain,
+                storage_address_from_base_and_offset(base, 0_u8),
+                value.answer.into()
```
augustbleeds: using impl defined above

contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -0,0 +1,358 @@
...
+            )
+        }
+    }
+
+
+    struct Storage {
+        // l1 sender is an ethereum address
+        _l1_sender: u256,
+        // maps round id to round transmission
+        _round_transmissions: LegacyMap<u128, RoundTransmission>,
```
augustbleeds: the first key is the round_id

contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -0,0 +1,358 @@
...
+    fn L1SenderTransferred(from_address: u256, to_address: u256) {}
+
+    impl Aggregator of IAggregator {
+        fn latest_round_data() -> Round {
+            require_access();
+            let latest_round_id = _latest_round_id::read();
+            let round_transmission = _round_transmissions::read(latest_round_id);
+            // bool -> felt252 -> u128
+            let answer: u128 = round_transmission.answer.into().try_into().unwrap();
+            Round {
```
augustbleeds, line 57: Round is the external struct returned by the Proxy, so we have to do some type conversions for round_id and answer. This is because internally round_id uses u128 and answer uses bool whereas externally the AggregatorProxy defines round_id as felt252 and answer as u128.

contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -0,0 +1,358 @@
...
+        }
+
+        fn round_data(round_id: u128) -> Round {
+            require_access();
+            assert(round_id < _latest_round_id::read(), 'invalid round id');
+            let round_transmission = _round_transmissions::read(round_id);
+            // bool -> felt252 -> u128
+            let answer: u128 = round_transmission.answer.into().try_into().unwrap();
+            Round {
+                round_id: round_id.into(),
```
augustbleeds, line 71: same conversions occur here

contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -0,0 +1,358 @@
...
+                answer: answer,
+                block_num: round_transmission.block_num,
+                started_at: round_transmission.started_at,
+                updated_at: round_transmission.updated_at,   
+            }
+        }
+
+        fn description() -> felt252 {
+            'L2 Sequencer Uptime Status Feed'
```
augustbleeds, line 80: No need to create storage variables (expensive) for these values which won't be modified

contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    #[constructor]
+    fn constructor(initial_status: bool, owner_address: ContractAddress) {
+        initializer(initial_status, owner_address);
+    }
+
+    #[l1_handler]
+    fn update_status(from_address: felt252, status: bool, timestamp: u64) {
+        assert(_l1_sender::read() == from_address.into(), 'EXPECTED_FROM_BRIDGE_ONLY');
```
augustbleeds: from_address is received as felt252 so we have to convert `into` a u256. U256 is the only value felt252 can natively convert into without error (due to larger value type)

contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -0,0 +1,358 @@
...
+            let round_id = latest_round_id + 1_u128;
+            _record_round(round_id, status, timestamp);
+        }
+    }
+
+    #[external]
+    fn set_l1_sender(address: u256) {
+        Ownable::assert_only_owner();
+        // because u256 literals are not supported
+        let ETH_ADDRESS_BOUND = u256 { high: 0x100000000_u128, low: 0_u128 }; // 2 ** 160
```
augustbleeds: 160 - 128 = 32. So there should be 32 zero bits on the high side. Each hex digit is 4 bits. So 32/4 = 8 hex digits should be 0

contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -0,0 +1,358 @@
...
+        };
+        _round_transmissions::write(round_id, round);
+
+        let sender = starknet::info::get_caller_address();
+
+        NewRound(round_id, sender, timestamp);
+        AnswerUpdated(status, round_id, timestamp);
+    }
+
+    fn _update_round(round_id: u128, mut round: RoundTransmission) {
```
augustbleeds: update round only updates the `updated_at` (aka the transmission timestamp). Remember that the round doesn't increase unless the status of the feed changes -- hence why you don't need to change any other values.

contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -0,0 +1,358 @@
...
+
+     #[external]
+    fn enable_access_check() {
+        SimpleWriteAccessController::enable_access_check()
+    }
+
+    #[external]
+    fn disable_access_check() {
+        SimpleWriteAccessController::disable_access_check()
+    }
```
archseer: Oh good catch, these also need to be added to aggregator.cairo and aggregator_proxy.cairo. Can you do that in this PR?

contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -0,0 +1,358 @@
...
+            _update_round(latest_round_id, latest_round);
+        } else {
+            // only increment round when status flips
+            let round_id = latest_round_id + 1_u128;
+            _record_round(round_id, status, timestamp);
+        }
+    }
+
+    #[external]
+    fn set_l1_sender(address: u256) {
```
cfal: couldn't the L1 address be a felt252 if you only need 160 bits, similar to the old contract? the `l1_handler` also seems to propagate the `from_address` as a felt

contracts/src/emergency/sequencer_uptime_feed.cairo
cfal: this could be simplified to have implicit returns eg

```
if self == 0 {
  Option::Some(false)
} else if self == 1 {
  Option::Some(true)
} else {
  None
}
```

contracts/src/emergency/sequencer_uptime_feed.cairo
cfal: don't feel strongly about it, but i feel like it's reasonable here to reuse of Transmission/u128 as the answer to reduce the boilerplate

contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    #[l1_handler]
+    fn update_status(from_address: felt252, status: bool, timestamp: u64) {
+        assert(_l1_sender::read() == from_address.into(), 'EXPECTED_FROM_BRIDGE_ONLY');
+
+        let latest_round_id = _latest_round_id::read();
+        let latest_round = _round_transmissions::read(latest_round_id);
+
+        if (timestamp <= latest_round.started_at) {
```
cfal: brackets can be omitted - i wonder if `cairo-format` cleans this up automatically

contracts/src/emergency/sequencer_uptime_feed.cairo
archseer: Agreed, when we talked about it I forgot that this actually has to match the IAggregator interface. Since we're reading it into a bool, then immediately casting it to u128 to match the interface, we might as well just use the same types

archseer: Though the Solidity version does do this: https://github.com/smartcontractkit/chainlink/blob/d2ef7d4b2d51331cba981b6d0230a63059b26b39/contracts/src/v0.8/dev/ArbitrumSequencerUptimeFeed.sol#L159-L166

I think it's probably so it uses less space. Since in our case a bool and u128 both use a single slot it doesn't matter

contracts/src/emergency/sequencer_uptime_feed.cairo
archseer: probably does, I'd drop the parens

contracts/src/emergency/sequencer_uptime_feed.cairo
archseer: Yoo should be able to store the `_l1_sender` as felt252 and then compare the two as felts?

contracts/src/emergency/sequencer_uptime_feed.cairo
augustbleeds: Sounds good, I'll change to use the other Transmission struct

contracts/src/emergency/sequencer_uptime_feed.cairo
augustbleeds: felts are field elements so there's no natural comparison between the two ( i think it has to do with modular arithmetic). So we'll have to convert to u256 anyways to compare the two (I double checked with starknet discord)

I don't have a strong preference between representing l1_sender as  u256 or doing felt252. Felt252 leads to less code but u256 is more semantically correct. Probably want to optimize for less code though so I'll change it to felt252

cfal: does this work for the felt comparison? https://github.com/starkware-libs/cairo/blob/1198e1a9420cf4a82fe40448f4965ea43ad749bf/corelib/src/lib.cairo#L145

contracts/src/emergency/sequencer_uptime_feed.cairo
augustbleeds: yeah definitely!

contracts/src/emergency/sequencer_uptime_feed.cairo
augustbleeds: yup! == and != work. Whereas <, <=, >, and >= don't work

contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -0,0 +1,305 @@
+impl Felt252TryIntoBool of TryInto::<felt252, bool> {
+    fn try_into(self: felt252) -> Option<bool> {
+        if self == 0 {
+            Option::Some(false);
+        }
+        if self == 1 {
+            Option::Some(true);
+        }
+        Option::None(())
```
archseer:
```suggestion
        if self == 0 {
            Option::Some(false)
        } else if self == 1 {
            Option::Some(true)
        } else {
            Option::None(())
        
        }
```

The current implementation always returns `None`

contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -0,0 +1,305 @@
...
+    }
+}
+
+impl BoolIntoFelt252 of Into::<bool, felt252> {
+    fn into(self: bool) -> felt252 {
+        if self {
+            return 1;
+        }
+        return 0;
```
archseer:
```suggestion
        if self {
            return 1;
        }
        0
```

archseer: Do we actually need these traits now? I'd also consider opening a PR upstream

contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -0,0 +1,305 @@
...
+            _record_round(round_id, status, timestamp);
+        }
+    }
+
+    #[external]
+    fn set_l1_sender(address: felt252) {
+        Ownable::assert_only_owner();
+
+        // convert both to u256 (felts don't implement PartialOrd)
+        assert(address.into() < ETH_ADDRESS_BOUND.into(), 'invalid eth address');
```
archseer, line 125: Is this correct? I thought the ETH addresses are above a certain bound (have to be long enough)

contracts/src/emergency/sequencer_uptime_feed.cairo
augustbleeds: oops

contracts/src/emergency/sequencer_uptime_feed.cairo
augustbleeds: yessir you're correct , we don't need them anymore 🥳

contracts/src/emergency/sequencer_uptime_feed.cairo
augustbleeds: Eth addresses are 160 bytes, so if I'm not mistaken we just need to check if the address type is lower than the highest possible value. 

in the cairo token_bridge examples [they do something simila](https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-starknet/test_data/token_bridge.cairo#L27)r (comparing two u256s)  so I think we should be golden 🏅

PR #229, opened 2023-04-13
add cairo make commands, run scarb tests in CI

cfal:


.github/workflows/lint.yml
```diff
@@ -7,20 +7,28 @@ on:
       - main
   pull_request:
 
+env:
+  CAIRO_VERSION: v1.0.0-alpha.7
```
archseer: Let's use `alpha.6` for now since 7 has some breaking changes (it strips turbofish)

cfal:
cairo-format is still unable to run correctly here, because it conflicts with python cairo-lang's cairo-format
even though cairo-lang was removed from requirements.txt (which maybe it shouldn't be?), it's still a requirement for starknet-devnet

it might still be cleaner to do the download/install step in the Makefile - we can put the cairo/scarb binaries in a bin subdir here, and it would guarantee that we build/test/format the same across our local environments

edit: or, removing `python39Packages.venvShellHook` could be enough as a CI fix

archseer:
You could refer to $GITHUB_PATH directly in the CI workflow? That would give it priority over the venv. 


> it might still be cleaner to do the download/install step in the Makefile

Since it's a binary we could do what we did with terrad and just wrap it in a nix derivation.



> edit: or, removing `python39Packages.venvShellHook` could be enough as a CI fix

Hmm that will probably break some of the tests down the road because they manually spawn `starknet-devnet` (search for `network.ts`)

cfal:
> You could refer to $GITHUB_PATH directly in the CI workflow? That would give it priority over the venv.

GITHUB_PATH is already prepended to PATH according to docs - `nix develop make format-check` causes the venvShellHook to execute and to prepend the python wrappers in front of what's in GITHUB_PATH

that could be the fix actually - just do `make format-check` without `nix develop` prefixed (or would make not be available then?)

> 
> > it might still be cleaner to do the download/install step in the Makefile
> 
> Since it's a binary we could do what we did with terrad and just wrap it in a nix derivation.

venvShellHook will still end up with python having precedence, i think

contracts/src/multisig.cairo
```diff
@@ -116,7 +116,7 @@ mod Multisig {
             address_domain: u32, base: StorageBaseAddress, value: Transaction
         ) -> SyscallResult::<()> {
             storage_write_syscall(
-                address_domain, storage_address_from_base_and_offset(base, 0_u8), value.to.into(),
+                address_domain, storage_address_from_base_and_offset(base, 0_u8), value.to.into(),
```
archseer, line 119: Not sure why but the formatter adds a trailing space here 🤷🏻

contracts/src/ocr2/aggregator.cairo
```diff
@@ -63,13 +61,10 @@ fn split_felt(felt: felt252) -> (u128, u128) {
...
-
 // TODO: wrap hash_span
-impl SpanLegacyHash<
-    T,
-    impl THash: LegacyHash::<T>,
-    impl TCopy: Copy::<T>
-> of LegacyHash::<Span<T>> {
+impl SpanLegacyHash<T,
+impl THash: LegacyHash::<T>,
+impl TCopy: Copy::<T>> of LegacyHash::<Span<T>> {
```
archseer, line 67: This is a mess too 🤷🏻

PR #230, opened 2023-04-18
txm: minimal feature set

aalu1418:
- [x] Nonce management: https://smartcontract-it.atlassian.net/browse/BCI-1112
- [x] Removing Tx batching: https://smartcontract-it.atlassian.net/browse/BCI-1113
- [x] Confirmation logic + tx tracking: https://smartcontract-it.atlassian.net/browse/BCI-1114
- [x] fixing test cases


aalu1418:
note: there are config interface changes that break the build process with core

relayer/pkg/chainlink/keys/nonce.go
```diff
@@ -0,0 +1,126 @@
...
+	return nm.n[addr.String()], nil
+}
+
+func (nm *nonceManager) IncrementNextSequence(addr caigotypes.Hash, chainId string, currentNonce *big.Int) error {
+	if err := nm.validate(addr, chainId); err != nil {
+		return err
+	}
+
+	nm.lock.Lock()
+	defer nm.lock.Unlock()
```
cfal, line 96: we briefly chatted about this - are we expecting this function to be called in parallel, and is there an issue with a higher nonce being received and propagated by some nodes before it receives the lower nonce?

relayer/pkg/chainlink/keys/nonce.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (nm *nonceManager) validate(addr caigotypes.Hash, id string) error {
+	if id != nm.chainId {
+		return fmt.Errorf("chain id does not match: %s (expected) != %s (got)", nm.chainId, id)
+	}
+
+	nm.lock.RLock()
+	defer nm.lock.RUnlock()
+	if _, exists := nm.n[addr.String()]; !exists {
```
cfal, line 108: do we need an existence check like this in `NextSequence` L97?

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -46,21 +50,29 @@ type starktxm struct {
...
+	txm := &starktxm{
+		lggr:    logger.Named(lggr, "StarknetTxm"),
+		queue:   make(chan Tx, MaxQueueLen),
+		stop:    make(chan struct{}),
+		client:  utils.NewLazyLoad(getClient),
+		ks:      keystore,
+		cfg:     cfg,
+		txStore: ChainTxStore{},
+	}
+	client, err := txm.client.Get()
```
cfal: please ignore if this is incorrect, not super familiar - but is there a point to the LazyLoad if we're calling `client.Get()` immediately?

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -220,12 +264,35 @@ func (txm *starktxm) HealthReport() map[string]error {
...
 	default:
-		return errors.Errorf("failed to enqueue transaction: %+v", tx)
+		return fmt.Errorf("failed to enqueue transaction: %+v", tx)
 	}
 
 	return nil
 }
+
+func (txm *starktxm) InflightCount() (queue int, unconfirmed int) {
+	list := maps.Values(txm.txStore.GetUnconfirmed())
```
cfal: imo it seems a little inconsistent that `GetUnconfirmed()` on `ChainTxStore` counts across all keys, and does the iteration that is done in this function, while `ChainTxStore.InflightCount` is for a specific hash. and also that `InflightCount` doesn't make use of the other `InflightCount` functions to count a total :)

 i wonder if it would be more readable to have `GetAllUnconfirmed()`, `GetAllInflightCount()` on ChainTxStore when counting across all senders, and `GetUnconfirmed(hash)`, and `GetInflightCount(hash)` (if necessary) for per-sender.

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -69,135 +81,167 @@ func (txm *starktxm) Name() string {
...
+				for i := range hashes[addr] {
+					hash := hashes[addr][i]
+					status, err := client.Gw.TransactionStatus(ctx, gateway.TransactionStatusOptions{
+						TransactionHash: hashes[addr][i],
+					})
+					if err != nil {
+						txm.lggr.Errorw("failed to fetch transaction status", "hash", hash, "error", err)
+					}
+
+					if status.TxStatus == "ACCEPTED_ON_L1" || status.TxStatus == "ACCEPTED_ON_L2" || status.TxStatus == "REJECTED" {
```
cfal, line 233: i'm probably missing something, but some questions:
- i don't see any code for handling rejected transactions (eg. requeueing) - does that need to be handled?
- what's the point of txStore? i don't see any functional changes in the broadcast loop depending on its state. is it just so that we can calculated `InflightCount`?

relayer/pkg/chainlink/keys/nonce.go
aalu1418: i think it's unlikely that these functions get called in parallel - but the locks are mostly to prevent broadcasting the same nonce

> is there an issue with a higher nonce being received and propagated by some nodes before it receives the lower nonce?

good question...the honest answer is that I'm not sure yet. I think the mempool can handle a higher nonce, but I'm not 100% on that

relayer/pkg/chainlink/keys/nonce.go
aalu1418: so we use `validate` before we acquire the lock, are you suggesting that we should check the lock again once the write lock is acquired?

relayer/pkg/chainlink/txm/txm.go
aalu1418: good question!
`Get()` fetches a client based on the chains/nodes implementation. so if there's an error you can clear the client and retrieve another once from the list (assuming there are more than one implemented)

this prefetching is mostly just an optimization so the nonce manager can use it easily

relayer/pkg/chainlink/txm/txm.go
aalu1418: > i don't see any code for handling rejected transactions (eg. requeueing) - does that need to be handled?

do you mean rejected by the RPC during initial broadcast or rejected by the confirmation process?
either way, we currently do not requeue - which forces the OCR2 process to trigger rebroadcasting via another CL node. also if the feed is running with a fast heartbeat a missed report can be superseded by a more recent report (like on Solana)

but I do agree that handling rejected Txs is eventually needed - this is mostly a minimal set so we are able to broadcast txs with local nonce management and understand why they may fail with the confirmation logic

> what's the point of txStore? i don't see any functional changes in the broadcast loop depending on its state. is it just so that we can calculated InflightCount?

txStore is mostly for the confirmation process so we know which txs need to be checked

relayer/pkg/chainlink/txm/txm.go
aalu1418: nice! i like that idea - cleaner and clearer naming 🎉

aalu1418: adjusted the naming, let me know what you think :)

relayer/pkg/chainlink/keys/test_helpers.go
```diff
@@ -0,0 +1,18 @@
...
+	// devnet key derivation
+	// https://github.com/Shard-Labs/starknet-devnet/blob/master/starknet_devnet/account.py
+	DevnetClassHash, _ = new(big.Int).SetString("1803505466663265559571280894381905521939782500874858933595227108099796801620", 10)
+	DevnetSalt         = big.NewInt(20)
+)
+
+func (key Key) DevnetAccountAddrStr() string {
+	return "0x" + hex.EncodeToString(PubKeyToAccount(key.pub, DevnetClassHash, DevnetSalt))
+
+}
```
archseer: If you're pre-fetching the address list is this still necessary?

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -220,12 +264,35 @@ func (txm *starktxm) HealthReport() map[string]error {
...
 	}
 
 	return nil
 }
+
+func (txm *starktxm) InflightCount() (queue int, unconfirmed int) {
+	list := maps.Values(txm.txStore.GetAllInflightCount())
+	for i := range list {
+		unconfirmed += list[i]
+	}
```
archseer:
```suggestion
	for _, count := range list {
		unconfirmed += count
	}
```
?

relayer/pkg/chainlink/keys/nonce.go
cfal: ah, i missed that this was the function actually called by NextSequence :)

relayer/pkg/chainlink/txm/txm.go
cfal: still slightly confused by this - if an error happens during this `txm.client.Get()`, `txm, err` is returned, but it doesn't seem like a harmless error.. because in the chain code, it'll still fail to initialize: https://github.com/smartcontractkit/chainlink/blob/a80d630ef6c3470871bdc908ac25aed7644dd663/core/chains/starknet/chain.go#L49-L52

so it seems like the only client that ever gets used in `txm.client.Get()` calls below is the one initialized on the `getClient` invocation in this constructor (since LazyLoad will just keep returning the same client value)

wouldn't this be equivalent to just doing `getClient()` directly once here without the lazy loading? or should we not return `txm, err` here when the client doesn't initialize? in which case, since the same client is repeatedly returned on `txm.client.Get()`, we could just iterate through the getClient list for a working client in the constructor first, and avoid locking/using LazyLoad altogether as well.. if that makes sense :)

relayer/pkg/chainlink/txm/txm.go
cfal: > do you mean rejected by the RPC during initial broadcast or rejected by the confirmation process?

ah, i meant rejected when checking the transaction status in this loop

> but I do agree that handling rejected Txs is eventually needed - this is mostly a minimal set so we are able to broadcast txs with local nonce management and understand why they may fail with the confirmation logic

got it - so i guess we'll be adding more logic to the `confirmLoop` function here? my comment was just that it seems like this `confirmLoop` isn't necessary, since we don't do anything besides logging regardless of if the transaction status was rejected or not

relayer/pkg/chainlink/txm/txm.go
cfal: lgtm, thanks for updating!

cfal (commented):
overall lgtm (although i'm still unsure if the LazyLoad is necessary, but nbd)

will leave it to @archseer for the final approval

relayer/pkg/chainlink/txm/txm.go
aalu1418: yeah in the current state - it's not a part of critical functionality, but it's really useful to see in the CL node why TXs are failing (from previous experience with Solana)

in the future, it should be replaced by the txm platform which will have retry logic built in with the confirmer

relayer/pkg/chainlink/txm/txm.go
aalu1418: i think that makes sense! (just double checking my understanding - you're suggesting we could just go through all the clients to find a valid one at the beginning, and then we don't need lazy load)

based on that, I guess my question would be what happens if the endpoint goes down and the client starts to fail? lazy load would allow for it to switch to a different valid client

cfal: i might have missed this, sorry - does lazy load switch it to a different client? it seemed like client.Get would still keep returning the same client once getClient succeeds here the first time, even if a function call on the client fails

cfal: took another look - i think we'd need to have `txm.client.Reset()` calls to allow switching for another client?

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -69,135 +81,167 @@ func (txm *starktxm) Name() string {
...
+
+	nonce, err := txm.nonce.NextSequence(accountAddress, client.Gw.ChainId)
 	if err != nil {
-		return txhash, errors.Errorf("failed to create new account: %s", err)
+		return txhash, fmt.Errorf("failed to get nonce: %+w", err)
 	}
 
 	// get fee for txm
+	// optional - pass nonce to fee estimate (if nonce gets ahead, estimate may fail)
```
archseer, line 159: Actually, if you don't pass in a nonce, it'll fetch one from the chain for you anyway which is what you were trying to avoid: https://github.com/smartcontractkit/caigo/blob/41120ca1f9f3689da24469fe19942707abcf1425/account.go#L261-L266

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -69,135 +81,167 @@ func (txm *starktxm) Name() string {
...
 	if err != nil {
-		return txhash, errors.Errorf("failed to estimate fee: %s", err)
+		return txhash, fmt.Errorf("failed to estimate fee: %+w", err)
 	}
 
 	fee, _ := big.NewInt(0).SetString(string(feeEstimate.OverallFee), 0)
 	expandedFee := big.NewInt(0).Mul(fee, big.NewInt(int64(FEE_MARGIN)))
 	max := big.NewInt(0).Div(expandedFee, big.NewInt(100))
 	details := caigotypes.ExecuteDetails{
 		MaxFee: max,
```
archseer, line 170: Is there a point to doing this fee estimation + calculation rather than relying on the builtin in caigo since it's essentially the same code/margin?

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -69,135 +81,167 @@ func (txm *starktxm) Name() string {
...
+		return txhash, fmt.Errorf("broadcast: failed to fetch client: %+w", err)
+	}
 	// create new account
-	account, err := caigo.NewGatewayAccount(privKey, accountAddress.String(), txm.client.Gw, caigo.AccountVersion1)
+	account, err := caigo.NewGatewayAccount(privKey, accountAddress.String(), client.Gw, caigo.AccountVersion1)
+	if err != nil {
+		return txhash, fmt.Errorf("failed to create new account: %+w", err)
+	}
+
+	nonce, err := txm.nonce.NextSequence(accountAddress, client.Gw.ChainId)
```
archseer, line 153: Is fetching the nonce that expensive that we need local nonce management? It just does a read against the local RPC node. It's faster to avoid it, but I'm thinking about cases like someone using the same account (e.g. to transfer funds) while the node is running, it would mess up the txm until a reboot.

Would it make sense to add a recovery mechanism in the nonce manager where it would re-fetch the latest nonce from the account (same logic as Register) if it falls behind? Or is that an edge case we don't see happening since the account will only be used by the txm?

archseer: e.g.

> // TODO: investigate if nonce management is needed (nonce is requested queried by the sdk for now)

I haven't actually seen a doc or anything that led to the decision that we do need nonce management

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -69,135 +81,167 @@ func (txm *starktxm) Name() string {
...
 
 	// handle nil pointer
 	if res == nil {
 		return txhash, errors.New("execute response and error are nil")
 	}
 
-	return res.TransactionHash, nil
+	// update nonce if transaction is successful
+	err = errors.Join(
+		txm.nonce.IncrementNextSequence(accountAddress, client.Gw.ChainId, nonce),
```
archseer, line 190: Since this is technically fallible, why not do the increment optimistically? e.g. `txm.nonce.NextSequence()` should return the next nonce to use, and increase the counter by 1. That way it's an atomic operation -- if transactions fail we just end up using the next nonce in the sequence after that

relayer/pkg/chainlink/txm/txm.go
aalu1418: ah right - that must have gotten lost in the reshuffling. I will add that

Update: added!

relayer/pkg/chainlink/txm/txm.go
aalu1418: that's true! but that's not a reliable source for the nonce - especially if you need to send multiple transactions soon after each other since the chain may not have included the first tx, you would end broadcasting two txs that have the same nonce.

for estimation this is fine since we are just looking for fee values, but for actual broadcasting we'd run into transaction failures.

relayer/pkg/chainlink/txm/txm.go
aalu1418: not sure? if i remember correctly, you added this piece a while ago so I've just kept it in for now, but we can definitely get rid of it if you think that is better

I think the original purpose was to provide additional buffer on top of the estimation

relayer/pkg/chainlink/txm/txm.go
aalu1418: see comment here on why local nonce management is needed: https://github.com/smartcontractkit/chainlink-starknet/pull/230#discussion_r1196561541

the EVM txm also uses this model for the same reason (summary: always requesting nonce from the chain only works if you only have 1 inflight tx at a time)

also the account used by the CL node should only be used for the CL node - the EVM txm also has that constraint, and would require a re-sync if the nonce changed outside of interactions from the CL node

I think there could be edge cases with the nonce manager, but it's honestly hard to say without seeing this run on a live network. testing locally only goes so far when the RPC behaves 🤷🏼

relayer/pkg/chainlink/txm/txm.go
aalu1418: afaik it's more robust to split it into a get + increment functionality

for example: if the tx encounters an error (estimation failure, signing issues, RPC rejecting, client errors), you wouldn't want to increment the tx until you are sure that the tx was accepted by the RPC. otherwise it would be very easy to fall out of sync

archseer: do the transactions need to be sequential? If not it's perfectly time if the id doesn't end up being used because of an error etc

relayer/pkg/chainlink/txm/txm.go
cfal: i might be nit picking at this now, feel free to ignore if you are happy with this - but the client error is still returned on L72, which causes the chain not to be initialized at https://github.com/smartcontractkit/chainlink/blob/a80d630ef6c3470871bdc908ac25aed7644dd663/core/chains/starknet/chain.go#L49-L52

since in the rest of the file, we're committing to not having fatal client endpoint errors, should we just not return `err`, or skip the `client.Get()` in the constructor? or is the idea to fix the code in chain.go so it doesn't return `nil, err` on error?

i feel like it would still be useful to know if _all_ clients aren't working.. that seems like it could be a valid case for a fatal error in New, but maybe that's too much (or perhaps it shouldn't even be fatal)

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -69,135 +82,168 @@ func (txm *starktxm) Name() string {
...
+			}
+
+			hashes := txm.txStore.GetAllUnconfirmed()
+			for addr := range hashes {
+				for i := range hashes[addr] {
+					hash := hashes[addr][i]
+					status, err := client.Gw.TransactionStatus(ctx, gateway.TransactionStatusOptions{
+						TransactionHash: hashes[addr][i],
+					})
+					if err != nil {
```
cfal, line 223: should we do a `txm.client.Reset()`, when the API call fails?

relayer/pkg/chainlink/txm/txm.go
aalu1418: fair point! let me see if can change anything where we might not need to fetch the client right at the beginning 🤔

relayer/pkg/chainlink/txm/txm.go
aalu1418: i thought about that - it's a little weird since sometimes returned errors aren't just relevant to the the API endpoint failing. so there would probably need to be more logic about what sort of errors are seen that should reset the client - right now I don't have enough information to set that up

but I would agree that API failures should trigger a client reset - if we know which errors to reset on

relayer/pkg/chainlink/txm/txm.go
aalu1418: oh interesting - my assumption was that nonces always have to be processed sequentially by the chain otherwise the chain will wait until it sees the expected next nonce and hold all other txs in the mempool. EVM chains operate like this, and it would surprise me if starknet used nonces differently?

aalu1418: based on this it sounds like they need to be sequential: https://docs.starknet.io/documentation/architecture_and_concepts/State/starknet-state/

> Similarly to Ethereum, the contract’s nonce is used to provide replay protection at the protocol level. When a transaction is sent to a given contract, the transaction’s nonce must match the nonce of the contract. After the transaction is executed, the contract’s nonce is increased by one.

.github/workflows/integration-tests-smoke.yml
```diff
@@ -53,7 +53,7 @@ jobs:
         uses: smartcontractkit/chainlink-github-actions/chainlink-testing-framework/build-image@3ec816fe2e6dd3f70511e8115d86e296818c58f8 # cosmos_one_off
         with:
           cl_repo: smartcontractkit/chainlink
-          cl_ref: 6b935c6ff7ec12b16b3f644b64080dec29b5a2c4
+          cl_ref: d0bb5e7811b9b590037452f48e4d381998349e54
```
aalu1418, line 56: this should be switched back to using the input rather than a hard coded input `${{ github.event.inputs.cl_branch_ref }}`

once the corresponding PR in core is merged: https://github.com/smartcontractkit/chainlink/pull/9273

aalu1418 (commented):
not sure why e2e tests are failing with test image build failures - CL image builds successfully

.github/workflows/integration-tests-smoke.yml
cfal: i have this as part of https://github.com/smartcontractkit/chainlink-starknet/pull/267, feel free to merge it separately though if blocking

cfal:
> not sure why e2e tests are failing with test image build failures - CL image builds successfully

yeah a few fixes are needed - work happening on https://github.com/smartcontractkit/chainlink-starknet/pull/267

cfal:
hmm guess https://github.com/smartcontractkit/chainlink-starknet/actions/runs/5041718431/jobs/9041661280 is flaky, it's failing at ToT but seems to have passed here

archseer:
Yeah I reported it to Aaron, it seems to be a race in the test

PR #233, opened 2023-04-19
Bump nix install so it will handle nix 14

tateexon:
Otherwise we will run into this: https://github.com/cachix/cachix-action/issues/144


aalu1418 (approved):

tateexon:
The fix for the image build for the integration tests is coming with Damjan's changes. Merging just in case the nix issue ends up blocking his changes.

PR #234, opened 2023-04-20
BCI-1188: Rewrite Link Token Cairo 1.0

augustbleeds:
Questions:
- [x]  ~why did we require the receiver contract (in transfer_and_call) to support erc-165 in starknet when [solidity contracts don't]~(https://github.com/smartcontractkit/chainlink/blob/bb2564be5e0e295699bb1f23b0b2ae3e0bfa2cd8/contracts/src/v0.4/ERC677Token.sol)? Just to be safer?
- [x]  ~the function selector i got (using solidity) was different than [the one in cairo 0.x ](https://github.com/smartcontractkit/chainlink-starknet/blob/develop/contracts/src/chainlink/cairo/token/ERC677/library.cairo#L11)-- are we supposed to use the solidity function signature?~

Please see in-line PR comments.

Notes:
* The ERC-165 version of the supportsInterface function (based off of OZ contracts in cairo 1.0) takes in u32 (as opposed to felts) and the IDs for common ERC contracts are indeed based off of the solidity function selectors


contracts/src/libraries/token/erc20.cairo
```diff
@@ -0,0 +1,208 @@
+// taken from OpenZeppelin
```
krebernisak: Can we replace this with an import?

augustbleeds: Eventually it will be. But OZ hasn't merged the code to master yet

augustbleeds: So it's not available as a scarb package yet

contracts/src/libraries/token/erc677.cairo
```diff
@@ -0,0 +1,40 @@
...
+    const IERC677_RECEIVER_ID: felt252 = 0xa53f2491;
+
+    #[event]
+    fn Transfer(from: ContractAddress, to: ContractAddress, value: u256, data: Array<felt252>) {}
+
+    fn transfer_and_call(to: ContractAddress, value: u256, data: Array<felt252>) -> bool {
+        let sender = starknet::info::get_caller_address();
+
+        ERC20::transfer(to, value);
+        Transfer(sender, to, value, data.clone());
```
augustbleeds, line 31: had to use clone here because events do not support references. I thought about using @ (snapshot) but I'd have to implement Serde on Array<felt252> type which involves desnapping, but desnapping is only valid on Copy-able but Array isn't copyable

contracts/src/libraries/token/erc677.cairo
```diff
@@ -0,0 +1,40 @@
...
+    use chainlink::libraries::token::erc20::ERC20;
+    use array::ArrayTrait;
+    use array::SpanTrait;
+    use clone::Clone;
+    use array::ArrayTCloneImpl;
+
+    use super::IERC677ReceiverDispatcher;
+    use super::IERC677ReceiverDispatcherTrait;
+
+    const IERC677_RECEIVER_ID: felt252 = 0xa53f2491;
```
augustbleeds: this is what I got in remix by doing the XOR of the solidity function selectors of onTokenTransfer(address,uint256,bytes) and supportsInterface(bytes4)

contracts/src/lib.cairo
```diff
@@ -3,7 +3,7 @@
 mod ocr2;
 mod libraries;
 mod emergency;
-mod multisig;
```
augustbleeds, line 6: temporarily comment this out because there are compile errors

contracts/src/tests.cairo
```diff
@@ -1,5 +1,6 @@
 // Cairo files in folders can only be imported from a file of the same name as the folder
 
 mod test_aggregator;
-mod test_multisig;
```
augustbleeds, line 4: temporarily copy out due to compile errors

contracts/src/tests/test_erc677.cairo
```diff
@@ -0,0 +1,70 @@
+use chainlink::libraries::token::erc677::ERC677;
+use starknet::ContractAddress;
+use starknet::contract_address_const;
+use starknet::testing::set_caller_address;
+use array::ArrayTrait;
+use traits::Into;
+use zeroable::Zeroable;
+
+
+// todos are dependent on cairo-test support for contract calls
```
augustbleeds: waiting on cairo-test support

contracts/src/lib.cairo
archseer: Can you leave it in but commented out? It seems to be replaced here

contracts/src/libraries/token/erc677.cairo
```diff
@@ -0,0 +1,41 @@
...
+
+        ERC20::transfer(to, value);
+        Transfer(sender, to, value, data.clone());
+
+        let receiver = IERC677ReceiverDispatcher { contract_address: to };
+
+        let supports = receiver.supports_interface(IERC677_RECEIVER_ID);
+        // if supports {
+        //     receiver.on_token_transfer(sender, value, data);
+        // }
```
archseer: Can you re-enable this, finish the tests then mark them as `#[ignore]` for now (and leave a comment that it's blocked on nested call support in cairo-test)

archseer (commented):
Since the tests obsolete some of the old typescript tests, let's delete those in this PR too (under `contracts/tests/token`)

contracts/src/libraries/token/erc677.cairo
archseer: I think a clone here is fine! You need a copy of the data for both calls

contracts/src/libraries/token/erc677.cairo
archseer: (https://cairo-book.github.io/ch08-01-how-to-write-tests.html#ignoring-some-tests-unless-specifically-requested)

contracts/src/lib.cairo
archseer: Compilation should be fixed now, just need to rebase

contracts/src/libraries/token/erc677.cairo
augustbleeds: Sounds good

augustbleeds: Btw I just found out that they now support contract calls within cairo-test in this PR https://github.com/starkware-libs/cairo/pull/2912/files, but it's part of the new alpha release so we can't include the tests for it unless we update to 0v1.0.0-alpha.7 (which isn't supported on mainnet or starknet-devnet) yet.

archseer: alpha.7 was released 2 weeks ago, this will be in alpha.8

contracts/src/lib.cairo
archseer: Do the tests still fail for you? They pass on the cairo-1.0 branch with the new CI check

PR #236, opened 2023-04-24
keys: remove hardcoded contract hash, don't use AccountAddressStr

cfal:


archseer (dismissed):
LGTM but would also like a review from @aalu1418

aalu1418 (commented):
so the removal looks fine - are we planning to fix tests in this PR or a separte one?

I think this has changes that need to propagate to the job spec which impacts the `contract_transmitter`, `txm`, and `e2e tests`

aalu1418 (dismissed):
overall, looks really good to me! great job 🎉 

although I am a little confused why the e2e tests didn't fail... 🤔 
since the e2e tests need the account address passed in via the job spec: https://github.com/smartcontractkit/chainlink-starknet/blob/aaa45c923e495573eb517f4e2b4a92dbe6275ed8/integration-tests/common/common.go#L150
but it's not there? maybe something to validate that there is an address would be useful in the relayer...

aalu1418:
also can you update the the job spec examples here: https://github.com/smartcontractkit/chainlink-starknet/tree/develop/examples/spec

cfal:
> overall, looks really good to me! great job 🎉
> 
> although I am a little confused why the e2e tests didn't fail... 🤔 since the e2e tests need the account address passed in via the job spec:
> 
> https://github.com/smartcontractkit/chainlink-starknet/blob/aaa45c923e495573eb517f4e2b4a92dbe6275ed8/integration-tests/common/common.go#L150
> 
> 
> but it's not there? maybe something to validate that there is an address would be useful in the relayer...

added - along with the extra set of quotes since the other fields have it - do you know why it's necessary?

> also can you update the the job spec examples here: https://github.com/smartcontractkit/chainlink-starknet/tree/develop/examples/spec

done

examples/spec/ocr2-oracle-simple.spec.toml
```diff
@@ -8,6 +8,7 @@ p2pBootstrapPeers    = ["somep2pkey@localhost-tcp:port"]      # optional, overri
 p2pPeerID            = "<insert p2p id>"                      # optional, overrides P2P_PEER_ID
 ocrKeyBundleID       = "<insert starknet ocr2 key bundle id>" # optional, overrides OCR2_KEY_BUNDLE_ID
 transmitterID        = "<insert starknet transmitter key id>"
+accountAddress       = "<insert account contract address>"
```
aalu1418: this should be in the `[relayConfig]` section :)

integration-tests/common/common.go
```diff
@@ -148,8 +148,9 @@ func (c *Common) CreateJobsForContract(cc *ChainlinkClient, observationSource st
 
 	// Defining relay config
 	relayConfig := job.JSONConfig{
-		"nodeName": fmt.Sprintf("\"starknet-OCRv2-%s-%s\"", "node", uuid.NewV4().String()),
-		"chainID":  fmt.Sprintf("\"%s\"", c.ChainId),
+		"nodeName":       fmt.Sprintf("\"starknet-OCRv2-%s-%s\"", "node", uuid.NewV4().String()),
+		"accountAddress": fmt.Sprintf("\"%s\"", c.Account),
```
aalu1418, line 152: so i'm not confident `c.Account` is what you are looking for? (i'm not sure what that references)

having it in this struct means that it would be the same for each node - but you would want it to be different for each one...I think it would probably be somewhere in the for-loop that is further down in the file?

aalu1418 (changes requested):

integration-tests/common/common.go
cfal: wondering if i shouldn't have removed this field now:

https://github.com/smartcontractkit/chainlink/commit/edb338fd2536e4ea21c8592cb3a38b405718f3e7#diff-3c0b7bfb84ed7fb48bba1eb635465a3d5d6801d02cfcfe247c1873d70b87c3fdL379

it makes sense that RelayConfig needs to be created per for loop iteration now - should the above change be reverted so we can refer to `cc.NKeys[nIdx].TXKey.Data.Attributes.AccountAddr`?

examples/spec/ocr2-oracle-simple.spec.toml
cfal: oops, thanks

integration-tests/common/common.go
aalu1418: hmmm I think that change is still ok? you just need a different way to save the account address value once it's been deployed in the gauntlet section

cfal:
@archseer if this is blocking your work - feel free to merge it in first, and i'll follow up with @aalu1418 on the last outstanding comment

PR #240, opened 2023-04-27
cairo 1.0 gauntlet deploy

cfal:


contracts/Scarb.toml
```diff
@@ -15,4 +15,8 @@ test = "cairo-test --starknet ."
 # quaireaux = { git = "https://github.com/keep-starknet-strange/quaireaux.git" }
 
 [[target.starknet-contract]]
-
+sierra = true
```
augustbleeds, line 18: I'm guessing you're using scarb 0.2 (so not the version here https://github.com/smartcontractkit/chainlink-starknet/pull/224) since these options aren't available until scarb 0.2, which is not compatible with Starknet mainnet/testnet. 

I was talking to @archseer about this, but are we all operating under the assumption that Starknet will update their version to be compatible with cairo alpha-7 by launch date? How confident are we about this? Cause if they don't, we're compiling code and writing tests for code that won't be supported on mainnet.

contracts/hardhat.config.ts
```diff
@@ -46,25 +46,29 @@ const config: HardhatUserConfig = {
...
   networks: {
     devnet: {
       url: 'http://127.0.0.1:5050',
+      args: ['--cairo-compiler-manifest', '../vendor/cairo/Cargo.toml'],
     },
     integratedDevnet: {
       url: 'http://127.0.0.1:5050',
       venv: 'active',
-      args: ['--lite-mode'],
+      args: ['--lite-mode', '--cairo-compiler-manifest', '../vendor/cairo/Cargo.toml'],
```
augustbleeds, line 55: We should pull in the latest https://github.com/0xSpaceShard/starknet-hardhat-plugin/releases/tag/v0.8.0-alpha.1 hardhat plugin because it defaults to cairo-lang 0.11.0.2. They removed manifestPath option and replaced it with cairo1BinDir

cfal: - we need for their alpha.2 release which comes with things we want
- manifestPath is for `starknet` (for interacting with the starknet network), and not for this devnet set of arguments
- we don't actually use hardhat's compilation functionality (as we use scarb), so it doesn't apply to us

augustbleeds: Thanks for clarifying, I thought this was for starknet-hardhat plugin but it's for starknet-devnet

augustbleeds:
Can you rebase on the lastest cairo-1.0?

contracts/test/ocr2/aggregator.test.ts
```diff
@@ -17,44 +17,33 @@ const CHUNK_SIZE = 31
 // Observers - max 31 oracles or 31 bytes
 const OBSERVERS_MAX = 31
 const OBSERVERS_HEX = '0x00010203000000000000000000000000000000000000000000000000000000'
-const INT128_MIN = BigInt(-2) ** BigInt(128 - 1)
-const INT128_MAX = BigInt(2) ** BigInt(128 - 1) - BigInt(1)
+const UINT128_MAX = BigInt(2) ** BigInt(128) - BigInt(1)
 
-function encodeBytes(data: Uint8Array): string[] {
-  let felts = []
+function packUint8Array(data: Uint8Array | Buffer): bigint {
```
krebernisak: These util functions seem duplicated from here: [packages-ts/starknet-gauntlet-ocr2/src/lib/encoding.ts](https://github.com/smartcontractkit/chainlink-starknet/blob/0fadd26e2e2a993197a6e1cbaf515399ea85be29/packages-ts/starknet-gauntlet-ocr2/src/lib/encoding.ts)

Consider finding a shared place for those - probably `packages-ts/starknet` would make sense.

packages-ts/starknet-gauntlet-ocr2/src/lib/encoding.ts
```diff
@@ -20,46 +42,64 @@ export function bytesToFelts(data: Uint8Array): string[] {
...
+    data.push(...unpackedValues)
+    remainingLength -= chunkSize
   }
 
   return Buffer.from(data)
 }
 
 export const decodeOffchainConfigFromEventData = (data: string[]): encoding.OffchainConfig => {
-  const oraclesLen = Number(num.toBigInt(data[3]))
-  /** SetConfig event data has the following info:
```
krebernisak, line 83: These docs still seem useful to peek into the structure. Consider adding similar docs to the new version.

contracts/test/setup.ts
```diff
@@ -0,0 +1,91 @@
+import * as fs from 'fs'
```
krebernisak, line 1: What's the benefit of keeping hardhat around?
At this point, we use it just for integration tests, right?

augustbleeds: ^ Some functionality is hard or impossible to reproduce in cairo test. For example, the replace class syscall isn't supported by cairo test. Also, the test harness setup logic for aggregator is not easily achievable in cairo-test because it involves a lot of cryptography.

contracts/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "compile:cairo": "hardhat starknet-compile",
+    "compile:cairo": "scarb --profile release build",
```
augustbleeds: Is it still possible to compile with hardhat starknet-compile if you supply the cairo1 bin dir? Not that I think we should (since scarb is much more simpler) but just curious

archseer: I'd completely remove `compile:cairo` and invoke `scarb` where needed. No need to forward through yarn

contracts/package.json
```diff
@@ -4,7 +4,7 @@
...
     "compile:solidity": "hardhat compile",
     "compile": "yarn compile:cairo && yarn compile:solidity",
```
archseer: So this should then be removed too, instead `scarb ... && yarn compile:solidity` can be called in the makefile

contracts/test/ocr2/aggregator.test.ts
cfal: moved to packages-ts/starknet

packages-ts/starknet-gauntlet-ocr2/src/lib/encoding.ts
cfal: possibly just me, but i found the diagram harder to follow than reading through the struct.. which is why i added the struct in the comments, but also variables detailing the index of each field in the struct. hoping that's sufficient for readers.. also, ascii drawing is hard 😂

contracts/test/setup.ts
cfal: yes, we only really need it for the L1-> L2 test, and we could switch the rest to starknet.js and our own util functions (eg. `startNetwork`).. and we could avoid having to work within hardhat-starknet-plugin's constraints like this file is doing. will leave it for a future changeset

contracts/package.json
cfal: it's possible, i haven't tested it

packages-ts/starknet-gauntlet-example/src/lib/contracts.ts
```diff
@@ -1,14 +1,25 @@
...
+      ),
+    ),
+    casm: json.parse(
+      fs.readFileSync(
+        `${__dirname}/../../../../contracts/target/release/chainlink_${name}.casm.json`,
+        'utf-8',
+      ),
+    ),
+  }
 }
```
archseer: Since the paths are now flattened out, this `loadContract` function looks the same everywhere (except it's typed by `CONTRACT_LIST`). Let's extract it and share it across the files? Would make it easier to update

.github/workflows/integration_gauntlet.yml
```diff
@@ -14,11 +14,15 @@ jobs:
...
       - name: Install Nix
         uses: cachix/install-nix-action@29bd9290ef037a3ecbdafe83cbd2185e9dd0fa0a # v20
         with:
           nix_path: nixpkgs=channel:nixos-unstable
 
+      - name: Install Cairo
+        uses: ./.github/actions/install-cairo
+
       - name: Install starknet-devnet (via venv+pip)
         run: pip install -r contracts/requirements.txt -c contracts/constraints.txt
```
archseer:
```suggestion
        run: nix develop -c pip install -r contracts/requirements.txt -c contracts/constraints.txt
```

contracts/hardhat.config.ts
```diff
@@ -42,25 +42,29 @@ const config: HardhatUserConfig = {
...
```
archseer, line 55: Is this still necessary? I was hoping to remove the `vendor/` dirs later on since we can use binary artifacts

archseer: I guess it's necessary until devnet supports the latest rc0

packages-ts/starknet-gauntlet-token/src/commands/token/deploy.ts
```diff
@@ -8,15 +8,17 @@ import { CATEGORIES } from '../../lib/categories'
 import { tokenContractLoader, CONTRACT_LIST } from '../../lib/contracts'
 
 type UserInput = {
+  minter?: string
   owner?: string
 }
 
-type ContractInput = [owner: string]
+type ContractInput = [minter: string, owner: string]
```
archseer, line 15: Why the need for this new input?

packages-ts/starknet-gauntlet/src/encoding/index.ts
```diff
@@ -0,0 +1,78 @@
...
+  let len = data.byteLength
+  felts.push(num.toBigInt(len).toString())
+  // chunk every 31 bytes
+  for (let i = 0; i < data.length; i += CHUNK_SIZE) {
+    const chunk = data.slice(i, i + CHUNK_SIZE)
+    // cast to int
+    felts.push(new BN(chunk, 'be').toString())
+  }
+  return felts
+}
```
archseer: Could we move this to the test suite to remove the `bn.js` import here?

archseer (commented):
LGTM just some minor comments!

contracts/hardhat.config.ts
cfal: at last check, it was still necessary because the compiled casm hash is different from rc0 yeah

packages-ts/starknet-gauntlet-token/src/commands/token/deploy.ts
cfal: https://github.com/smartcontractkit/chainlink-starknet/pull/251/files#diff-11b0c99c87d3f83241a188380c3ca06bb049e6a77cdb4ac9589eed751a945353R47

packages-ts/starknet-gauntlet/src/encoding/index.ts
cfal: ah snap. i forgot to move the tests, thanks

packages-ts/starknet-gauntlet-example/src/lib/contracts.ts
cfal: done

archseer (approved):
Looks good, great work!

PR #242, opened 2023-04-28
BCI-992: Add Gas Adjustment

augustbleeds:
For more context please look at https://www.notion.so/chainlink/Add-K-value-to-Validator-Contract-e21e7e9a85d04da48071535c249a2fe9?pvs=4


contracts/src/chainlink/solidity/emergency/StarknetValidator.sol
```diff
@@ -22,9 +22,13 @@ interface AggregatorValidatorInterface {
 /// @title StarknetValidator - makes cross chain calls to update the Sequencer Uptime Feed on L2
 contract StarknetValidator is TypeAndVersionInterface, AggregatorValidatorInterface, SimpleWriteAccessController {
   // Config for L1 -> L2 message cost approximation
+  // Calculated as k * gasEstimate * gasPriceL1Feed / 100
   struct GasConfig {
+    // gas units derived from starknet estimate_message_fee
     uint256 gasEstimate;
     address gasPriceL1Feed;
+    // scalar coefficient
+    uint32 k;
```
archseer: Both the comment and the unit name aren't very descriptive. We used `gasAdjustment` on cosmos: https://github.com/smartcontractkit/chainlink-cosmos/blob/837a92027ab677d8b5a0168864c7fa03ef91a249/contracts/ocr2/src/contract.rs#L1498-L1500

And it's in terms of percent (e.g. 140 = scale the resulting estimate to 140% to give some buffer)

augustbleeds: cool, I'll update the names

augustbleeds: I'll consider making it more intuitive percent-wise

augustbleeds: Updated the comments. It should work exactly the same as cosmos. 100 = 100% of original value

archseer (dismissed):

contracts/src/chainlink/solidity/emergency/StarknetValidator.sol
```diff
@@ -50,7 +54,7 @@ contract StarknetValidator is TypeAndVersionInterface, AggregatorValidatorInterf
   error AccessForbidden();
 
   /// @notice This event is emitted when the gas config is set.
-  event GasConfigSet(uint256 gasEstimate, address indexed gasPriceL1Feed);
+  event GasConfigSet(uint256 gasEstimate, address indexed gasPriceL1Feed, uint32 k);
```
krebernisak: Why a different name here?

```suggestion
  event GasConfigSet(uint256 gasEstimate, address indexed gasPriceL1Feed, uint32 gasAdjustment);
```

contracts/src/chainlink/solidity/emergency/StarknetValidator.sol
```diff
@@ -22,9 +22,13 @@ interface AggregatorValidatorInterface {
 /// @title StarknetValidator - makes cross chain calls to update the Sequencer Uptime Feed on L2
 contract StarknetValidator is TypeAndVersionInterface, AggregatorValidatorInterface, SimpleWriteAccessController {
   // Config for L1 -> L2 message cost approximation
+  // Message Cost = gasAdjustment * gasEstimate * gasPriceL1Feed / 100
   struct GasConfig {
+    // gas units derived from starknet estimate_message_fee
     uint256 gasEstimate;
     address gasPriceL1Feed;
+    // gasAdjustment of 100 equals 1x (see setGasConfig for more info)
+    uint32 gasAdjustment;
```
krebernisak, line 24: Let's add an in-line comment about what's our default recommendation (e.g., 130).

We should also add the same check to Gauntlet commands deploying or configuring the `StarknetValidator` contract.

tldr; we should align on the default, document the default, and Gauntlet should ask the user to confirm if he's using anything other than the recommended default.

contracts/src/chainlink/solidity/emergency/StarknetValidator.sol
```diff
@@ -22,9 +22,13 @@ interface AggregatorValidatorInterface {
 /// @title StarknetValidator - makes cross chain calls to update the Sequencer Uptime Feed on L2
 contract StarknetValidator is TypeAndVersionInterface, AggregatorValidatorInterface, SimpleWriteAccessController {
   // Config for L1 -> L2 message cost approximation
+  // Message Cost = gasAdjustment * gasEstimate * gasPriceL1Feed / 100
   struct GasConfig {
+    // gas units derived from starknet estimate_message_fee
     uint256 gasEstimate;
```
krebernisak, line 19: What is the estimate given by `estimate_message_fee`? Please document the approximate value in-line here, and also try to add a default + check to Gauntlet if users use a substantially different value. This value should be pretty static, until our L2 contract changes, or Starknet introduces big gas metering changes.

krebernisak: Please also update/add info here: https://github.com/smartcontractkit/chainlink-starknet/tree/develop/docs/emergency-protocol

contracts/src/chainlink/solidity/emergency/StarknetValidator.sol
krebernisak: Please also update/add info here: https://github.com/smartcontractkit/chainlink-starknet/tree/develop/docs/emergency-protocol

contracts/src/chainlink/solidity/emergency/StarknetValidator.sol
```diff
@@ -176,8 +181,13 @@ contract StarknetValidator is TypeAndVersionInterface, AggregatorValidatorInterf
 
   /// @notice L1 oracle is asked for a fast L1 gas price, and the price multiplied by the configured gas estimate
   function _approximateFee() internal view returns (uint256) {
+    uint256 gasPrice = approximateGasPrice();
+    return gasPrice * s_gasConfig.gasEstimate;
+  }
+
+  function approximateGasPrice() public view returns (uint256) {
```
krebernisak, line 181: We now changed the API surface of our contract, just to test...

A better way to do this would be to keep the method internal, then extend the contract with another (test contract) and there expose the functionality we need to test.

We're short on time, so leave it as is, but please at least add a comment to this public fn.

contracts/src/chainlink/solidity/emergency/StarknetValidator.sol
augustbleeds: good catch :)

contracts/src/chainlink/solidity/emergency/StarknetValidator.sol
augustbleeds: I'll need to deploy on alpha-goerli to test the value. Will do it today

augustbleeds: The value I got was 17266 units. 
```
tarknet estimate_message_fee --from_address 0x00 --address=0x06f4279f832de1afd94ab79aa1766628d2c1e70bc7f74bfba3335db8e728a7e6 --function update_status --inputs 0x1 123123
```

augustbleeds: Which comes out to around $3 based on today's gas price

augustbleeds: I don't think we should add a default to gauntlet because for testing you'd like a negligible fee in order to make it easier to test.

augustbleeds: Update: I added an error for gasAdjustment if it's below 100

contracts/src/chainlink/solidity/emergency/StarknetValidator.sol
augustbleeds: Do we want to decide on 130?

archseer:
Once this is merged, I'll rebase the `cairo-1.0` branch again. It's currently not running a lot of the tests because of conflicts with develop

contracts/src/chainlink/solidity/emergency/StarknetValidator.sol
krebernisak: > I don't think we should add a default to gauntlet because for testing you'd like a negligible fee in order to make it easier to test.

Wouldn't such a test fail with the incorrect fee?

One of the main use cases for Gauntlet is to save operators from making mistakes and educate them about the operations they are executing. So I'd suggest we add the proper checks in, and then we might add a flag to avoid them so we can tweak execution in tests and while debugging.

For example, we could validate so `gasAdjustment` can't go below 100, unless an explicit `--unsafe-skip-validations` flag is set.

contracts/src/chainlink/solidity/emergency/StarknetValidator.sol
krebernisak: Let's settle on 130 for now, be explicit about it (docs, contract, gauntlet), and point it out to the auditors. Easy for us to change just before the launch, or update at any point.

Think about if there is a way for us to devise an integration test, which probably needs to use real mainnet contracts (L1 fast gas feed, L2 upkeep) + starknet CLI to check the current values and make sure they are above some threshold we set.

augustbleeds: Yeah, I added it to the docs. The check for gas price would be better suited for continuous monitoring (grafana alerting) because an integration test is only ran periodically. It should be straightforward -- query approximateGasPrice() and compare it to the gas price returned by the sequencer. I can create a ticket to add a grafana alert

contracts/src/chainlink/solidity/emergency/StarknetValidator.sol
augustbleeds: The test won't fail on devnet or testnet with a negligible fee as long as the fee is not 0. But I guess it's probably safer to have the same defaults across all devnet/testnet/mainnet for simplicity sake. I'll update the docs and change the test to include a safety flag

contracts/src/chainlink/solidity/emergency/StarknetValidator.sol
krebernisak: Might be useful for monitoring, so let's leave it in.

contracts/test/emergency/StarknetValidator.test.ts
```diff
@@ -461,7 +514,7 @@ describe('StarknetValidator', () => {
       expect(mockStarknetMessaging.address).to.hexEqual(loadedFrom)
     })
 
-    it('should send a message to the L2 contract', async () => {
+    it.only('should send a message to the L2 contract', async () => {
```
krebernisak: Debugging artifact?

```suggestion
    it('should send a message to the L2 contract', async () => {
```

packages-ts/starknet-gauntlet-emergency-protocol/README.md
```diff
@@ -19,12 +19,16 @@ This deploys a new instance of the `StarknetValidator` contract on **L1**
 `<SOURCE>` is the address of source aggregator. Source aggregator should be added to access control of starknet validator contract to be able to call
 `validate` on the starknet validator
 
-`<GAS_ESTIMATE>` is a number that represents l1 gas estimate
+`<GAS_ESTIMATE>` is a number that represents l1 gas estimate. We recommend a value of 17300 on mainnet. On devnet or testnet you may choose 1 to save gas.
```
krebernisak: This depends on the receiving contract, so it would be valuable to add the specific contract version here. Once that contract changes, the gas will value will most likely change.

krebernisak (dismissed):

contracts/test/emergency/StarknetValidator.test.ts
augustbleeds: yup haha

packages-ts/starknet-gauntlet-emergency-protocol/README.md
augustbleeds: True, since we don't have versions I'll just add something that references the date of writing

contracts/test/emergency/StarknetValidator.test.ts
augustbleeds: good catch

packages-ts/starknet-gauntlet-emergency-protocol/README.md
krebernisak: Hmm, we should have versions for each contract inside the contract like [this](https://github.com/smartcontractkit/chainlink-starknet/blob/develop/contracts/src/chainlink/cairo/emergency/SequencerUptimeFeed/sequencer_uptime_feed.cairo#LL78C6-L78C22):

```
@view
func type_and_version{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}() -> (
    meta: felt
) {
    let (meta) = SequencerUptimeFeed.type_and_version();
    return (meta,);
}
```

As influenced from here [ITypeAndVersion.sol](https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.8/shared/interfaces/ITypeAndVersion.sol).
Can you please double-check and add it to the contracts missing? cc @archseer

augustbleeds: You're right, we do have versions for them

augustbleeds: @krebernisak added the version # to the README docs

krebernisak: Do we have versions for all contracts? Please double-check.

krebernisak (approved):

PR #251, opened 2023-05-04
BCI-1259: Add upgradeability

augustbleeds:
https://smartcontract-it.atlassian.net/browse/BCI-1259

The reasoning behind making contracts upgradeable is because in the shifting landscape of Starknet, there may be feature deprecation and/or unknown changes required. ~This PR only makes the core data feed contracts (uptime feed, aggregator, and proxy aggregator) upgradeable. However, I think we should discuss if we want to make the multisig contract and link token upgradeable because they'd be equally affected by breaking changes as the core data feed contracts.~ We reached a team decision to make all contracts that we own upgradeable

Notes:
* ~Contracts such as Ownable, SimpleReadAccessController, etc. don't have upgrade external functions because they are used as libraries and never intended to be deployed by themselves~ access controllers need to be deployed as contracts so they are upgradeable as well
* ~I included the Ownable:assert_only_owner() in the upgradeable library~ ownable assertion not imported to reduce dependencies
* I can't actually test the replace class hash syscall yet because cairo test does not support it yet (even in rc0) --> we will test the functionality on devnet in a seperate PR

### DIFF
--- a/contracts/src/emergency/sequencer_uptime_feed.cairo
+++ b/contracts/src/emergency/sequencer_uptime_feed.cairo
@@ -145,9 +145,11 @@ mod SequencerUptimeFeed {
     /// Upgradeable
     ///
 
+    // todo add test calling with non owner once calvin's changes are merged
     #[external]
     fn upgrade(impl_hash: ClassHash) {
-        Upgradeable::upgrade_only_owner(impl_hash)
+        Ownable::assert_only_owner();
+        Upgradeable::_upgrade(impl_hash)
     }
 
     ///
--- a/contracts/src/emergency/sequencer_uptime_feed.cairo
+++ b/contracts/src/emergency/sequencer_uptime_feed.cairo
@@ -7,6 +7,8 @@ mod SequencerUptimeFeed {
     use starknet::storage_read_syscall;
     use starknet::storage_write_syscall;
     use starknet::storage_address_from_base_and_offset;
+    use starknet::class_hash::ClassHash;
+    use starknet::class_hash::ClassHashZeroable;
 
     use box::BoxTrait;
     use traits::Into;
@@ -21,6 +23,7 @@ mod SequencerUptimeFeed {
     use chainlink::ocr2::aggregator::IAggregator;
     use chainlink::ocr2::aggregator::Aggregator::Transmission;
     use chainlink::ocr2::aggregator::Aggregator::TransmissionStorageAccess;
+    use chainlink::libraries::upgradeable::Upgradeable;
 
     const ETH_ADDRESS_BOUND: felt252 = 0x10000000000000000000000000000000000000000; // 2**160
 
@@ -139,6 +142,14 @@ mod SequencerUptimeFeed {
         _l1_sender::read()
     }
 
+    ///
+    /// Upgradeable
+    ///
+
+    #[external]
+    fn upgrade(impl_hash: ClassHash) {
+        Upgradeable::upgrade(impl_hash)
+    }
 
     ///
     /// Aggregator
--- a/contracts/src/emergency/sequencer_uptime_feed.cairo
+++ b/contracts/src/emergency/sequencer_uptime_feed.cairo
@@ -8,7 +8,6 @@ mod SequencerUptimeFeed {
     use starknet::storage_write_syscall;
     use starknet::storage_address_from_base_and_offset;
     use starknet::class_hash::ClassHash;
-    use starknet::class_hash::ClassHashZeroable;
 
     use box::BoxTrait;
     use traits::Into;
@@ -148,7 +147,7 @@ mod SequencerUptimeFeed {
 
     #[external]
     fn upgrade(impl_hash: ClassHash) {
-        Upgradeable::upgrade(impl_hash)
+        Upgradeable::upgrade_only_owner(impl_hash)
     }
 
     ///
--- a/contracts/src/emergency/sequencer_uptime_feed.cairo
+++ b/contracts/src/emergency/sequencer_uptime_feed.cairo
@@ -147,9 +147,9 @@ mod SequencerUptimeFeed {
 
     // todo add test calling with non owner once calvin's changes are merged
     #[external]
-    fn upgrade(new_class_hash: ClassHash) {
+    fn upgrade(new_impl: ClassHash) {
         Ownable::assert_only_owner();
-        Upgradeable::upgrade(new_class_hash)
+        Upgradeable::upgrade(new_impl)
     }
 
     ///
--- a/contracts/src/emergency/sequencer_uptime_feed.cairo
+++ b/contracts/src/emergency/sequencer_uptime_feed.cairo
@@ -147,9 +147,9 @@ mod SequencerUptimeFeed {
 
     // todo add test calling with non owner once calvin's changes are merged
     #[external]
-    fn upgrade(impl_hash: ClassHash) {
+    fn upgrade(new_class_hash: ClassHash) {
         Ownable::assert_only_owner();
-        Upgradeable::_upgrade(impl_hash)
+        Upgradeable::upgrade(new_class_hash)
     }
 
     ///
--- a/contracts/src/libraries/upgradeable.cairo
+++ b/contracts/src/libraries/upgradeable.cairo
@@ -6,20 +6,13 @@ mod Upgradeable {
     use starknet::class_hash::ClassHash;
     use starknet::class_hash::ClassHashZeroable;
 
-    use chainlink::libraries::ownable::Ownable;
-
     #[event]
     fn Upgraded(implementation: ClassHash) {}
-
-    fn upgrade_only_owner(impl_hash: ClassHash) {
-        Ownable::assert_only_owner();
-        upgrade(impl_hash);
-    }
 
     // this method assumes replace_class_syscall has a very low possibility of being deprecated
     // but if it does, we will either have upgraded the contract to be non-upgradeable by then
     // because the starknet ecosystem has stabilized or we will be able to upgrade the contract to the proxy pattern
-    fn upgrade(impl_hash: ClassHash) {
+    fn _upgrade(impl_hash: ClassHash) {
         assert(!impl_hash.is_zero(), 'Class hash cannot be zero');
         replace_class_syscall(impl_hash);
         Upgraded(impl_hash);
--- a/contracts/src/libraries/upgradeable.cairo
+++ b/contracts/src/libraries/upgradeable.cairo
@@ -16,6 +16,9 @@ mod Upgradeable {
         upgrade(impl_hash);
     }
 
+    // this method assumes replace_class_syscall has a very low possibility of being deprecated
+    // but if it does, we will either have upgraded the contract to be non-upgradeable by then
+    // because the starknet ecosystem has stabilized or we will be able to upgrade the contract to the proxy pattern
     fn upgrade(impl_hash: ClassHash) {
         assert(!impl_hash.is_zero(), 'Class hash cannot be zero');
         replace_class_syscall(impl_hash);
--- a/contracts/src/libraries/upgradeable.cairo
+++ b/contracts/src/libraries/upgradeable.cairo
@@ -2,19 +2,21 @@
 mod Upgradeable {
     use zeroable::Zeroable;
 
+    use starknet::SyscallResult;
+    use starknet::SyscallResultTrait;
     use starknet::syscalls::replace_class_syscall;
     use starknet::class_hash::ClassHash;
     use starknet::class_hash::ClassHashZeroable;
 
     #[event]
-    fn Upgraded(implementation: ClassHash) {}
+    fn Upgraded(new_class_hash: ClassHash) {}
 
     // this method assumes replace_class_syscall has a very low possibility of being deprecated
     // but if it does, we will either have upgraded the contract to be non-upgradeable by then
     // because the starknet ecosystem has stabilized or we will be able to upgrade the contract to the proxy pattern
-    fn _upgrade(impl_hash: ClassHash) {
-        assert(!impl_hash.is_zero(), 'Class hash cannot be zero');
-        replace_class_syscall(impl_hash);
-        Upgraded(impl_hash);
+    fn upgrade(new_class_hash: ClassHash) {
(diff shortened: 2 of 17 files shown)


contracts/src/libraries/upgradeable.cairo
```diff
@@ -0,0 +1,20 @@
...
+    use starknet::syscalls::replace_class_syscall;
+    use starknet::class_hash::ClassHash;
+    use starknet::class_hash::ClassHashZeroable;
+
+    use chainlink::libraries::ownable::Ownable;
+
+    #[event]
+    fn Upgraded(implementation: ClassHash) {}
+
+    fn upgrade(impl_hash: ClassHash) {
```
krebernisak: Can we check if this `ClassHash` exists and is declared?

contracts/src/libraries/upgradeable.cairo
```diff
@@ -0,0 +1,20 @@
...
+        Ownable::assert_only_owner();
+        assert(!impl_hash.is_zero(), 'Class hash cannot be zero');
+        replace_class_syscall(impl_hash);
+        Upgraded(impl_hash);
```
krebernisak: Let's also emit the previous implementation.

contracts/src/libraries/upgradeable.cairo
```diff
@@ -0,0 +1,20 @@
+
+mod Upgradeable {
```
krebernisak, line 2: Please add some quick inline docs about this contract.

e.g. "This method assumes that the replace_call syscall will not be deprecated and the Proxy pattern in Cairo 0.x will not be the dominant pattern for upgradeable contracts in Cairo 1.x."

contracts/src/libraries/upgradeable.cairo
archseer: The implementation here is copied verbatim from OZ, I think we can't check for it but `replace_class_syscall` will fail if it doesn't exist

contracts/src/libraries/upgradeable.cairo
augustbleeds: That's a good idea, but I don't think that there is currently any way to retrieve the class hash of an address on-chain

contracts/src/libraries/upgradeable.cairo
augustbleeds: Sounds good. I'll add extra notes as well to mention where it's inspired from

contracts/src/libraries/upgradeable.cairo
```diff
@@ -0,0 +1,27 @@
...
+
+    use chainlink::libraries::ownable::Ownable;
+
+    #[event]
+    fn Upgraded(implementation: ClassHash) {}
+
+    fn upgrade_only_owner(impl_hash: ClassHash) {
+        Ownable::assert_only_owner();
+        upgrade(impl_hash);
+    }
```
calvwang9: Curious about the choice to bake in the authorization method (Ownable::assert_only_owner()) into the Upgradeable library - would it be better to let the parent contract handle authentication, and simply provide an unprotected `upgrade` function in the library?

calvwang9: Alternatively, [OZ's Upgradable library impl](https://github.com/OpenZeppelin/cairo-contracts/pull/603/files#diff-41a9680000506ee92336bbd3f923e942a4e198ae75785c29be3c2def2910d0ea) maintains its own self-contained, simple auth mechanism, which means that there are no dependencies between libraries = less risk, clearer separation of responsibilities (though may not be much of a concern for our use cases?)

archseer: Yeah this is based on the OZ implementation and I think we should follow the comment here: https://github.com/OpenZeppelin/cairo-contracts/pull/603#pullrequestreview-1406490206

Only exposing the `_upgrade` utility method and the event. That way we can swap this copy out for the OZ library once stabilized.

contracts/src/tests/test_multisig.cairo
```diff
@@ -280,3 +281,25 @@ fn test_execute_confirmation_below_threshold() {
...
+#[available_gas(2000000)]
+#[should_panic(expected = ('only multisig allowed', ))]
+fn test_upgrade_not_multisig() {
+    let account = contract_address_const::<777>();
+    set_caller_address(account);
+
+    Multisig::upgrade(class_hash_const::<1>())
+}
+
+// wait until replace_class_syscall is implemented by cairo test
```
krebernisak: Does Devnet support `replace_class_syscall`?
Can we add a couple of dummy integration contract upgrade tests?

contracts/src/libraries/upgradeable.cairo
```diff
@@ -11,8 +11,15 @@ mod Upgradeable {
...
-    fn upgrade(impl_hash: ClassHash) {
+    fn upgrade_only_owner(impl_hash: ClassHash) {
         Ownable::assert_only_owner();
+        upgrade(impl_hash);
+    }
+
+    // this method assumes replace_class_syscall has a very low possibility of being deprecated
+    // but if it does, we will either have upgraded the contract to be non-upgradeable by then
+    // because the starknet ecosystem has stabilized or we will be able to upgrade the contract to the proxy pattern
+    fn upgrade(impl_hash: ClassHash) {
```
krebernisak: This seems dangerous as the caller might forget to add access controls or expect they are already in place.

Can we have just one `upgrade` entry point which internally always checks access by using an abstract fn like `assert_only_upgrade_admin()`, which then downstream contracts need to implement?

Ownable contracts can implement using `Ownable::assert_only_owner();` while others have their own preferred mechanism. 

The point is this will be enforced and explicit.

contracts/src/libraries/upgradeable.cairo
krebernisak: Agreed. This shouldn't be coupled with `Ownable`, and access control should be explicitly required for every upgrade path - added a suggestion here: https://github.com/smartcontractkit/chainlink-starknet/pull/251#discussion_r1185848788

contracts/src/libraries/upgradeable.cairo
augustbleeds: There's no class inheritence (and abstract classes/functions) in cairo, :(. I think as mentioned above we'll keep upgrade non-protected and assert ownership when necessary

contracts/src/tests/test_multisig.cairo
augustbleeds: I think it should. I had some difficulty getting cairo-1 tests to run on starknet-devnet though but will try again today

contracts/src/libraries/upgradeable.cairo
augustbleeds: Since we're using the contracts in a very controlled and specific manner, I did it this way to save some code duplication (asserting owner at every contract that uses this library) and that way we'd also only need to test the upgradable.cairo for core logic. I'll change it to be less coupled

contracts/src/tests/test_multisig.cairo
augustbleeds: I'll wait until Alex's PR https://github.com/smartcontractkit/chainlink-starknet/pull/240/files is merged to add these contract upgrade tests because he has logic to get cairo-1 contracts deployed and working on devnet.

contracts/src/libraries/upgradeable.cairo
```diff
@@ -0,0 +1,20 @@
...
+    use starknet::class_hash::ClassHash;
+    use starknet::class_hash::ClassHashZeroable;
+
+    #[event]
+    fn Upgraded(implementation: ClassHash) {}
+
+    // this method assumes replace_class_syscall has a very low possibility of being deprecated
+    // but if it does, we will either have upgraded the contract to be non-upgradeable by then
+    // because the starknet ecosystem has stabilized or we will be able to upgrade the contract to the proxy pattern
+    fn _upgrade(impl_hash: ClassHash) {
```
cfal: could we be consistent with the var names? ie `impl_hash`, `implementation` both used here. since the syscall param is also called `class_hash`, that seems like a good choice as well, or something like `new_class_hash`

contracts/src/libraries/upgradeable.cairo
```diff
@@ -0,0 +1,20 @@
...
+        assert(!impl_hash.is_zero(), 'Class hash cannot be zero');
+        replace_class_syscall(impl_hash);
```
cfal: don't you need to unwrap_syscall the result to make sure it succeeded?

contracts/src/multisig.cairo
```diff
@@ -219,6 +223,12 @@ mod Multisig {
 
     /// Externals
 
+    #[external]
+    fn upgrade(impl_hash: ClassHash) {
+        _require_multisig();
+        Upgradeable::_upgrade(impl_hash)
```
cfal: if `_upgrade` is exposed and used by contracts, do we need the underline prefix? we don't have one for our other helper classes ie SimpleReadAccessController

contracts/src/libraries/upgradeable.cairo
augustbleeds: Good catch, I don't know why it isn't included in the OZ PR because it'd silently fail then.

contracts/src/multisig.cairo
augustbleeds: I'll remove the underscore

contracts/src/libraries/upgradeable.cairo
augustbleeds: That works 👍

archseer:
> * Contracts such as Ownable, SimpleReadAccessController, etc. don't have upgrade external functions because they are used as libraries and never intended to be deployed by themselves

SimpleReadAccessController/WriteAccessController are actually meant to work both as a library and as a standalone controller. The billing access controller used by the aggregator is deployed as a separate instance with it's own contract address for example.

cfal (approved):

archseer (changes requested):
> * Contracts such as Ownable, SimpleReadAccessController, etc. don't have upgrade external functions because they are used as libraries and never intended to be deployed by themselves

SimpleReadAccessController/WriteAccessController are actually meant to work both as a library and as a standalone controller. The billing access controller used by the aggregator is deployed as a separate instance with it's own contract address for example.

calvwang9 (approved):

calvwang9:
> > * Contracts such as Ownable, SimpleReadAccessController, etc. don't have upgrade external functions because they are used as libraries and never intended to be deployed by themselves
> 
> SimpleReadAccessController/WriteAccessController are actually meant to work both as a library and as a standalone controller. The billing access controller used by the aggregator is deployed as a separate instance with it's own contract address for example.

@archseer Could we refactor this into separate library and standalone versions then? I'd prefer this for simplicity and also to decouple ownable from that library (not for this PR)

contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -145,9 +145,11 @@ mod SequencerUptimeFeed {
     /// Upgradeable
     ///
 
+    // todo add test calling with non owner once calvin's changes are merged
     #[external]
-    fn upgrade(impl_hash: ClassHash) {
-        Upgradeable::upgrade_only_owner(impl_hash)
+    fn upgrade(new_class_hash: ClassHash) {
+        Ownable::assert_only_owner();
+        Upgradeable::upgrade(new_class_hash)
```
krebernisak: There is no need to repeat the type name as part of the var name:

```suggestion
    fn upgrade(new_impl: ClassHash) {
        Ownable::assert_only_owner();
        Upgradeable::upgrade(new_impl)
```

contracts/src/libraries/upgradeable.cairo
```diff
@@ -2,26 +2,21 @@
...
+    fn Upgraded(new_class_hash: ClassHash) {}
 
     // this method assumes replace_class_syscall has a very low possibility of being deprecated
     // but if it does, we will either have upgraded the contract to be non-upgradeable by then
     // because the starknet ecosystem has stabilized or we will be able to upgrade the contract to the proxy pattern
-    fn upgrade(impl_hash: ClassHash) {
-        assert(!impl_hash.is_zero(), 'Class hash cannot be zero');
-        replace_class_syscall(impl_hash);
-        Upgraded(impl_hash);
+    fn upgrade(new_class_hash: ClassHash) {
```
krebernisak:
```suggestion
    #[internal]
    fn upgrade(new_class_hash: ClassHash) {
```

contracts/src/libraries/upgradeable.cairo
krebernisak: Ok, sounds good. I couldn't find a better alternative so looks like we'll leave it as is. OZ Cairo lib implementations also use a similar pattern.

contracts/src/token/link_token.cairo
```diff
@@ -74,6 +79,44 @@ mod LinkToken {
...
+    }
+
+    #[external]
+    fn renounce_ownership() {
+        Ownable::renounce_ownership()
+    }
+
 
     //
     // ERC20
```
krebernisak, line 122: There is a TODO comment at the end of this contract:

```
fn only_minter() {
        let caller = starknet::get_caller_address();
        let minter = minter();
        // todo: can we remove this check because it's already in the constructor
        assert(!minter.is_zero(), 'minter is 0');
        assert(caller == minter, 'only minter');
    }
```

Please double-check with @archseer, but I think this should be safe to remove as `caller` should never be `0` afaik.

augustbleeds:
> > * Contracts such as Ownable, SimpleReadAccessController, etc. don't have upgrade external functions because they are used as libraries and never intended to be deployed by themselves
> 
> SimpleReadAccessController/WriteAccessController are actually meant to work both as a library and as a standalone controller. The billing access controller used by the aggregator is deployed as a separate instance with it's own contract address for example.

Gotcha, thanks for clarifying. I'll add the upgrade functionality to them as well

contracts/src/token/link_token.cairo
augustbleeds: I left this comment originally. I'm thinking the same thing...I think we can remove it since the constructor already checks that minter is not zero. Minter will be the starkgate contract if I recall @archseer

augustbleeds: ^ went ahead and removed it

PR #255, opened 2023-05-08
[BCI-1261] On demand soak

smickovskid:


.github/workflows/integration-tests-smoke.yml
```diff
@@ -33,15 +33,15 @@ jobs:
       - name: Collect Metrics
         if: always()
         id: collect-gha-metrics
-        uses: smartcontractkit/push-gha-metrics-action@808c183d1f5c26a4f3fd50b75d3cf7b58d9aa293
+        uses: smartcontractkit/push-gha-metrics-action@90fcbaac8ebf86da9c4d55dba24f6fe3029f0e0b
         with:
-          basic-auth: ${{ secrets.GRAFANA_CLOUD_BASIC_AUTH }}
-          hostname: ${{ secrets.GRAFANA_CLOUD_HOST }}
+          basic-auth: TEST_AUTH
+          hostname: TEST_AUTH
```
aalu1418: just curious, where do these secrets come from?

integration-tests/soak/ocr2_test.go
```diff
@@ -4,15 +4,13 @@ package soak_test
...
-	"testing"
-	"time"
-
 	"github.com/smartcontractkit/chainlink-starknet/integration-tests/common"
 	"github.com/smartcontractkit/chainlink-starknet/ops/gauntlet"
 	"github.com/smartcontractkit/chainlink-starknet/ops/utils"
 	"github.com/smartcontractkit/chainlink/integration-tests/actions"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap/zapcore"
+	"testing"
```
aalu1418: kinda weird formatting? (shouldn't `testing` be at the top with the other native packages)

.github/workflows/integration-tests-soak.yml
```diff
@@ -81,9 +90,10 @@ jobs:
     env:
       CHAINLINK_ENV_USER: ${{ github.actor }}
       L2_RPC_URL: ${{ github.event.inputs.l2_rpc_url }}
+      TEST_DURATION: ${{ github.event.inputs.test_duration }}
       NODE_COUNT: ${{ github.event.inputs.node_count }}
-      PRIVATE_KEY: ${{ secrets.GOERLI_PRIVATE_KEY }}
-      ACCOUNT: ${{ secrets.GOERLI_ACCOUNT }}
+      PRIVATE_KEY: ${{ github.event.inputs.private_key }}
+      ACCOUNT: ${{ github.event.inputs.account }}
```
aalu1418, line 96: so secrets aren't stored on GH anymore - to me that implies that we will have to manually run soaks each time? or is the plan to still run it via CI?

if the plan is to manually start them each time - can we remove the trigger that automatically runs soaks after PR merge. not sure if this is true, but would it seems like these runs after merge would fail since the secrets wouldn't be present?

.github/workflows/integration-tests-smoke.yml
smickovskid: That is currently being figured out. I have asked multiple people and handling that in another task. These are the grafana secrets that we use to send metrics so we can parse the results

.github/workflows/integration-tests-soak.yml
smickovskid: Valid point, I overlooked this. I will set up a default value so the workflow continues to work as before but also allows us to kick off on demand soaks for Testnet

smickovskid: Oh wait, I just realized our regular trigger on merge runs a soak against devnet where we don't need the Account and private key. This is specifically for on demand test to run on Testnet. I added to option to specify our own Private keys and Account. The test logic checks if we have defined L2_RPC_URL and only then assumes we are testing on Testnet, otherwise it ignores and runs on devnet

aalu1418 (approved):
lgtm!

PR #256, opened 2023-05-08
BCI-1259: Add upgradeability tests in hardhat

augustbleeds:
Note: 
* You only need to look at [contracts/test/upgradeable.test.ts](https://github.com/smartcontractkit/chainlink-starknet/pull/256/files#diff-9a6c6995a53ff6b3a247a61bd08941da93994731bc5a9bea825fb2fc8ea1c723) . ~The rest of the diffs come because [the parent](https://github.com/smartcontractkit/chainlink-starknet/pull/240) branch is not updated to have upgradeability features yet~
* I rebased on Alex's https://github.com/smartcontractkit/chainlink-starknet/pull/240 PR
* ~I had to update the hardhat plugin to accept the ClassHash input arguments. I will create a quick PR to add that to the shardlabs branch.~ it was merged but we still need to build the dist/ ourselves on Alex's fork


contracts/test/upgradeable.test.ts
```diff
@@ -0,0 +1,83 @@
...
+                'upgrade',
+                {
+                    new_impl: newClassHash
+                }
+            )
+
+            // should error because the contract has upgraded and no longer has foo function
+            await expectCallError(upgradeableContract.call('foo'))
+
+            // we're not able to directly call bar via hardhat StarknetContract because
```
cfal: you can just do nonUpgradeableFactory.getContractAt() with your contract address to get an updated `StarknetContract` instance

augustbleeds: Thanks 👍 that's much better haha

contracts/package.json
```diff
@@ -4,18 +4,19 @@
...
   },
   "author": "",
   "license": "MIT",
   "devDependencies": {
+    "@starkware-industries/starkware-crypto-utils": "^0.1.0",
     "@ethereum-waffle/mock-contract": "^4.0.4",
     "@nomicfoundation/hardhat-chai-matchers": "^1.0.3",
     "@nomiclabs/hardhat-ethers": "^2.0.5",
-    "@shardlabs/starknet-hardhat-plugin": "^0.8.0-alpha.0",
+    "@shardlabs/starknet-hardhat-plugin": "git+https://github.com/augustbleeds/starknet-hardhat-plugin#ae53af4",
```
cfal: your PR got merged 🥳 so i pointed the deploy PR to latest tip-of-tree on the spaceshard repo (realized i could have done this before instead of using a fork as well)

so you should be able to remove this change now

cfal (commented):
can you rebase? the changeset should be much easier to see now

contracts/package.json
cfal: ah, i remember now that doesn't work because dist isn't automatically built for the plugin like it is for starknetjs

augustbleeds: Yeah, why is that? It has a yarn build command, so I don't know why npm/yarn doesn't pick that up

augustbleeds: But cool, I'll rebase on your branch

contracts/test/upgradeable.test.ts
```diff
@@ -0,0 +1,73 @@
...
+            expect((await afterUpgradeContract.call('bar')).response).to.equal(true)
+        })
+
+        it('reverts if new implementation class hash does not exist', async () => {
+            expect((await upgradeableContract.call('foo')).response).to.equal(true)
+
+            const nonExistentClassHash = `0x${starkwareCrypto.pedersen(["random", "hash"])}`
+
+            // class hash is not declared
+            await expectInvokeError(
```
cfal, line 67: nbd, but there's `expectInvokeErrorMsg` to make sure the error being raised is the expected one.. guessing there's a error code for failure to upgrade

contracts/test/upgradeable.test.ts
```diff
@@ -0,0 +1,73 @@
...
+            await expectCallError(upgradeableContract.call('foo'))
+
+            const afterUpgradeContract = nonUpgradeableFactory.getContractAt(upgradeableContract.address)
+            expect((await afterUpgradeContract.call('bar')).response).to.equal(true)
+        })
+
+        it('reverts if new implementation class hash does not exist', async () => {
+            expect((await upgradeableContract.call('foo')).response).to.equal(true)
+
+            const nonExistentClassHash = `0x${starkwareCrypto.pedersen(["random", "hash"])}`
```
cfal, line 64: you should be able to use `require('starknet').ec.starkCurve.pedersen` instead of requiring a new dep

cfal (changes requested):

PR #257, opened 2023-05-09
BCI-1289: Add Ownable test helper

calvwang9:
https://smartcontract-it.atlassian.net/browse/BCI-1289

PR implements the `should_behave_like_ownable_contract` test utility function for use in Cairo tests, based off the hardhat version.

TODOs:
- Both `SimpleReadAccessController` and `SimpleWriteAccessController` are currently both a library as well as a standalone deployable contract, missing the Ownable interface implementation despite relying on Ownable for some functions. This is a bug for the standalone use case, but is probably better addressed through a later refactor in a separate PR to separate library and deployable versions.


augustbleeds:
I think it's definitely an oversight (mistake) that they do not have the ownable view and external methods. 

Can you add them? So that your tests are complete? Because I don't immediately see the value in having two versions of the contract. It seems like code duplication to achieve the same effect. It's akin to having an inherited contract in solidity where you can still deploy the parent contracts.

So just adding 

```
   ///
    /// Ownership
    ///

    #[view]
    fn owner() -> ContractAddress {
        Ownable::owner()
    }

    #[view]
    fn proposed_owner() -> ContractAddress {
        Ownable::proposed_owner()
    }

    #[external]
    fn transfer_ownership(new_owner: ContractAddress) {
        Ownable::transfer_ownership(new_owner)
    }

    #[external]
    fn accept_ownership() {
        Ownable::accept_ownership()
    }

    #[external]
    fn renounce_ownership() {
        Ownable::renounce_ownership()
    }
```

to both access controllers

calvwang9:
> Because I don't immediately see the value in having two versions of the contract. It seems like code duplication to achieve the same effect. It's akin to having an inherited contract in solidity where you can still deploy the parent contracts.

@augustbleeds I agree that this pattern makes sense in Solidity, but we don't really see the same inheritance patterns in Cairo. I was thinking that splitting the lib and deployable contracts allows for better separation of responsibilities and removes dependencies between components (e.g. no need to be coupled with Ownable), which generally leads to less overhead and risks when using these contracts. Keen to hear thoughts

contracts/src/libraries/simple_write_access_controller.cairo
```diff
@@ -100,9 +100,39 @@ mod SimpleWriteAccessController {
         }
     }
 
+    ///
+    /// Ownership
```
cfal: nit: Ownable?

contracts/src/tests/test_ownable.cairo
```diff
@@ -121,3 +122,45 @@ fn test_renounce_ownership_panics_if_not_owner() {
...
+#[abi]
+trait IOwnable {
+    fn owner() -> ContractAddress;
+    fn proposed_owner() -> ContractAddress;
+    fn transfer_ownership(new_owner: ContractAddress);
+    fn accept_ownership();
+    fn renounce_ownership();
+}
+
+fn should_behave_like_ownable_contract(contract_addr: ContractAddress, owner: ContractAddress) {
```
cfal: how about should_implement_ownable?

cfal (approved):

contracts/src/libraries/simple_write_access_controller.cairo
calvwang9: Was for consistency with other contracts - I prefer `Ownable` too but will change it across everything then

contracts/src/tests/test_ownable.cairo
calvwang9: I do like that better

calvwang9:
Currently pending https://github.com/smartcontractkit/chainlink-starknet/pull/260 to be reviewed and merged or rejected

archseer:
LGTM but needs another merge to resolve the conflicts

contracts/src/multisig.cairo
```diff
@@ -1,21 +1,17 @@
...
     let len = a.len();
     _assert_unique_values_loop(a, len, 0_usize, 1_usize);
 }
 
-fn _assert_unique_values_loop<T,
-impl TCopy: Copy<T>,
-impl TDrop: Drop<T>,
-impl TPartialEq: PartialEq<T>,
+fn _assert_unique_values_loop<
+    T, impl TCopy: Copy<T>, impl TDrop: Drop<T>, impl TPartialEq: PartialEq<T>,
```
archseer: I think you need to run `cairo-format src/**/*.cairo`, the linting fails here otherwise

contracts/src/tests/test_ownable.cairo
```diff
@@ -121,3 +122,45 @@ fn test_renounce_ownership_panics_if_not_owner() {
...
+//
+
+#[abi]
+trait IOwnable {
+    fn owner() -> ContractAddress;
+    fn proposed_owner() -> ContractAddress;
+    fn transfer_ownership(new_owner: ContractAddress);
+    fn accept_ownership();
+    fn renounce_ownership();
+}
```
archseer, line 136: We could declare this inside the ownable module so it can be imported elsewhere?

contracts/src/multisig.cairo
calvwang9: my local cairo-format doesnt seem to pick this up 🤔 
```
local setup:
cairo-lang-formatter 1.0.0-rc0
cairo: 1.0.0-rc0
scarb 0.2.0-alpha.1
```

I'll fix this manually for now

contracts/src/tests/test_ownable.cairo
calvwang9: I copied it over but it doesnt seem to work in this test file when importing from Ownable 🤔

cfal: does `use chainlink::libraries::ownable::IOwnableDispatcher` work?

calvwang9: tried that but still doesnt work (methods not found on type IOwnableDispatcher)

archseer (approved):

PR #258, opened 2023-05-09
BCI-1224.rewrite example contracts

augustbleeds:
This PR refactors both example folders.

Because all our cairo contracts are compiled on cairo-rc-0, they are not deployable to starknet goerli alpha. Therefore, the aggregator consumer examples are only tested on devnet.

I had to rework the functionality of the examples/ for the price consumer with sequencer check because it didn't make sense in the last revision. So for now, I eliminated the tests since they were testing defunct logic. (i.e. it takes longer than 60 seconds to deploy the contracts so the reports will always be stale)


package.json
```diff
@@ -6,7 +6,8 @@
   "workspaces": [
     "contracts",
     "examples/contracts/*",
-    "packages-ts/*"
+    "packages-ts/*",
+    "examples/new_contracts/*"
```
archseer: I'd create an initial commit that removes the old contracts then place the new ones under `examples/contracts`. That way the scripts can at least be reviewed as a diff and less adjustments will be needed on the CI

archseer (changes requested):
The format check can now be re-enabled: https://github.com/smartcontractkit/chainlink-starknet/blob/cfd1e36e2e566f60af5fa5a2466b02f0ee15b8a8/Makefile#L149-L151

And if the path is the same the CI should now be running the `aggregator-consumer` example as a test: https://github.com/smartcontractkit/chainlink-starknet/blob/cfd1e36e2e566f60af5fa5a2466b02f0ee15b8a8/Makefile#L224-L225

archseer:
`yarn compile` should be replaced here similarly to `build-contracts`:

https://github.com/smartcontractkit/chainlink-starknet/blob/cfd1e36e2e566f60af5fa5a2466b02f0ee15b8a8/Makefile#L91-L103


https://github.com/smartcontractkit/chainlink-starknet/blob/cfd1e36e2e566f60af5fa5a2466b02f0ee15b8a8/Makefile#L261-L263

I'm not sure what the difference between `build-contracts` and `build-ts-contracts` is. I guess the latter one still needs `yarn compile:solidity`

EDIT: ah ok, this is partially tackled in https://github.com/smartcontractkit/chainlink-starknet/pull/240

package.json
augustbleeds: sounds good! I initially was following the pattern we did for the contracts but you're right in that they should be compared against each other since we're not holding on to the old examples

examples/contracts/aggregator-consumer/src/ocr2.cairo
```diff
@@ -0,0 +1,3 @@
+mod mocks;
```
archseer: Seems unused?

contracts/src/ocr2/mocks/mock_aggregator.cairo
```diff
@@ -25,6 +25,9 @@ mod MockAggregator {
         answer: u128, block_num: u64, observation_timestamp: u64, transmission_timestamp: u64
     ) {
         let new_round_id = _latest_aggregator_round_id::read() + 1_u128;
+
+        // update both storage variables
+        _latest_aggregator_round_id::write(new_round_id);
```
archseer: @cfal pointed this out on slack, the bug was already fixed on L45

examples/contracts/aggregator-consumer/test/mock/Examples_test.test.ts
cfal, line 1: did we want to delete the example tests?

augustbleeds: Yeah, I left a comment in the description on why I had to remove it.

archseer:
I think this needs a merge/rebase? The diff includes changes that are already on develop

PR #260, opened 2023-05-11
Refactor access controller to separate lib and deployable contracts

calvwang9:
Based on the discussion here: https://chainlink-core.slack.com/archives/C04Q61M2MFT/p1683631059251589

Refactoring the access controller(s) to separate library contracts from deployable contracts. This leads to a bit more code across the board, but helps with the separation of responsibilities and reduces implicit behaviour (e.g. access controller initializer used to initialize Ownable as well, but now a contract should initialize both separately).

Test coverage isn't the best, but will aim to address these in later PRs.

Some other changes:
- `has_access` changed from external to view across multiple contracts
- removed `check_access` across multiple contracts as this is a 'modifier', not intended to be exposed as a standalone function
- minor internal logic change: merged Read and Write Access Controllers libraries in favor of a simpler single Access Controller lib. This means that direct off-chain calls (caller_addr = zero) always have access, though this shouldn't have any material impact. I've checked all the use cases across our contracts and there is no impact on functionality.

Before:
image

After:
image


contracts/src/access_control/simple_write_access_controller.cairo
```diff
@@ -0,0 +1,83 @@
...
+
+    #[constructor]
+    fn constructor(owner_address: ContractAddress) {
+        Ownable::initializer(owner_address);
+        AccessController::initializer();
+    }
+
+    #[view]
+    fn has_access(user: ContractAddress, data: Array<felt252>) -> bool {
+        AccessController::has_access(user, data)
```
cfal: this behavior is different than before. the write access controller previously didn't have the `caller.is_zero()` check. does it matter? @archseer

contracts/src/ocr2/aggregator.cairo
```diff
@@ -226,7 +225,7 @@ mod Aggregator {
 
     fn require_access() {
         let caller = starknet::info::get_caller_address();
-        SimpleReadAccessController::check_access(caller);
+        AccessController::check_access(caller);
```
cfal: probably can't be helped with multi inheritance and cairo.. but it does seems strange that we're using all the functionality here of SimpleWriteAccessController and have the exact same set of functions 😅 but we don't call it that directly, and a reader would have to go through the contract and the functions to figure out what access control functionality a contract has.

it probably wasn't better before, although i guess when you saw an instance of `SimpleWriteAccessController::function_call`, it hints that the entire `SimpleWriteAccessController` interface is implemented

cfal (commented):
seems ok to me, my only worry being if there are any functional changes or bugs introduced... but that's probably what tests and audits are for 😄 

i'll defer to @archseer on the final call

contracts/src/access_control/simple_write_access_controller.cairo
calvwang9: yep i called this out in the pr description - essentially direct off-chain calls always have access. Our current use cases for this function are not impacted, but this shouldn't have any material impact overall

contracts/src/ocr2/aggregator.cairo
calvwang9: The goal of merging read and write access controllers was to unify them into only one interface that covers everything access control related. I'm not sure there would ever be a case where an inheriting contract would only use read functions (can't read anything if you don't write), and the previous implementation of the read access controller only added the zero address check which I've moved into the underlying `has_access` function itself, as above.

calvwang9: On that note, would we ever need the current deployable `SimpleReadAccessController` contract? 🤔

contracts/src/access_control/simple_write_access_controller.cairo
archseer: In solidity:

```
 * @dev SimpleReadAccessController is not suitable for access controlling writes
 * since it grants any externally owned account access! See
 * SimpleWriteAccessController for that.
```

https://github.com/smartcontractkit/libocr/blob/master/contract2/dev/SimpleReadAccessController.sol#L13-L15

ReadAccessController is only used as a library, and WriteAccessController is used as a contract.

Let's do this:
- Add a `has_read_access()` function that does the `caller.is_zero() || has_access()`
- Use this inside aggregator/aggregator_proxy etc
- Rename `WriteAccessController` to `AccessController` (only expose has_access there)

This way we have a generic access controller that can be deployed, but there's also a specialization of the check in the library for those contracts that need it

It would also cut down on the amount of the code

WDYT?

calvwang9: I like that approach 👍 are we also deleting `SimpleReadAccessController` then?

archseer: Ah yeah, we'd remove `SimpleReadAccessController` 👍🏻

archseer (approved):

contracts/src/libraries/access_control.cairo
```diff
@@ -0,0 +1,102 @@
...
+
+        // NOTICE: read access is granted to direct calls, to enable off-chain reads.
+        if user.is_zero() {
+            return true;
+        }
+
+        false
+    }
+
+    fn check_access(user: ContractAddress) {
```
augustbleeds, line 52: I don't think this is used in your current iteration of the contracts.

calvwang9: correct, see https://chainlink-core.slack.com/archives/C03BX6QP7LN/p1683871383784279?thread_ts=1683870275.888489&cid=C03BX6QP7LN

PR #263, opened 2023-05-11
Use eth address type

augustbleeds:


contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -22,10 +28,8 @@ mod SequencerUptimeFeed {
     use chainlink::ocr2::aggregator::Aggregator::TransmissionStorageAccess;
     use chainlink::libraries::upgradeable::Upgradeable;
 
-    const ETH_ADDRESS_BOUND: felt252 = 0x10000000000000000000000000000000000000000; // 2**160
-
     struct Storage {
-        // l1 sender is an ethereum address
+        // l1 sender is an starknet validator ethereum address
         _l1_sender: felt252,
```
archseer, line 33: Are we not able to use `EthAddress` here? I'm guessing it doesn't implement storage?

contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -118,19 +122,19 @@ mod SequencerUptimeFeed {
     }
 
     #[external]
-    fn set_l1_sender(address: felt252) {
+    fn set_l1_sender(address: EthAddress) {
```
cfal, line 125: - should the `l1_sender` view function underneath this use EthAddress as well?
- should the `l1_handler` function use EthAddress?

examples/contracts/aggregator-consumer/package.json
```diff
@@ -13,7 +13,7 @@
   "license": "MIT",
   "devDependencies": {
     "@nomiclabs/hardhat-ethers": "^2.1.0",
-    "@shardlabs/starknet-hardhat-plugin": "git+https://github.com/cfal/starknet-hardhat-plugin#a68f0dc7cf535440bdeacfed5d5d752c3e87e6a3",
+    "@shardlabs/starknet-hardhat-plugin": "0.8.0-alpha.2",
```
cfal: this was upgraded on cairo-1.0

contracts/src/emergency/sequencer_uptime_feed.cairo
augustbleeds: l1_handler isn't able to use EthAddress. You get a compiler error that the first argument must be felt252. 

for l1_sender view function it should use it, thanks for reminding me

examples/contracts/aggregator-consumer/package.json
augustbleeds: Thanks, I think it wasn't when I created the PR. But I see I forgot the carat symbol too

augustbleeds: I guess I'll do a rebase

contracts/src/emergency/sequencer_uptime_feed.cairo
augustbleeds: Yup, it doesn't. If you want, I can implement storage for it, but I didn't think it was necessary because it'd bloat the contract size when we can just do a felt252 conversion

archseer: It would probably result in the same bytecode: a cast followed by a store

contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -1,5 +1,10 @@
 #[contract]
 mod SequencerUptimeFeed {
+    use starknet::EthAddress;
+    use starknet::EthAddressSerde;
+    use starknet::EthAddressIntoFelt252;
+    use starknet::Felt252TryIntoEthAddress;
+    use starknet::EthAddressZeroable;
```
archseer, line 7: I think trait imports are no longer required but we can test that out afterwards

PR #265, opened 2023-05-15
Fix aggregator billing bug

calvwang9:
Fixing a small bug in the Aggregator which was using an outdated interface for Access Controller after the refactor from https://github.com/smartcontractkit/chainlink-starknet/pull/260. Also expanded test coverage to cover the changes

### DIFF
--- a/contracts/src/emergency/sequencer_uptime_feed.cairo
+++ b/contracts/src/emergency/sequencer_uptime_feed.cairo
@@ -216,11 +216,6 @@ mod SequencerUptimeFeed {
         AccessControl::has_access(user, data)
     }
 
-    #[view]
-    fn check_access(user: ContractAddress) {
-        AccessControl::check_access(user)
-    }
-
     #[external]
     fn add_access(user: ContractAddress) {
         Ownable::assert_only_owner();
--- a/contracts/src/ocr2/aggregator.cairo
+++ b/contracts/src/ocr2/aggregator.cairo
@@ -88,9 +88,9 @@ mod Aggregator {
     use starknet::storage_address_from_base_and_offset;
     use starknet::class_hash::ClassHash;
 
+    use chainlink::utils::split_felt;
     use chainlink::libraries::ownable::Ownable;
     use chainlink::libraries::access_control::AccessControl;
-    use chainlink::utils::split_felt;
     use chainlink::libraries::upgradeable::Upgradeable;
 
     // NOTE: remove duplication once we can directly use the trait
@@ -99,6 +99,16 @@ mod Aggregator {
         fn balance_of(account: ContractAddress) -> u256;
         fn transfer(recipient: ContractAddress, amount: u256) -> bool;
     // fn transfer_from(sender: ContractAddress, recipient: ContractAddress, amount: u256) -> bool;
+    }
+
+    // NOTE: remove duplication once we can directly use the trait
+    #[abi]
+    trait IAccessController {
+        fn has_access(user: ContractAddress, data: Array<felt252>) -> bool;
+        fn add_access(user: ContractAddress);
+        fn remove_access(user: ContractAddress);
+        fn enable_access_check();
+        fn disable_access_check();
     }
 
     const GIGA: u128 = 1000000000_u128;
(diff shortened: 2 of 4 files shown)


contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -216,11 +216,6 @@ mod SequencerUptimeFeed {
         AccessControl::has_access(user, data)
     }
 
-    #[view]
-    fn check_access(user: ContractAddress) {
```
cfal, line 220: - i realize that this is a simple function that's doing an assert on `has_access`, but is there a reason we decided to remove this again? 
- if we're removing this, do we still need the `check_access` in access_control.cairo?
- can you grep for other places with references to `check_access`? eg. and the sequencer uptime feed test should be updated (you can test it by starting up starknet-devnet using the script in ops/scripts and running `yarn test --grep SequencerUptimeFeed` from contracts)

contracts/src/ocr2/aggregator.cairo
```diff
@@ -940,7 +945,9 @@ mod Aggregator {
 
         let access_controller = _billing_access_controller::read();
         let access_controller = IAccessControllerDispatcher { contract_address: access_controller };
-        access_controller.check_access(caller);
+        assert(
+            access_controller.has_access(caller, ArrayTrait::new()), 'caller does not have access'
```
cfal, line 949: ..also, it does seem like a reasonably harlmess helper function if we are going to be inlining it instead

contracts/src/tests/test_aggregator.cairo
```diff
@@ -54,16 +57,39 @@ fn test_pow_2_0() {
...
+    calldata.append(acc1.into()); // owner = acc1;
+    let (billingAccessControllerAddr, _) = deploy_syscall(
+        AccessController::TEST_CLASS_HASH.try_into().unwrap(), 0, calldata.span(), false
+    ).unwrap();
+    let billingAccessController = IAccessControllerDispatcher {
+        contract_address: billingAccessControllerAddr
+    };
+
+    // return accounts and billing access controller
+    (acc1, acc2, billingAccessControllerAddr, billingAccessController)
```
cfal, line 86: i think you can do `billingAccessController.contract_address` and avoid having to return it directly. or just only create in `test_set_billing_config_as_acc_with_access`

contracts/src/emergency/sequencer_uptime_feed.cairo
calvwang9: 1. this was removed because rather than relying on an external call to a view function to revert, I figured it would be better practice and more robust to read data from external contacts and handle the revert internally
2. access_control.cairo is the library, intended to be imported and used as such, e.g. DAO contract inherits access control functionality and uses it to assign management roles. access_contoller.cairo (being interacted with here) is the deployable contract version and doesn't expose check_access as an external function, for the reason above.
3. yep this seems to be the only place we interact with a deployed access controller, but will double check. haven't been updating the hardgat tests, but I do have another pr in progress to reimplement in Cairo if we want to replace them moving forwards

cfal: thanks for explaining!

1. hmm ok, i guess this is common solidity practice? i'm unfamiliar, what makes it more robust?
2. it seems strange that the access control library has the `has_read_access` and `check_read_access` function then, with the only difference being the offchain reads. that seems pretty specific to be in a maintain-an-acl-list-only library
3. thanks! would it make sense to remove/update the `check_access` sequencer uptime feed test now so the tests can all pass? i'm working on getting the CI working and it'd be nice if i could see those beautiful green checkmarks :-) hopefully it will be easier to catch these failures without manual runs soon.

cfal (commented):
approving due to the time crunch and the rest of the comments are test related.. not sure we have time for a longer discussion here? 😅  @archseer

cfal (approved):

contracts/src/emergency/sequencer_uptime_feed.cairo
archseer: 1. I did think it's a bit unclear if a library function reverts or not because the name doesn't specify if a bool is returned or if the function panics. We could make it clearer by prefixing with `assert_`, eg` assert_has_access` but for now I think the change is OK 👍🏻

PR #267, opened 2023-05-16
cairo 1.0 CI: remaining fixes, update cairo-lang and starknet-devnet

cfal:
BCI-1323


ops/devnet/environment.go
```diff
@@ -71,7 +71,7 @@ func defaultProps() map[string]any {
 		"starknet-dev": map[string]any{
 			"image": map[string]any{
 				"image":   "shardlabs/starknet-devnet",
-				"version": "v0.3.5",
```
archseer, line 74: I think this actually isn't used, it uses that helm chart template instead \cc @smickovskid

archseer (approved):
Great work, LGTM!

PR #268, opened 2023-05-16
Add access control and sequencer uptime feed tests

calvwang9:
https://smartcontract-it.atlassian.net/browse/BCI-1288

Copying Sequencer Uptime Feed tests from hardhat into cairo-test, but original hardhat tests were left alone, with the exception of the broken test related to an old check_access function.

Also adding access control tests to be used across all contracts inheriting the access control lib


contracts/test/emergency/sequencer_uptime_feed.test.ts
```diff
@@ -85,15 +85,6 @@ describe('SequencerUptimeFeed', function () {
...
 
-    it('should error on `check_access` without access', async function () {
-      await uptimeFeedContract.call('check_access', { user: user })
-
-      await expectInvokeError(
-        owner.invoke(uptimeFeedContract, 'check_access', { user: user + 1 }),
-        'user does not have access',
-      )
-    })
-
```
calvwang9, line 96: check_access no longer exists, removing old test

cfal: feel free to remove the relevant hardhat tests as well (and same for ownable) - i had originally commented on keeping them on the first PR to make sure we have the code for updating from 0.x -> 1.0 in our git tree incase we ever wanted to refer to it

contracts/src/tests/test_access_controller.cairo
```diff
@@ -19,3 +28,58 @@ fn test_upgrade_not_owner() {
...
+    fn remove_access(user: ContractAddress);
+    fn enable_access_check();
+    fn disable_access_check();
+}
+
+fn should_implement_access_control(contract_addr: ContractAddress, owner: ContractAddress) {
+    let contract = IAccessControllerDispatcher { contract_address: contract_addr };
+    let acc2: ContractAddress = contract_address_const::<2222987765>();
+
+    set_contract_address(owner); // required to call contract as owner
```
augustbleeds, line 65: can you elaborate what this is used for? I thought it was so you can set the value of starknet::info::get_contract_address() but its use seems to be very different.

augustbleeds: Nvm I found out why[ here](https://discord.com/channels/793094838509764618/1065544063245365288/1105507121749049404): 

> you can view your test as a contract that calls another contract. If you want the get_caller_address() in your remote contract to be mocked, then you will need to use set_contract_address() in the test

contracts/src/tests/test_sequencer_uptime_feed.cairo
```diff
@@ -0,0 +1,107 @@
...
+
+fn setup() -> (ContractAddress, ContractAddress, ISequencerUptimeFeedDispatcher) {
+    let account: ContractAddress = contract_address_const::<777>();
+    set_caller_address(account);
+
+    // Deploy seqeuencer uptime feed
+    let mut calldata = ArrayTrait::new();
+    calldata.append(0); // initial status
+    calldata.append(account.into()); // owner
+    let (sequencerFeedAddr, _) = deploy_syscall(
```
augustbleeds, line 36: I don't think there's a necessary requirement to deploy the contract. I think it's only necessary when we're expecting to do contract to contract interactions (calling one contract that calls another contract) or need to manage multiple contracts of the same type. In my opinion it adds some unnecessary complexity when not necessary (as opposed to just initializing the contract as a singleton). Just a nit though.

augustbleeds (dismissed):

contracts/src/tests/test_sequencer_uptime_feed.cairo
calvwang9: yep i try avoid deployment where possible, but in this case the access control and ownable tests require a deployed contract to interact with a contract through IOwnableDispatcher for example

contracts/test/emergency/sequencer_uptime_feed.test.ts
calvwang9: removed both, thanks

contracts/src/tests/test_sequencer_uptime_feed.cairo
cfal: yeah, it would be nice if there was a way to interact through an interface or templating eg. if you could write a generic function that calls `<any class>::has_access`, but seems like deploying and using dispatcher is the only way to do so

cfal (changes requested):
- you need to remove other function calls to `shouldBehaveLikeOwnableContract` and references to ownable.ts - take a look at the failing typescript tests CI, it should pass now if all is well
- you might want to ping the security channel and maybe fix up your .gitconfig - noticed that all your commits are `Unverified`

cfal (approved):

PR #270, opened 2023-05-18
use caigotypes instead of junotypes

cfal:
BCI-1329

### DIFF
--- a/relayer/pkg/starknet/utils.go
+++ b/relayer/pkg/starknet/utils.go
@@ -6,14 +6,16 @@ import (
 	"fmt"
 	"math/big"
 
-	junotypes "github.com/NethermindEth/juno/pkg/types"
 	caigotypes "github.com/smartcontractkit/caigo/types"
 
 	"github.com/pkg/errors"
 	"golang.org/x/exp/constraints"
 )
 
-const chunkSize = 31
+const (
+	FeltLength = 32
+	chunkSize  = 31
+)
 
 // padd bytes to specific length
 func PadBytes(a []byte, length int) []byte {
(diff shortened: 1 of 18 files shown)


cfal:
rebased ontop of #267 for CI tests

relayer/pkg/starknet/utils.go
```diff
@@ -28,6 +28,10 @@ func PadBytes(a []byte, length int) []byte {
 	return a
 }
 
+func FeltToBytes(felt *caigotypes.Felt) []byte {
+	return PadBytes(felt.Bytes(), FeltLength)
+}
```
archseer: Should we just fix this in caigo? And make `felt.Bytes()` return the right format

relayer/pkg/chainlink/keys/utils.go
```diff
@@ -18,17 +17,25 @@ var (
 )
 
 // PubKeyToContract implements the pubkey to deployed account given contract hash + salt
-func PubKeyToAccount(pubkey PublicKey, classHash, salt *big.Int) []byte {
-	hash := pedersen.ArrayDigest(
+func PubKeyToAccount(pubkey PublicKey, classHash, salt *big.Int) ([]byte, error) {
```
archseer: Is this still used? We no longer derive the account address

cfal: looks like it's still used for deriving the default wallet address for testing: https://github.com/smartcontractkit/chainlink-starknet/blob/baa6a5c2cbdfb74f7165b995778caaa3058f5561/integration-tests/common/test_common.go#L47

archseer: Oh that might be a bug? Since we aren't deploying with a static salt anymore, it might not match

cfal: from the link in https://github.com/smartcontractkit/chainlink-starknet/blob/f6324ce41bf6efc6fa3d91015e59641b2ff71bea/relayer/pkg/chainlink/keys/test_helpers.go - looks like it's because devnet hardcodes salt and class hash: https://github.com/0xSpaceShard/starknet-devnet/blob/master/starknet_devnet/account.py#L34-L35

which perhaps is reasonable since devnet provides ready-made accounts on startup. moved it into test_common and made it devnet specific

archseer (approved):

PR #272, opened 2023-05-18
BCI-1327: Remove Signed Integer Arithmetic

augustbleeds:
We are not storing our own representations of signed integers on-chain, therefore we can get rid of the unnecessary logic that is getting or posting on-chain data as if it were signed.

I added errors in two places (you can search for "starknet does not support" to find them)
* if [BuildReport() ](https://github.com/smartcontractkit/chainlink-starknet/pull/272/files#diff-d245ce16d3e10e9f866422946d06fbebde28ed4c8aee53b4902bd7679ebd52acR41)is called with a negative value 
* in [EncodeFromFelt() ](https://github.com/smartcontractkit/chainlink-starknet/pull/272/files#diff-3d87d94a54ca53465ba31c06c57809e60cbeb121bac5fe8fa13bd57022ba2a3dR74)when the onchain config is being built
For both cases though I wonder if we should not error though and if a submitted negative value should just be converted to positive representation because then an adversarial node could submit negative values to cause other nodes to error/crash.

It goes without saying that when we are reading from on-chain there is no need to error

This PR is *not* getting rid of the big.Int data type however, because it is the most natural type to use for arbitrarily large unsigned numbers (caigotypes and junotypes) both use it behind the scenes. In addition, it is required by lib ocr. We just make sure that the big.Int is non-negative when necessary.

Please see in-line PR comments for more details

Open Question:
[] should we wrap around positive big.Int values larger than the maximum felt size (take modulo)? I see that we don't do this already so it is theoretically possible for someone to report a value greater than felt252 and then the report will error when it the transaction fails


relayer/pkg/chainlink/ocr2/medianreport/onchainconfig.go
```diff
@@ -32,40 +31,30 @@ func (codec OnchainConfigCodec) DecodeToFelts(b []byte) ([]*big.Int, error) {
 		return []*big.Int{}, fmt.Errorf("unexpected length of OnchainConfig, expected %v, got %v", length, len(b))
 	}
 
-	configVersion, err := bigbigendian.DeserializeSigned(byteWidth, b[:32])
-	if err != nil {
-		return []*big.Int{}, fmt.Errorf("unable to decode version: %s", err)
-	}
+	// convert from bytes to *big.Int
+	configVersion := new(big.Int).SetBytes(b[:32])
```
augustbleeds, line 36: We can get rid of the DeserializeSigned usage because we assume all reads from on-chain are unsigned

relayer/pkg/chainlink/ocr2/medianreport/onchainconfig.go
```diff
@@ -87,24 +76,10 @@ func (codec OnchainConfigCodec) EncodeFromFelt(version, min, max *big.Int) ([]by
 		return nil, fmt.Errorf("unexpected version of OnchainConfig, expected %v, got %v", OnchainConfigVersion, version.Int64())
 	}
 
-	versionBytes, err := bigbigendian.SerializeSigned(byteWidth, version)
```
augustbleeds, line 90: no need to serialize signed data on-chain

relayer/pkg/starknet/utils.go
```diff
@@ -87,26 +87,14 @@ func DecodeFelts(felts []*big.Int) ([]byte, error) {
 	return data, nil
 }
 
-// SignedBigToFelt wraps negative values correctly into felt
-func SignedBigToFelt(num *big.Int) *big.Int {
-	return new(big.Int).Mod(num, caigotypes.MaxFelt.Big())
+func FeltToUnsignedBig(felt *caigotypes.Felt) (num *big.Int) {
```
augustbleeds: this replaces  FeltToSignedBig. We can remove this since it's a one-liner but I feel it makes it understandable that the results will be unsigned.

Note there's no possibility that FeltToUnsignedBig will return a negative big.Int

relayer/pkg/starknet/utils.go
```diff
@@ -87,26 +87,14 @@ func DecodeFelts(felts []*big.Int) ([]byte, error) {
 	return data, nil
 }
 
-// SignedBigToFelt wraps negative values correctly into felt
```
augustbleeds, line 92: No need for SignedBigToFelt anymore because there is no need to wrap around to a positive BigInt number that is a valid felt to represent negative numbers on-chain. In all places where this was used we can just use the big.Int value as is

relayer/pkg/starknet/utils.go
```diff
@@ -87,26 +87,14 @@ func DecodeFelts(felts []*big.Int) ([]byte, error) {
...
-func SignedBigToFelt(num *big.Int) *big.Int {
-	return new(big.Int).Mod(num, caigotypes.MaxFelt.Big())
+func FeltToUnsignedBig(felt *caigotypes.Felt) (num *big.Int) {
+	// will always return a non-negative value because it uses Bytes() behind the scenes
+	return felt.Big()
 }
 
-// FeltToSigned unwraps felt into negative values
-func FeltToSignedBig(felt *caigotypes.Felt) (num *big.Int) {
```
augustbleeds, line 98: No need for this since we're not storing negative values on-chain

relayer/pkg/starknet/utils.go
augustbleeds: Question: do we want to keep SignedBigToFelt logic around though because it will also wrap values that are too large? Or should we error that values are beyond the capacity of felt252

archseer:
Looks like `integration-tests` lint is failing, it's still using the removed method somewhere https://github.com/smartcontractkit/chainlink-starknet/actions/runs/5027808868/jobs/9017784383?pr=272

archseer:
^ bump on this! There's still a couple of test failures

augustbleeds:
> ^ bump on this! There's still a couple of test failures

Sorry, just got around to this again. Fixing rn

relayer/pkg/chainlink/ocr2/config_digester_test.go
```diff
@@ -29,7 +29,7 @@ var testConfig = types.ContractConfig{
 	F: 1,
 	OnchainConfig: []byte{
 		0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, // version
-		255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 246, // min (-1)
```
augustbleeds, line 32: only non negative encoding, so i change the min value to 1 to be more readable/user-friendly

relayer/pkg/chainlink/ocr2/events_test.go
```diff
@@ -51,7 +51,7 @@ var (
 		"0x1",
 		"0x3",
 		"0x1",
-		"0x800000000000010fffffffffffffffffffffffffffffffffffffffffffffff7",
```
augustbleeds, line 54: change to +10 instead of -10

relayer/pkg/starknet/utils.go
augustbleeds: @archseer

archseer: The wrapping was done because that's how negative values worked on 0.x, the upper part of the felt was treated as negative by certain functions. We should error here because this is not supported anymore and wrapping would produce invalid values (very large positive values)

augustbleeds: Sounds good. I'll add an error check if the value is greater than the maximum possible felt value

archseer (approved):

PR #274, opened 2023-05-19
contracts/ocr2/aggregator.cairo: fix item count when hashing

cfal:


cfal:
tests are failing due to some upstream dep change (prob starknet-devnet related), which is fixed by upgrading in #267

PR #281, opened 2023-05-24
keystore abstraction and adaption betweeen Loopp and Caigo

krehermann:
Adds Keystore abstraction to the transaction manager

Most of the work is translating from the LOOPp interface definition which is raw bytes to the `caigo` keystore definition which is big.Ints.

In order to achieve that translation, we introduce a Keystore Adapter  that implements both Keystore interfaces and leverages an intermediate representation of the signature to handle the roundtrip between the formats.


relayer/pkg/chainlink/keys/keystore.go
```diff
@@ -1,18 +1,54 @@
...
+}
+
+// caigoAdapter implements [KeystoreAdapter]
+type caigoAdapter struct {
+	looppKs loop.Keystore
+}
+
+// Sign implements the caigo Keystore Sign func.
+func (ca *caigoAdapter) Sign(senderAddress string, hash *big.Int) (*big.Int, *big.Int, error) {
+	raw, err := ca.looppKs.Sign(context.Background(), senderAddress, hash.Bytes())
```
aalu1418: noted in other PR - but having a context here would be useful so it doesn't block forever if something breaks :)

relayer/pkg/chainlink/keys/keystore.go
```diff
@@ -22,4 +58,117 @@ type NonceManager interface {
...
+const (
+	maxPointByteLen = 32 // stark curve max is 252 bits
+	signatureLen    = 2 * maxPointByteLen
+)
+
+// MemKeystore is an in-memory implementation of the LOOPp Keystore interface
+type MemKeystore struct {
+	mu   sync.Mutex
+	keys map[string]*big.Int
+}
```
aalu1418: just curious: is this for testing or something that will be a part of the codebase?

aalu1418 (commented):
✅ 

overall direction looks good to me!

relayer/pkg/chainlink/keys/keystore.go
```diff
@@ -22,4 +58,117 @@ type NonceManager interface {
...
+func signatureFromBytes(b []byte) (*signature, error) {
+	if len(b) != signatureLen {
+		return nil, fmt.Errorf("expected slice len %d got %d", signatureLen, len(b))
+	}
+	x := b[:maxPointByteLen]
+	y := b[maxPointByteLen:]
+
+	return &signature{
+		x: new(big.Int).SetBytes(x),
+		y: new(big.Int).SetBytes(y),
```
jmank88: Can these ints be negative? b/c `Bytes()` doesn't preserve the sign. I ran in to this with LOOPs and ended up using a protobuf https://github.com/smartcontractkit/chainlink-relay/blob/main/pkg/loop/internal/pb/big.go

relayer/pkg/chainlink/keys/keystore.go
krehermann: yes, this was test code. reorganized now

relayer/pkg/chainlink/keys/keystore.go
krehermann: done, thanks for catching

relayer/pkg/chainlink/keys/keystore.go
krehermann: that's a good question. 

yes, based on the elliptic curve, they can be negative.

the current OCR key signing for starknet is using big.Int.Bytes, and i largely copied that.

@aalu1418 what is the expectation regarding the sign of the points on the curve -- is the OCR implementation consciously choosing to ignore/drop it, or is that an oversight in the implementation? 

https://github.com/smartcontractkit/chainlink-starknet/blob/develop/relayer/pkg/chainlink/keys/ocr2key.go

relayer/go.sum
```diff
@@ -103,6 +103,16 @@ github.com/shirou/gopsutil v3.21.11+incompatible h1:+1+c1VGhc88SSonWP6foOcLhvnKl
...
+github.com/smartcontractkit/caigo v0.0.0-20230522212959-54ab5254f3c4/go.mod h1:2QuJdEouTWjh5BDy5o/vgGXQtR4Gz8yH1IYB5eT7u4M=
+github.com/smartcontractkit/caigo v0.0.0-20230522225550-5b0d833be5f6 h1:0a9HvcZ3n9R+jrci/ar0S/MCZxjr+MiSLQ/sMHgoJ+4=
+github.com/smartcontractkit/caigo v0.0.0-20230522225550-5b0d833be5f6/go.mod h1:2QuJdEouTWjh5BDy5o/vgGXQtR4Gz8yH1IYB5eT7u4M=
+github.com/smartcontractkit/caigo v0.0.0-20230523144837-b73c89a4d794 h1:CznBrxnPvN92UWKeHTbBbSE4vjWxAe3OkgI3y8m1pxU=
+github.com/smartcontractkit/caigo v0.0.0-20230523144837-b73c89a4d794/go.mod h1:2QuJdEouTWjh5BDy5o/vgGXQtR4Gz8yH1IYB5eT7u4M=
+github.com/smartcontractkit/caigo v0.0.0-20230524144942-06b177f4d7b8 h1:S3TTNUO/GXLuvJr+OO4BE7ruYSRCmpll2Cx+kD2kvHA=
+github.com/smartcontractkit/caigo v0.0.0-20230524144942-06b177f4d7b8/go.mod h1:2QuJdEouTWjh5BDy5o/vgGXQtR4Gz8yH1IYB5eT7u4M=
+github.com/smartcontractkit/caigo v0.0.0-20230524192636-ed64d702624d h1:DQKEZ1MehwEZeyPCsxTNyJfo22Vn2BpIvXqTUBdP5vY=
+github.com/smartcontractkit/caigo v0.0.0-20230524192636-ed64d702624d/go.mod h1:2QuJdEouTWjh5BDy5o/vgGXQtR4Gz8yH1IYB5eT7u4M=
```
aalu1418: nit: probs need a tidy here 😄

relayer/pkg/chainlink/keys/keystore.go
aalu1418: for the OCR key - i remember that being a conscious decision to only use the positive values

however, for the transmitter key/signatures I'm not sure - since we have previously passed the private key caigo handled the signs / encodings internally 🤔

krehermann: i looked at the official python implementation. in my reading, all the int values are positive. can one of you read the python and/or caigo give your assessment? cc @archseer 

i'm also going to ask research on the general topic of big.Int.Set/Bytes and elliptic curves.

the private key is positve     https://github.com/starkware-libs/starkex-for-spot-trading/blob/607f0b4ce507e1d95cd018d206a2797f6ba4aab4/src/starkware/crypto/starkware/crypto/signature/signature.py#L103

the msgHash is positive https://github.com/starkware-libs/starkex-for-spot-trading/blob/607f0b4ce507e1d95cd018d206a2797f6ba4aab4/src/starkware/crypto/starkware/crypto/signature/signature.py#L139


the signature components --
r is positive:         https://github.com/starkware-libs/starkex-for-spot-trading/blob/607f0b4ce507e1d95cd018d206a2797f6ba4aab4/src/starkware/crypto/starkware/crypto/signature/signature.py#L156

s is positive: 

https://github.com/starkware-libs/starkex-for-spot-trading/blob/607f0b4ce507e1d95cd018d206a2797f6ba4aab4/src/starkware/crypto/starkware/crypto/signature/signature.py#L170

https://github.com/starkware-libs/starkex-for-spot-trading/blob/607f0b4ce507e1d95cd018d206a2797f6ba4aab4/src/starkware/crypto/starkware/crypto/signature/signature.py#L111

https://github.com/starkware-libs/starkex-for-spot-trading/blob/607f0b4ce507e1d95cd018d206a2797f6ba4aab4/src/starkware/crypto/starkware/crypto/signature/math_utils.py#L52

jmank88: Would it make sense to sanity check and panic if negative?

krehermann: yes, good idea. i think the right place for that is the `Decode` implementation, which i've now added

integration-tests/go.mod
```diff
@@ -14,6 +14,13 @@ require (
...
 )
 
+require (
+	github.com/hashicorp/go-hclog v1.2.2 // indirect
+	github.com/hashicorp/go-plugin v1.4.9 // indirect
+	github.com/hashicorp/yamux v0.0.0-20200609203250-aecfd211c9ce // indirect
+	github.com/oklog/run v1.1.0 // indirect
+)
+
 require (
```
jmank88:
```suggestion
```

relayer/pkg/chainlink/keys/nonce_test.go
```diff
@@ -19,10 +19,10 @@ import (
 
 func newTestNonceManager(t *testing.T, chainID string, initNonce *big.Int) (keys.NonceManager, caigotypes.Hash, func()) {
 	// setup
-	ks := mocks.NewKeystore(t)
+	//ks := mocks.NewKeystore(t)
```
jmank88:
```suggestion
```

ops/go.mod
```diff
@@ -87,7 +93,11 @@ require (
 	github.com/robfig/cron/v3 v3.0.1 // indirect
 	github.com/russross/blackfriday v1.6.0 // indirect
 	github.com/shirou/gopsutil v3.21.11+incompatible // indirect
+<<<<<<< HEAD
 	github.com/smartcontractkit/chainlink-relay v0.1.7-0.20230525161650-dce1bc58b504 // indirect
+=======
+	github.com/smartcontractkit/chainlink-relay v0.1.7-0.20230525195208-f0926e42a26c // indirect
+>>>>>>> c18b8a9 (Add Starknet curve check when decoding raw signature; clean up go mods)
```
jmank88: unresolved conflicts

relayer/pkg/chainlink/keys/keystore.go
jmank88: Where is the sanity check? I see `(s *signature) bytes()` checking length, but not sign.

jmank88:
We'd like to wait to merge until after #284 (reverts breaking changes), so we can get this in to core ahead of those breaking changes.
Edit: Merged! Sorry for the conflicts :facepalm:

relayer/pkg/chainlink/keys/keystore_test.go
```diff
@@ -0,0 +1,110 @@
...
+			key = nil
+			err = fmt.Errorf("failed to generate test key in allotted time")
+			break GENERATEKEY
+		default:
+			key, err = caigo.Curve.GetRandomPrivateKey()
+			if err == nil {
+				break GENERATEKEY
+			}
+		}
+	}
```
jmank88: WDYT about trying to avoid the label somehow?
```suggestion
	generate := func() bool {
		select {
		case <-timer.C:
			key = nil
			err = fmt.Errorf("failed to generate test key in allotted time")
			return true
		default:
			key, err = caigo.Curve.GetRandomPrivateKey()
			if err == nil {
				return true
			}
		}
		return false
	}
	for !generate() { }
```

krehermann: sure.

what is the objection about the named loop?

integration-tests/go.mod
jmank88: There's still an odd separate `// indirect` block here.

relayer/pkg/chainlink/keys/keystore_test.go
jmank88: Just that they are generally more difficult to reason about and prone to errors in future edits. Or sometimes a code smell that there is too much nesting/indentation regardless.
I usually find alternative styles to be simpler anyways, except maybe in cases of tight double for loops?

relayer/pkg/chainlink/keys/keystore.go
```diff
@@ -1,18 +1,122 @@
...
+		x: x,
+		y: y,
+	}
+	return s.bytes()
+}
 
-type Keystore interface {
-	Get(id string) (Key, error)
+// TODO what is this supposed to return for starknet?
+func (lk *LooppKeystore) Accounts(ctx context.Context) ([]string, error) {
```
jmank88, line 118: Does starknet have a balance monitor? That's what this was supporting.

krehermann: @aalu1418?

the keystore interface that had been used in this repo doesn't need this func as far as i can tell and i didn't find anything like a balance monitor

aalu1418: we don't have one yet, but that's usually one of the first features NOPs request

relayer/pkg/chainlink/txm/txm_test.go
```diff
@@ -32,34 +33,37 @@ func TestIntegration_Txm(t *testing.T) {
...
 	for i := range accounts {
 		privKey, err := caigotypes.HexToBytes(accounts[i].PrivateKey)
 		require.NoError(t, err)
 
-		key := keys.Raw(privKey).Key()
-		assert.Equal(t, caigotypes.HexToHash(accounts[i].PublicKey), caigotypes.HexToHash(key.ID()))
-		localKeys[key.ID()] = key
-		localAccounts[key.ID()] = accounts[i].Address
+		senderAddress := caigotypes.HexToHash(accounts[i].PublicKey).String()
+		//assert.Equal(t, caigotypes.HexToHash(accounts[i].PublicKey), caigotypes.HexToHash(key.ID()))
```
jmank88: Drop or restore?

integration-tests/go.mod
krehermann: i thought it would have been fixed magically my one of my many go mod tidy incantations. surprised that it wasn't. should be fixed now

relayer/pkg/chainlink/txm/txm_test.go
krehermann: are you getting pinged to review this over and over? i've been battling the code all day and it's been WIP. i thought you would only get pinged if i manually request review.

going forward would you prefer i put it back in draft when i'm pushing a lot of changes that i don't expect to be final?

integration-tests/go.mod
jmank88: yeah the tooling doesn't undo these, and it occasionally introduces them for some reason :confused:

relayer/pkg/chainlink/txm/txm_test.go
jmank88: NBD I was just trying to stay ahead since we are blocked on the error signature change due to an odd testing module dependency between this repo and solana.
Draft could still be helpful though :shrug:

relayer/pkg/chainlink/keys/keystore.go
jmank88: bump ^

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -48,25 +49,25 @@ type starktxm struct {
 	done    sync.WaitGroup
 	stop    chan struct{}
 	queue   chan Tx
-	ks      keys.Keystore
+	ks      keys.KeystoreAdapter //caigo.Keystore //*keys.CaigoAdapter //loop.Keystore
```
jmank88: What does this comment mean?

krehermann: it's WIP garbage that needs to be deleted. thanks for finding it

relayer/pkg/chainlink/keys/keystore.go
```diff
@@ -22,4 +126,57 @@ type NonceManager interface {
...
+}
+
+const (
+	maxPointByteLen = 32 // stark curve max is 252 bits
+	signatureLen    = 2 * maxPointByteLen
+)
+
+// signature is an intermediate representation for translating between a raw-bytes signature and a caigo
+// signature comprised of big.Ints
+type signature struct {
```
jmank88: Looking forward towards build separation (or even just our current blocking test entanglement), I think this type will be problematic because it can't be shared from both sides. We would need to have it represented in `chainlink-relay` to do that. Is it important for this to be a custom encoding? Or could we reuse the existing protobuf that already supports the reporting interfaces? https://github.com/smartcontractkit/chainlink-relay/blob/main/pkg/loop/internal/pb/relayer.proto#L306

relayer/pkg/chainlink/keys/keystore.go
krehermann: Oh right, i meant to reply to this.

It turns out that this is not needed. it's not possible to check that they are negative -- the SetBytes assumes an absolute value, so you would need a different en/decoding to represent a negative number and therefore you can't even check for negatives in a meaningful way here.

the same idea applies to the bytes() func: since it's using big.Int.Bytes(), it is not possible to encode negative values.

the only place a negative could be represented is in line 110-112. but there again the value is coming from the caigo.Sign, which returns positive values. 

we could add a panic there i suppose. wdyt?

i've written tests for the round trip in keystore and signature. do you see anything missing in them?

relayer/pkg/chainlink/keys/keystore.go
krehermann: this is private and isn't shared, unless i'm misunderstanding you.

it's only needed in this repo and more specifically, only for the txm.

it's helper layer that is used in a specific implementation of the keystore adaption logic. in principle the keystore adapter is an interface and another implementation could be used, but there are a lot of constraints that have to be met, and this is private way to enforce those constraints for a specific implementation.

the protobuf definition doesn't help here because we need an encoding that turns bytes into multiple big.Ints. we would change the problem into needing an encoding for multiple protobuf instances

relayer/pkg/chainlink/keys/keystore.go
jmank88: If `(s *signature) bytes()` is called with a signature that has negative values, it will silently drop that information. IMHO this is where the panic should occur: at the narrowest scope without making assumptions about anything else.

relayer/pkg/chainlink/keys/keystore.go
krehermann: more generally, we need a good story for this kind of pattern.

there is no constraint on new chains that they have to accept bytes as a signature and therefore there will be a need to adapt the loop.keystore interface to chain-specific logic

jmank88: > this is private and isn't shared, unless i'm misunderstanding you.

We need a way to understand the encoding on both ends. It would be better not to have separate redundant/duplicate code that needs to be kept in sync.

> we would change the problem into needing an encoding for multiple protobuf instances

Yes, but that is a trivial problem.

> more generally, we need a good story for this kind of pattern.
>
> there is no constraint on new chains that they have to accept bytes as a signature and therefore there will be a need to adapt the loop.keystore interface to chain-specific logic

This is why I suggested using a dynamic protobuf message instead of raw `[]byte`.

jmank88: I think https://protobuf.dev/programming-guides/proto3/#oneof is probably the idiomatic way to support different types of signatures here :shrug:

krehermann: >We need a way to understand the encoding on both ends. It would be better not to have separate redundant/duplicate code that needs to be kept in sync.

I don't think the explicit encoding needs to be exposed, rather we need a consistent use of the implemented interfaces.

the encoding is hidden behind the interfaces. if we use consistent implementations on both sides then only the implementors/owners of the chain need to think about this problem.

i dislike changing the protobuf defintion because it gets into a gray area of leaking the implementation details about a specific chains.

if a 3rd chain comes along with some other signature, we would need to update the proto rather to support it. that is a lot of overhead and coupling. instead we can leave it to the implementors to encode and decode the bytes as they need for there use case

it feels like the same kind of problem as marshaling and unmarshaling json or http -- as an implementor you need to know how to interpret the bytes, but the protocol that transports them doesn't care

jmank88: > if we use consistent implementations on both sides then only the implementors/owners of the chain need to think about this problem.

The implementation must exist in chainlink-relay, or be duplicated.

> i dislike changing the protobuf defintion because it gets into a gray area of leaking the implementation details about a specific chains.

The details are already leaked. This is just a way of standardizing them.

> if a 3rd chain comes along with some other signature, we would need to update the proto rather to support it. that is a lot of overhead and coupling. instead we can leave it to the implementors to encode and decode the bytes as they need for there use case

Core already needs to be updated, so a new relay version is not much more IMHO.

> it feels like the same kind of problem as marshaling and unmarshaling json or http -- as an implementor you need to know how to interpret the bytes, but the protocol that transports them doesn't care

The protocol is not the issue here. Core cares and it cannot import the chainlink-starknet.

krehermann: updated to use common code from the chainlink-relay repo. please review again

jmank88 (dismissed):

jmank88 (dismissed):

jmank88 (approved):

PR #283, opened 2023-05-26
gauntlet: Integrate RDD

archseer:
Includes:
- https://smartcontract-it.atlassian.net/browse/BCI-1374
- https://smartcontract-it.atlassian.net/browse/BCI-1375
- https://smartcontract-it.atlassian.net/browse/BCI-1376


packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/setConfig.ts
```diff
@@ -24,7 +26,58 @@ type ContractInput = [
...
+      maxDurationObservationNanoseconds: time.durationToNanoseconds(config.maxDurationObservation).toNumber(),
+      maxDurationReportNanoseconds: time.durationToNanoseconds(config.maxDurationReport).toNumber(),
+      maxDurationShouldAcceptFinalizedReportNanoseconds: time
+        .durationToNanoseconds(config.maxDurationShouldAcceptFinalizedReport)
+        .toNumber(),
+      maxDurationShouldTransmitAcceptedReportNanoseconds: time
+        .durationToNanoseconds(config.maxDurationShouldTransmitAcceptedReport)
+        .toNumber(),
+      configPublicKeys: operatorConfigPublicKeys,
+    };
```
archseer: This is a big blob that should be part of gauntlet-core. Duplicated on other chains

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/deploy.ts
```diff
@@ -20,13 +20,29 @@ type ContractInput = [
...
+      maxAnswer: aggregator.maxSubmissionValue,
+      minAnswer: aggregator.minSubmissionValue,
+      decimals: aggregator.decimals,
+      description: aggregator.name,
+      billingAccessController: aggregator.billingAccessController,
+      linkToken: aggregator.linkToken,
+      owner: aggregator.owner || env.account,
+    }
+  }
+
```
calvwang9: need to use `getRDD` in `makeUserInput` rather than `makeContractInput` as to capture the contract addr arg and pass input validation from gauntlet-core

calvwang9: manually tested to be all working with a local devnet, but noting that the rdd files from [this PR as an example](https://github.com/smartcontractkit/reference-data-directory/pull/2719/files#diff-b74a33dc406f1252576900011901b9afce87b90c27a398cf46ea2421b9616d77) are missing a few required fields: `owner`, `linkToken`, `billingAccessController` (not sure if these are intended to be passed as env vars instead?)

sdrug: 1. Yes, all RDD parsing should happen when handling user input in `makeUserInput`.
2. Those fields are to be passed with either flags or env vars. Afaik, gauntlet core parses it be default, e.g. if you pass billingAccessController address with `BILLING_ACCESS_CONTROLLER` env.

calvwang9: thanks @sdrug, seems a bit inconsistent in the way its being used, but ill add support so that these fields can be defined by either RDD, flags, or env vars (in that order of priority)

archseer: Yeah I switched it over from `makeUserInput` to  `makeContractInput` where possible since it seemed more reasonable (we're constructing a separate contract input, not manipulating user input) but we can use `makeUserInput` where not possible

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/setPayees.ts
```diff
@@ -0,0 +1,57 @@
...
+  payee: string
+}
+
+type UserInput = PayeeConfig[]
+
+type ContractInput = [payees: PayeeConfig[]]
+
+const makeUserInput = async (flags, args): Promise<UserInput> => {
+  if (flags.input) return flags.input as UserInput
+  const rdd = getRDD(flags.rdd)
```
calvwang9: this implementation means input must be provided directly through the `input` flag or through RDD

archseer: Since this is an ExecuteCommand, `makeContractInput`'s `ctx.rdd` will be available and you can avoid loading this manually in `makeUserInput`.

then this becomes a two-liner, like `setBilling`:

```js
if (ctx.rdd) {
  const aggregator = ctx.rdd[CONTRACT_TYPES.AGGREGATOR][ctx.contractAddress];
  input = aggregator.payees
}
```

calvwang9: thanks, changed to this implementation and added a manual way to define input params from cli args 👍

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/inspection/inspect.ts
```diff
@@ -3,11 +3,36 @@ import {
...
+    observationPaymentGjuels: string
+    transmissionPaymentGjuels: string
+    gasBase: string
+    gasPerSignature: string
+  }
+  linkAvailableForPayment: {
+    isNegative: boolean
+    absoluteDifference: string
+  }
+}
```
calvwang9, line 34: Assume this is all the required fields and expected formatting/types from the context given from the ticket - please let me know if anything was missed

calvwang9: example output:
```
{
  typeAndVersion: 'ocr2/aggregator.cairo 1.0.0',
  description: 'test / test',
  owner: '0x0727664d8de5dc8b8d200bfa10685ab1826e4144713e6522db7dc5c5a9138578',
  decimals: 8,
  latestConfigDetails: {
    configCount: 1,
    blockNumber: 6,
    configDigest: '0x432de9deb179d411ebe03070a1b05a3ab95f99eb4ff50fcc7be4414f45a4f'
  },
  transmitterInfo: [
    {
      transmitter: '0x06da90a36b310e54c14ec686ea3d10fc44a14303d45b1f90d93589db8f99f3ef',
      owedPayment: '0'
    },
    {
      transmitter: '0x04737098fd5346ad65b322da48dbd310a1d00cd0d6609160599c8531d0839e92',
      owedPayment: '0'
    },
    {
      transmitter: '0x03490e8705187e7bd691f516d96737265c71bcac14d0deb53c5f81fb6266c68c',
      owedPayment: '0'
    },
    {
      transmitter: '0x018358bee8a222fa809fa392acc497d1423ed086fca91f93bcf71c8e7b4af2b4',
      owedPayment: '0'
    }
  ],
  billing: {
    observationPaymentGjuels: '0',
    transmissionPaymentGjuels: '0',
    gasBase: '0',
    gasPerSignature: '0'
  },
  linkAvailableForPayment: { isNegative: true, absoluteDifference: '0' }
}
```

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/setConfig.ts
```diff
@@ -24,7 +26,82 @@ type ContractInput = [
   offchain_config: string[],
 ]
 
-const makeContractInput = async (input: SetConfigInput): Promise<ContractInput> => {
+const makeUserInput = async (flags, args, env): Promise<SetConfigInput> => {
+  if (flags.input) return flags.input as SetConfigInput
+  if (flags.rdd) {
```
calvwang9, line 31: had to move the rdd part into `makeUserInput` to pass the input validation from gauntlet-core (cc @archseer), open to any better ways to do this

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/deploy.ts
```diff
@@ -20,13 +26,36 @@ type ContractInput = [
...
+    const aggregator = rdd[CONTRACT_TYPES.AGGREGATOR][contractAddr]
+    return {
+      maxAnswer: aggregator.maxSubmissionValue,
+      minAnswer: aggregator.minSubmissionValue,
+      decimals: aggregator.decimals,
+      description: aggregator.name,
+      billingAccessController:
+        aggregator.billingAccessController ||
+        flags.billingAccessController ||
+        env.billingAccessController ||
```
sdrug: `billingAccessController` was commonly referred as `BILLING_ACCESS_CONTROLLER` in the env; `linkToken` as `LINK`. Please check with the gauntlet team and/or their codebase for the default values, let's stay consistent.

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/deploy.ts
```diff
@@ -20,13 +26,36 @@ type ContractInput = [
...
+        '',
+      linkToken: aggregator.linkToken || flags.linkToken || env.linkToken || '',
+      owner: aggregator.owner || env.account,
```
sdrug: Is the owner specified in RDD?

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/setBilling.ts
```diff
@@ -14,13 +18,21 @@ type ContractInput = [
...
 ]
 
-const makeContractInput = async (input: StarknetSetBillingInput): Promise<ContractInput> => {
+const makeContractInput = async (
+  input: StarknetSetBillingInput,
+  ctx: ExecutionContext,
+): Promise<ContractInput> => {
+  if (ctx.rdd) {
+    const contract = ctx.rdd[CONTRACT_TYPES.AGGREGATOR][ctx.contractAddress]
+    input = contract.billing
```
sdrug, line 27: @archseer current RDD doesn't contain those two additional fields. Should RDD tool be adjusted accordingly?

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/setConfig.ts
```diff
@@ -24,7 +26,82 @@ type ContractInput = [
...
+    const rdd = getRDD(flags.rdd)
+    const contractAddr = args[0]
+    const contract = rdd[CONTRACT_TYPES.AGGREGATOR][contractAddr]
+    const config = contract.config
+    const operators = contract.oracles.map((oracle) => rdd.operators[oracle.operator])
+    const publicKeys = operators.map((o) =>
+      o.ocr2OffchainPublicKey[0].replace('ocr2off_starknet_', ''),
```
sdrug: I'd propose all the replacements (i.e. content transformations) to happen in `makeContractInput` for consistency.

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/setConfig.ts
```diff
@@ -52,6 +129,9 @@ const makeContractInput = async (input: SetConfigInput): Promise<ContractInput>
   return [oracles, new BN(input.f).toNumber(), onchainConfig, 2, bytesToFelts(offchainConfig)]
 }
 
+// TODO: beforeExecute attempt to deserialize offchainConfig to check for validity
+// TODO: diff onchain config vs proposed config
```
sdrug, line 127: Is it planned to be addressed in this PR?

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/setPayees.ts
```diff
@@ -0,0 +1,66 @@
...
+  payee: string
+}
+
+type UserInput = PayeeConfig[]
+
+type ContractInput = [payees: PayeeConfig[]]
+
+const makeUserInput = async (flags, args): Promise<UserInput> => {
+  if (flags.input) return flags.input as UserInput
+  if ((args.length - 1) % 2 != 0) throw new Error('Invalid number of arguments for payee config')
```
sdrug: I think it's uncommon for this version of gauntlet to pass this data as arguments. Please check how it's done for other similar commands. I'd imagine `--transmitters` and `--payees` flags or something similar.

sdrug: This data is also available in RDD, isn't it?

sdrug: Let's handle RDD input here too, it's a form of user input.

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/setPayees.ts
```diff
@@ -0,0 +1,66 @@
...
+  return input.map((payeeConfig) => [
+    {
+      transmitter: payeeConfig.transmitter,
+      payee: payeeConfig.payee,
+    },
+  ]) as ContractInput
+}
+
+const commandConfig: ExecuteCommandConfig<UserInput, ContractInput> = {
+  contractId: 'ocr2',
```
sdrug: Check other commands, contractId and category aren't normally hardcoded.

packages-ts/starknet-gauntlet/src/commands/base/executeCommand.ts
```diff
@@ -111,6 +113,12 @@ export const makeExecuteCommand = <UI, CI>(config: ExecuteCommandConfig<UI, CI>)
         contract: new Contract(c.contract.abi, c.contractAddress ?? '', c.provider.provider),
       }
 
+      const rdd = flags.rdd || process.env.RDD
```
sdrug, line 116: Imo it's not a good place to inject an RDD dependency here since it's a user input.

packages-ts/starknet-gauntlet/src/rdd.ts
```diff
@@ -0,0 +1,23 @@
+import { existsSync, readFileSync } from 'fs'
+
+export enum CONTRACT_TYPES {
+  PROXY = 'proxies',
+  ACCESS_CONTROLLER = 'accessControllers',
+  AGGREGATOR = 'contracts',
+  VALIDATOR = 'validators',
+}
+
+// throw new Error(`No reference data directory specified!  Must pass in the '--rdd' flag or set the 'RDD' env var`)
```
sdrug: ?

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/setBilling.ts
```diff
@@ -30,10 +42,10 @@ const commandConfig: ExecuteCommandConfig<StarknetSetBillingInput, ContractInput
   makeUserInput: (flags: any, args: any): StarknetSetBillingInput => {
     if (flags.input) return flags.input as StarknetSetBillingInput
     return {
-      observationPaymentGjuels: flags.observationPaymentGjuels,
-      transmissionPaymentGjuels: flags.transmissionPaymentGjuels,
-      gasBase: flags.gasBase,
-      gasPerSignature: flags.gasPerSignature,
+      observationPaymentGjuels: parseInt(flags.observationPaymentGjuels),
```
krebernisak: These changes are already available in another PR: https://github.com/smartcontractkit/chainlink-starknet/pull/294

Consider rebasing on top of that branch.

archseer: The changes here pre-date this branch. We'll rebase after it's merged

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/setPayees.ts
calvwang9: I moved category into a `constants.ts`, though other commands seem to import from gauntlet-core, in which [contractID](https://github.com/smartcontractkit/gauntlet-core/blob/293122f4f75016e1e77c3107da3cb5f756125857/packages/contracts-ocr2/src/commands/setBilling.ts#L33) does seem to be hardcoded 🤔

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/setPayees.ts
calvwang9: thanks, have fixed to use flags instead of arg input.

re RDD: following up on @archseer 's comment here - https://github.com/smartcontractkit/chainlink-starknet/pull/283#discussion_r1210360452, lets decide on a consistent approach here

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/setConfig.ts
calvwang9: fair, looks like the replace was doubled up anyway in `makeContractInput` so removed this one 👍

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/deploy.ts
calvwang9: adding this to allow for the owner to be defined in RDD, but will fallback to flags or env var if not

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/deploy.ts
calvwang9: good catch, fixed thanks

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/setPayees.ts
sdrug: Other command sets within the same package do operate with contract_list and categories - [example](https://github.com/smartcontractkit/chainlink-starknet/blob/62ee8f90498188ca9a4d0f2ae3060244bcbf0111/packages-ts/starknet-gauntlet-ocr2/src/commands/proxy/proposeAggregator.ts#L45)

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/deploy.ts
sdrug: RDD isn't an arbitrary file with data. It's generated with [RDD tool](https://github.com/smartcontractkit/rddtool) and the `owner` field isn't part of it. Although it falls back to the correct value right now, we cannot know if `owner` field will be added in the future and what would be the meaning of it. It also causes confusion when reading the code. Same applies to the all aggregator fields, e.g. `linkToken` up there.

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/setConfig.ts
calvwang9: todo is captured in https://smartcontract-it.atlassian.net/browse/BCI-1379 (separate PR)

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/deploy.ts
calvwang9: i see, thanks for the context @sdrug  I'll keep it consistent with other parts of the codebase for this PR then, and we can maybe explore updating the RDD tool with additional fields as a later piece of work?

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/setPayees.ts
calvwang9: ah thanks, missed that one 👍 just noting that there are some inconsistencies between commands that import from gauntlet-core (contractID = 'ocr2') and commands that import locally from lib/contracts (contractID = 'Aggregator') but not sure if there are any implications and will leave it for now

packages-ts/starknet-gauntlet/src/commands/base/executeCommand.ts
calvwang9: as I understand it, this makes rdd available in the execution context which links back to the two different approaches to RDD handling - https://github.com/smartcontractkit/chainlink-starknet/pull/283#discussion_r1210360452

lets resolve in one thread

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/setBilling.ts
archseer: Similar to Terra we should yeah

packages-ts/starknet-gauntlet/src/commands/base/executeCommand.ts
archseer: I much prefer this solution since it injects RDD across **all** commands. It's a special input (`--rdd`) but most often actually set as an env var. It doesn't make sense as part of `makeUserInput` since we're not actually changing the input but we're instead sourcing the data from somewhere else. In my opinion entangling this into `makeUserInput` under each command resulted in a lot more code being added on Cosmos, and it was also hard to debug since your existing input would just disappear from an early return and other flags would be applied.

sdrug: The initial idea of `makeUserInput` was to resolve the conflicting inputs, do basic validation and raise exceptions if something is missing, then the input is being transformed into the actual contract input with `makeContractInput`. It's been consistently (well, more or less) done across our previous integrations and that it why I mention it here too. Anyways, I won't block you on that since this whole solution is about to be deprecated.

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/deploy.ts
sdrug: Please create a ticket for that and sync with John regarding scheduling it into the sprint. It's a blocker for EngOps if they can't source full data from RDD files.

calvwang9: Tracking this work here - https://smartcontract-it.atlassian.net/browse/BCI-1405

sdrug (approved):

PR #285, opened 2023-05-26
RDD integration for set_config

sdrug:


archseer:
Thanks! I already have an alternative here though https://github.com/smartcontractkit/chainlink-starknet/pull/283

PR #294, opened 2023-06-01
BCI-1335: remove bn.js

augustbleeds:


packages-ts/starknet-gauntlet-oz/src/commands/account/deploy.ts
```diff
@@ -6,7 +6,7 @@ import {
 } from '@chainlink/starknet-gauntlet'
 import { ec } from 'starknet'
 import { CATEGORIES } from '../../lib/categories'
-import { accountContractLoader, CONTRACT_LIST, equalAddress } from '../../lib/contracts'
```
augustbleeds, line 9: equalAddress is unused

packages-ts/starknet-gauntlet-ocr2/test/commands/ocr2.test.ts
```diff
@@ -182,14 +181,15 @@ describe('OCR2 Contract', () => {
...
       for (let i = 0; i < signers.length; i++) {
-        const signer = new BN(eventData[4 + 2 * i].replace('0x', ''), 16) // split according to event structure
+        const signer = BigInt(eventData[4 + 2 * i]) // split according to event structure
         eventSigners.push(signer)
       }
 
       expect(eventSigners).toEqual(
-        signers.map((s) => new BN(s.replace('ocr2on_starknet_', '').replace('0x', ''), 16)),
+        // eaiser to remove prefix and 0x and then add 0x back
+        signers.map((s) => BigInt(`0x${s.replace('ocr2on_starknet_', '').replace('0x', '')}`)),
```
krebernisak, line 192: Do we need this extra replace?

```suggestion
        signers.map((s) => BigInt(`s.replace('ocr2on_starknet_', '')`)),
```

packages-ts/starknet-gauntlet/src/commands/base/declareCommand.ts
krebernisak, line 1: Added by accident?

packages-ts/starknet-gauntlet/test/encoding/encoding.test.ts
```diff
@@ -12,8 +11,8 @@ export function bytesToFeltsDeprecated(data: Uint8Array): string[] {
   // chunk every 31 bytes
   for (let i = 0; i < data.length; i += CHUNK_SIZE) {
     const chunk = data.slice(i, i + CHUNK_SIZE)
-    // cast to int
-    felts.push(new BN(chunk, 'be').toString())
+    const feltHex = `0x${Array.from(chunk, (byte) => byte.toString(16).padStart(2, '0')).join('')}`
```
krebernisak: Hmm, what's happening here?

Can we extract this in a fn with a name and a potential comment?

packages-ts/starknet-gauntlet/src/commands/base/declareCommand.ts
augustbleeds: Whoops! Yeah that's by accident.

packages-ts/starknet-gauntlet-ocr2/test/commands/ocr2.test.ts
augustbleeds: Yeah, some of the testcases have the "0x" prefix and some don't. So it's easier to say "let's remove 0x if you have it and add it back later" rather than "let's only add 0x if you don't have it". 

Same thing for the ocr2_on_starknet_ infix as well

packages-ts/starknet-gauntlet/test/encoding/encoding.test.ts
augustbleeds: So this line is converting 31 bytes to a decimal string. But instead of using bn.js, we are converting each byte to a hex string and concatenating all the strings together. Then we turn the resultant hexadecimal string to a bigint and then convert it to a decimal string. I'll extract this into a function

augustbleeds: The .padStart is to ensure strings such as 'f' are represented by '0f' (pad with 0s) so each byte is represented by two characters so that when you concatenate the string no places are omitted

archseer: This is just a compat test with the old implementation, we could remove it altogether 👍🏻

archseer (approved):

PR #296, opened 2023-06-02
BCI-1381: Add Distinct declare vs. deploy commands

augustbleeds:
This PR adds the ability to 
1. create `declare` actions which only declare a contract
2. deploy (no declare) contracts using a given --classHash

So in practice it works something like this
```
# declare only
yarn gauntlet SequencerUptimeFeed:declare --network=testnet

# deploy only
yarn gauntlet SequencerUptimeFeed:deploy --initialStatus=0 --classHash=0x0149c2a0a58155a8a091ded689a5fdcccd596bb0c74b301f678c1b46b690f432 --network=testnet
```

~I implemented the feature + did it for one contract to start with. If this looks good I'll update all contracts + docs with similar functionalities~

`declare` commands and the `deploy --classHash` flag were added for the following contracts

[x] access_controller:
[x] ocr2
[x] proxy
[x] Example
[x] account
[x] argent_account
[x] multisig
[x] SequencerUptimeFeed
[x] token


packages-ts/starknet-gauntlet-emergency-protocol/src/commands/sequencerUptimeFeed/deploy.ts
```diff
@@ -24,6 +25,13 @@ const validateOwner = async (input) => {
   return true
 }
 
+const validateClassHash = async (input) => {
```
augustbleeds, line 28: class hashes and addresses are both pedersen hashes so they have the same validation check

packages-ts/starknet-gauntlet/src/commands/base/executeCommand.ts
```diff
@@ -183,13 +190,28 @@ export const makeExecuteCommand = <UI, CI>(config: ExecuteCommandConfig<UI, CI>)
...
+      if (classHash === undefined) {
+        tx = await this.provider.deployContract(
+          this.contract,
+          this.compiledContractHash,
+          this.input.contract,
+          false,
+          this.input?.user?.['salt'],
+        )
+      } else {
+        tx = await this.provider.onlyDeployContract(
```
augustbleeds: maybe this should be called `deployWithClassHash` ?

krebernisak: The original starknet.js names fit best:

- `deploy`
- `declareAndDeploy`

But heads up that the new Gauntlet fork will limit a single operation to a single side-effect, and declare + deploy are two transactions.

You can check out the Starknet core extension of the new Gauntlet here: https://github.com/smartcontractkit/gauntlet-temporal-example/pull/33/files

cfal: it might be a bigger change, but it seems cleaner if we call the existing `declareAndDeployContract` and let this be `this.provider.deployContract` - or is there conflicts elsewhere that makes this impossible?

packages-ts/starknet-gauntlet/src/provider/index.ts
```diff
@@ -104,6 +115,37 @@ class Provider implements IStarknetProvider {
     return response
   }
 
+  declareContract = async (contract: CompiledContract, compiledClassHash?: string, wait = true) => {
+    const tx = await this.account.declare({
+      contract,
+      compiledClassHash,
+    })
+
+    const response = wrapResponse(this, tx, 'not applicable for declares')
```
cfal, line 132: could skip the optional `address` param?

packages-ts/starknet-gauntlet-emergency-protocol/src/commands/sequencerUptimeFeed/deploy.ts
```diff
@@ -24,6 +25,13 @@ const validateOwner = async (input) => {
...
+  if (isValidAddress(input.classHash) || input.classHash === undefined) {
+    return true
+  }
+  throw new Error(`Invalid Owner Address: ${input.owner}`)
```
cfal: should this error message be invalid class hash?

packages-ts/starknet-gauntlet/src/commands/base/executeCommand.ts
```diff
@@ -205,6 +227,29 @@ export const makeExecuteCommand = <UI, CI>(config: ExecuteCommandConfig<UI, CI>)
...
+      const tx = await this.provider.declareContract(
+        this.contract,
+        this.compiledContractHash,
+        false,
+      )
+
+      deps.logger.loading(`Waiting for tx confirmation at ${tx.hash}...`)
+      const response = await tx.wait()
+
+      if (!response.success) {
```
krebernisak, line 255: I'm not sure this `success` flag is set properly. Please double check the [wait()](https://github.com/smartcontractkit/chainlink-starknet/blob/bf537fb7faf93eb4f516fcc47990b4d527e6c99c/packages-ts/starknet-gauntlet/src/provider/index.ts#L54) method.

It expects the underlying `provider.waitForTransaction` to throw, and if it doesn't throw it sets the `success` flag. The problem is it looks like the underlying fn will return a status enum, not throw.

packages-ts/starknet-gauntlet/src/commands/base/executeCommand.ts
augustbleeds: Maybe to clear some things up... 
1. `this.provider.deployContract` does ` declareAndDeploy` which will as expected both declare and deploy the contract. Additionally, it will skip declaring and only deploy if the contract has already been declared.
2. `this.provider.onlyDeployContract` only does the deploy part.

You might be wondering why we need `this.provider.onlyDeployContract` because `this.provider.deployContract` seems to already have the functionality we want -- that it'll only deploy if the contract has been declared once. We still need this new command because it is safer and faster to reuse a class hash than to recompile the contract (because the contract and the compiler may have changed -- especially with how starknet updates are). If the code changes or the compiler changes, we have the option is using previously declared versions. But if you want to use the updated code/compiler you'll need to redeclare the (technically new) contract without --classHash argument. This gives the user options.

So I think we still need 3 functions (which I'll also rename for clarity sake)

compiles the contract and declares it using the resultant generated ABI, and deploys it
`this.provider.deployContract` -> rename to `this.provider.declareAndDeployContract()`

deploys the contract given a class hash
`this.provider.onlyDeploy` -> rename to `this.provider.deployContract`

compiles the contract and declares it using the resultant generated ABI
`this.provider.declareContract` -> (no need to rename)

@krebernisak So eventually you're saying we'll need to split up the declare and deploy to two seperate Gauntlet commands?

packages-ts/starknet-gauntlet/src/commands/base/executeCommand.ts
augustbleeds: I can confirm it's set as expected -- we use this pattern already elsewhere in the code and looking at https://github.com/0xs34n/starknet.js/blob/b2e7a4046c8bda541cf0228469a318443d94e788/src/provider/rpc.ts#L405 it does seem that error states trigger an exception to be thrown

packages-ts/starknet-gauntlet/src/provider/index.ts
augustbleeds: Yeah, I just kept it there cause it's weird to have nothing printed out on the command line. It might confuse users who are looking for an address and suspect something is wrong

packages-ts/starknet-gauntlet-emergency-protocol/src/commands/sequencerUptimeFeed/deploy.ts
augustbleeds: good catch :) !

archseer (commented):
Since all the declares are essentially identical, could you derive them the same way you do for the upgrade command?

augustbleeds:
> Since all the declares are essentially identical, could you derive them the same way you do for the upgrade command?

yeah i can do that for sure 👍

PR #303, opened 2023-06-13
Add set_config diff 

calvwang9:
https://smartcontract-it.atlassian.net/browse/BCI-1379

Add a beforeExecute step on setConfig to print out a diff of the config changes

### DIFF
--- a/packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/inspection/configEvent.ts
+++ b/packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/inspection/configEvent.ts
@@ -23,6 +23,7 @@ export const getLatestOCRConfigEvent = async (
 
   // retrieve all block traces in the block in which the latest config was set
   const blockTraces = await provider.provider.getBlockTraces(latestConfigDetails.blockNumber)
+
   // retrieve array of all events across all internal calls for each tx in the
   // block, for which the contract address = aggregator contract and the first
   // event key matches 'ConfigSet'
@@ -33,6 +34,12 @@ export const getLatestOCRConfigEvent = async (
       .filter((event) => event.keys[0] === hash.getSelectorFromName('ConfigSet'))
   })
 
-  // assume last event found is the latest config set event
+  // if no config set events found in the given block, throw error
+  // this should not happen if block number in latestConfigDetails is set correctly
+  if (configSetEvents.length === 0)
+    throw new Error(`No ConfigSet events found in block number ${latestConfigDetails.blockNumber}`)
+
+  // assume last event found is the latest config, in the event that multiple
+  // set_config transactions ended up in the same block
   return configSetEvents[configSetEvents.length - 1].data
 }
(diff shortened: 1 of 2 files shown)


packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/inspection/configEvent.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  // retrieve block traces
+  const blockTraces = await provider.provider.getBlockTraces(latestConfigDetails.blockNumber)
+  // get list of all events across all internal calls in the block for which the key matches 'ConfigSet'
+  const configSetEvents = blockTraces.traces
+    .map((trace) => {
+      return trace.function_invocation.internal_calls
+        .filter((call) => call.contract_address === contractAddress)
+        .map((call) => call.events)
+        .flat()
```
mohamed-mehany: Nit: Would `flatMap((call) => call.events)` work here? just slightly more efficient.

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/setConfig.ts
```diff
@@ -123,8 +126,32 @@ const makeContractInput = async (
...
+  const { offchainConfig } = await encoding.serializeOffchainConfig(
+    input.user.offchainConfig,
+    input.user.secret,
+  )
+  const newOffchainConfig = encoding.deserializeConfig(offchainConfig)
+
+  const eventData = await getLatestOCRConfigEvent(context.provider, context.contractAddress)
+  if (eventData.length === 0) {
+    deps.logger.info('No previous config found, review the offchain config below:')
+    console.log(newOffchainConfig)
```
mohamed-mehany: Just curious, why do we need to use different logging methods here? Sorry if it's too trivial as I'm not yet familiar with this codebase.

mohamed-mehany (commented):
Code looks great! 
Just a few nits.

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/inspection/configEvent.ts
calvwang9: thanks, much neater!

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/setConfig.ts
calvwang9: good catch, this slipped through 😅

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/inspection/configEvent.ts
```diff
@@ -0,0 +1,38 @@
...
+  // block, for which the contract address = aggregator contract and the first
+  // event key matches 'ConfigSet'
+  const configSetEvents = blockTraces.traces.flatMap((trace) => {
+    return trace.function_invocation.internal_calls
+      .filter((call) => call.contract_address === contractAddress)
+      .flatMap((call) => call.events)
+      .filter((event) => event.keys[0] === hash.getSelectorFromName('ConfigSet'))
+  })
+
+  // assume last event found is the latest config set event
```
sdrug: Maybe handle the case when no events were found to prevent a negative index and let the caller know?

calvwang9: Thanks, this shouldnt happen if the block number from latest_config_details was set properly, but good to catch just in case👍

archseer (approved):

PR #306, opened 2023-06-19
Add libfunc validation

calvwang9:
https://smartcontract-it.atlassian.net/browse/BCI-1424

Note: this uses Scarb v0.4.0 (Cairo v1.1.0), however, there is no way to specify the Scarb version in the toml config file. The Scarb version is only defined in the CI.

For the latest Cairo version running on testnet (v1.1.0), we currently use some disallowed Sierra libfuncs in our contracts, namely `upcast`, `downcast`, `ec_point_from_x_nz`. These libfuncs are missing from the current testnet allowlist but are in the audited allowlist in the latest Cairo release candidate v2.0.0-rc2. 

This means that our Starknet contracts are not deployable until Cairo v2 is released, and a todo has been left to update the allowlist being used after the Cairo v2 release

### DIFF
--- a/.github/actions/install-cairo/action.yml
+++ b/.github/actions/install-cairo/action.yml
@@ -8,7 +8,7 @@ inputs:
     required: false
   scarb_version:
     description: Scarb release version
-    default: "v0.3.0"
+    default: "v0.4.0"
     required: false
 
 runs:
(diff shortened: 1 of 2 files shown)


archseer:
Scarb needs bumping here https://github.com/smartcontractkit/chainlink-starknet/blob/9bdc96508e8e5d4339831858e12d571d259e0812/.github/actions/install-cairo/action.yml#L9-L12

cfal (approved):

PR #307, opened 2023-06-23
 SonarQube integration

davidcauchi:


jmank88 (dismissed):

.github/workflows/relayer.yml
```diff
@@ -24,6 +24,16 @@ jobs:
 
       - name: Unit Test
         run: nix develop -c make test-unit-go
+      
+      - name: Upload Golangci relayer results
+        if: always()
```
krebernisak, line 29: Should not be needed if always:
```suggestion
```

.github/workflows/golangci-lint.yml
```diff
@@ -66,6 +78,12 @@ jobs:
           nix_path: nixpkgs=channel:nixos-unstable
       - name: Lint integration-tests
         run: nix develop -c make lint-go-test
+      - name: Store Golangci lint integration tests report artifact
+        if: always()
```
krebernisak, line 82: Should not be needed if always:
```suggestion
```

.github/workflows/golangci-lint.yml
```diff
@@ -52,6 +58,12 @@ jobs:
           nix_path: nixpkgs=channel:nixos-unstable
       - name: Lint ops
         run: nix develop -c make lint-go-ops
+      - name: Store Golangci lint ops report artifact
+        if: always()
```
krebernisak, line 62: Should not be needed if always:
```suggestion
```

.github/workflows/golangci-lint.yml
```diff
@@ -38,6 +38,12 @@ jobs:
           nix_path: nixpkgs=channel:nixos-unstable
       - name: Lint relayer
         run: nix develop -c make lint-go-relayer
+      - name: Store Golangci lint relayer report artifact
+        if: always()
```
krebernisak, line 42: Should not be needed if always:
```suggestion
```

.github/workflows/sonar-scan.yml
```diff
@@ -0,0 +1,97 @@
+name: SonarQube Scan
```
krebernisak, line 1: Is there a way to centralize this workflow so each project doesn't need to define it? (e.g., pull a list of projects and scan them in a central place).

Makefile
```diff
@@ -162,15 +162,15 @@ format-ts-check:
 
 .PHONY: lint-go-ops
 lint-go-ops:
```
krebernisak, line 159: Please use this opportunity to make these Makefile actions and CI workflows consistent across our projects (Solana, Cosmos, Starknet).

.github/workflows/sonar-scan.yml
```diff
@@ -0,0 +1,97 @@
...
+          exclude-workflow-ids: ""
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+        env:
+          DEBUG: "true"
+
+  sonarqube:
+    name: SonarQube Scan
+    needs: [ wait_for_workflows ]
+    runs-on: ubuntu-latest
+    if: always()
```
krebernisak, line 36: Should not be needed if always:
```suggestion
```

.github/workflows/relayer.yml
jmank88: Isn't always necessary to ensure we still run on failure? Default is success only IIRC

.github/workflows/golangci-lint.yml
chudilka1: We still want to get artifacts uploaded, no matter if the linter (a previous step) fails. So `if: (always)` is needed here, like in the rest of other cases.

.github/workflows/sonar-scan.yml
chudilka1: Unfortunately, not at the moment. The main reason, we have different workflows across our projects, different numbers of triggered pipelines, and variously named artifacts produced as a result of tests/linters runs. So as soon as we get to the point of a more unified/formalized project structure, your great improvement idea could take place.

chudilka1 (approved):

.github/workflows/sonar-scan.yml
```diff
@@ -0,0 +1,97 @@
...
+          name_is_regexp: true
+          name: golangci-lint-ops-report
+          if_no_artifact_found: warn
+      
+      - name: Download Go Integration tests report
+        uses: dawidd6/action-download-artifact@v2.27.0
+        with:
+          workflow: golangci-lint.yml
+          workflow_conclusion: ""
+          name_is_regexp: true
```
cfal, line 75: name doesn't need to be a regexp: https://github.com/dawidd6/action-download-artifact#usage

cfal (dismissed):
lgtm, left a nit about name_is_regexp

.github/workflows/golangci-lint.yml
```diff
@@ -66,6 +78,12 @@ jobs:
...
         run: nix develop -c make lint-go-test
+      - name: Store Golangci lint integration tests report artifact
+        if: always()
+        uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8 # v3.1.0
+        with:
+          name: golangci-lint-integration-tests-report
+          path: ./integration-tests/golangci-lint-integration-tests-report.xml
       # Note: I could not figure out why the golangci-lint-action would not work even though it is technically running the same as above, error message is this:
       # Running [/home/runner/golangci-lint-1.50.1-linux-amd64/golangci-lint run --out-format=github-actions --path-prefix=integration-tests --exclude=dot-imports] in [/home/runner/work/chainlink-starknet/chainlink-starknet/integration-tests] ...
       # level=warning msg="[runner] Can't run linter goanalysis_metalinter: inspect: failed to load package client: could not load export data: no export data for \"github.com/smartcontractkit/chainlink-testing-framework/client\""
```
chudilka1, line 89: The report is generated, take a look [here](https://github.com/smartcontractkit/chainlink-starknet/actions/runs/6082983043/job/16502027373?pr=307#step:5:9)
I believe the version of the linter, should be bumped to `v1.51.1 # this version has an --out-format checkstyle necessary for Sonar`
Ref: smartcontractkit/golangci-lint-action@54ab6c5f11d66a92d14c3f7cc41ea13f676644bd

To test, you can try adding the unused variable to a go file analyzed by linter, eventually you should get a code smell in Sonar with label `GOLANGCI-LINT`. :warning: Better to test whether all the separately analyzed parts of the code produce the corresponding results in Sonar

image

chudilka1 (changes requested):

smickovskid (approved):

PR #310, opened 2023-07-12
Use latest upstream sdk: starknet.go

archseer:


contracts/src/libraries/access_control.cairo
```diff
@@ -1,42 +1,66 @@
-#[contract]
+use starknet::ContractAddress;
+#[starknet::interface]
+trait IAccessController<TContractState> {
```
archseer: Maybe makes more sense under the access_controller module rather than in libraries

PR #312, opened 2023-08-01
[Functions] Add provider method stub in relayer

bolekk:


jmank88 (dismissed):

aalu1418 (changes requested):
I think you still need a `make format-go-mod-tidy` to update the dependencies in the various test files

aalu1418 (approved):
approving since test failures seem pre-existing (smoke tests) or completely unrelated (gauntlet tests), and the codebase is still being worked on pretty heavily

PR #325, opened 2023-09-18
BCI-1780: Failed Tx Inclusion RPC Changes

augustbleeds:


integration-tests/go.mod
```diff
@@ -388,6 +388,8 @@ require (
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
+replace github.com/NethermindEth/starknet.go => github.com/augustbleeds/starknet.go v0.4.2-0.20230912155424-048179bee72e
```
jmank88, line 391: Remove? 

Or group below and add a comment? And would it be worth creating a `smartcontkit/strknet.go` fork repo?

augustbleeds: thanks for the review, the replace was just meant to be a temp solution. I can remove it now because starknet.go has merged the changes into their repo!

archseer:
Subsumed by #310

PR #326, opened 2023-09-19
BCI-1923: Cairo Test Upgradeable

augustbleeds:
Changes
- [x] Introduce 2 mock contracts (for testing purposes)
- [x] Uncomments ignore macro to test upgradeability functionality
- [x] Adds additional test that deploys and upgrades a contract and tests the function calls are as expected  

note: we'll need to change these tests once contract composition is introduced because this will change the way we compose the Upgradeable module. See https://community.starknet.io/t/cairo-1-contract-syntax-is-evolving/94794?page=2#extensibility-and-components-11 for the proposed `#[starknet::component]` macro

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -9,6 +9,7 @@ mockery 2.22.1
 golangci-lint 1.51.2
 actionlint 1.6.12
 shellcheck 0.8.0
+scarb 0.7.0
 
 # Kubernetes
 k3d 5.4.4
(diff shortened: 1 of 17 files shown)


archseer:
Contract builds fail in a weird way. Maybe we need to bump the scarb version on CI?

augustbleeds:
> Contract builds fail in a weird way. Maybe we need to bump the scarb version on CI?

Thanks, I fixed the issue. It was failing locally too. I pinpointed the issue to the constructor of the mock contracts, which need to have a mutable reference instead of a snapshot to the contract state even if they are not writing anything to storage.

I also made some changes to implement the IUpgradeable trait. Do you think we should remove the `upgrade` function from the Multisig interface?

archseer:
Yeah I think that's fine, since we're no longer upstreaming the contract we can just implement `IUpgradeable` 👍🏻

archseer (approved):

PR #333, opened 2023-10-24
relayer/pkg/chainlink: move cmd/chainlink-starknet from chainlink repo

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-2701

Supports:
- https://github.com/smartcontractkit/chainlink/pull/11058

### DIFF
--- a/relayer/pkg/chainlink/cmd/chainlink-starknet/main.go
+++ b/relayer/pkg/chainlink/cmd/chainlink-starknet/main.go
@@ -1,0 +1,81 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/go-plugin"
+	"github.com/pelletier/go-toml/v2"
+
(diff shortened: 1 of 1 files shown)


relayer/pkg/chainlink/cmd/chainlink-starknet/main.go
jmank88, line 1: Is this a good spot? Or would `relayer/cmd/chainlink-starknet` make more sense? So that `cmd/` and `pkg/` are side-by-side?
cc @aalu1418

patrickhuie19: Are the only side-effects readability/cohesion?

patrickhuie19 (approved):

relayer/pkg/chainlink/cmd/chainlink-starknet/main.go
aalu1418: i think the location is fine :)

PR #343, opened 2024-01-09
Cairo 2.4.3 + components + starknet.go (RPC v0.6, tx v3)

archseer:


contracts/src/token/link_token.cairo
```diff
@@ -38,153 +75,45 @@ mod LinkToken {
...
+        self.erc20.initializer(NAME, SYMBOL);
         assert(!minter.is_zero(), 'minter is 0');
         self._minter.write(minter);
-        let mut ownable = Ownable::unsafe_new_contract_state();
-        Ownable::constructor(ref ownable, owner); // Ownable::initializer(owner);
+        self.ownable.initializer(owner);
     }
 
-    #[view]
+    // TODO #[view]
```
archseer, line 93: Note: need to still resolve this

contracts/src/token/link_token.cairo
```diff
@@ -38,153 +75,45 @@ mod LinkToken {
...
     fn minter(self: @ContractState) -> ContractAddress {
         self._minter.read()
     }
 
-    #[view]
+    // TODO #[view]
```
archseer: Note: need to add a type_and_version trait

.github/actions/install-cairo/action.yml
```diff
@@ -4,11 +4,11 @@ description: A composite action that installs cairo and scarb binaries
 inputs:
   cairo_version:
     description: Cairo release version
-    default: "v2.2.0"
+    default: "v2.4.3"
```
augustbleeds, line 7: in the future, we can create a scarb nix package and have CI use the nix shell

augustbleeds: We can also get rid of the cairo and scarb git submodules if we nixify. (scarb comes with cairo so i think we can also avoid the need to install cairo seperately from scarb).

contracts/src/token/link_token.cairo
```diff
@@ -14,21 +14,39 @@ mod LinkToken {
...
         _minter: ContractAddress,
+
+        #[substorage(v0)]
+        erc20: ERC20Component::Storage
+    }
+
+    #[event]
+    #[derive(Drop, starknet::Event)]
+    enum Event {
+        #[flat]
```
augustbleeds, line 60: what does this flat macro do?

.github/actions/install-cairo/action.yml
augustbleeds: also, we should add the new scarb version in .tool-versions file

augustbleeds (commented):
you can remove the #[available_gas...] macro for all cairo-tests (starting in 2.4.0) 😲

contracts/src/libraries/token/erc677.cairo
```diff
@@ -30,37 +37,42 @@ mod ERC677 {
...
         to: ContractAddress,
         value: u256,
         data: Array<felt252>
     }
 
-    fn transfer_and_call(
-        ref self: ContractState, to: ContractAddress, value: u256, data: Array<felt252>
-    ) -> bool {
-        let sender = starknet::info::get_caller_address();
+    #[embeddable_as(ERC677Impl)]
```
augustbleeds, line 51: from the cairo-book:

```Note: To avoid the confusion between the embeddable name and the impl name, we recommend keeping the suffix Impl in the impl name.```

~It's currently flipped. We should switch to~
~`#[embeddable_as(ERC677)]`~
~and~
~`impl ERC677Impl<...`~

EDIT: I don't think it really matters

File: contracts/src/libraries/token/erc677.cairo [hunk 6]
```diff
@@ -30,37 +37,42 @@ mod ERC677 {
...
+    impl ERC677<
+        TContractState, +HasComponent<TContractState>, +IERC20<TContractState>, +Drop<TContractState>,
```
augustbleeds: neat how you can enforce that the contract must implement ERC20 Interface. 

But why does ContractState need to be droppable? It's always used through a mutable reference or snapshot right?

contracts/src/tests/test_erc677.cairo
```diff
@@ -54,11 +55,13 @@ fn setup_invalid_receiver() -> (ContractAddress, MockInvalidReceiverDispatcher)
...
 }
 
+type ComponentState = ERC677Component::ComponentState<chainlink::token::link_token::LinkToken::ContractState>;
+
 fn transfer_and_call(receiver: ContractAddress) {
     let data = ArrayTrait::<felt252>::new();
     // have to send 0 because ERC20 is not initialized with starting supply when using this library by itself
-    let mut erc677 = ERC677::unsafe_new_contract_state();
-    //  TODO: ERC677::transfer_and_call(ref erc677, receiver, u256 { high: 0, low: 0 }, data);
+    let mut state: ComponentState = ERC677Component::component_state_for_testing();
```
augustbleeds, line 64: nice! good to know how this is done -- where'd you find this info haha?

contracts/src/libraries/ownable.cairo
```diff
@@ -1,110 +1,155 @@
...
+// TODO: consider replacing with OwnableTwoStep if https://github.com/OpenZeppelin/cairo-contracts/pull/809/ lands
+
 #[starknet::interface]
-trait IOwnable<TContractState> {
-    fn owner(self: @TContractState) -> ContractAddress;
-    fn proposed_owner(self: @TContractState) -> ContractAddress;
-    fn transfer_ownership(ref self: TContractState, new_owner: ContractAddress);
-    fn accept_ownership(ref self: TContractState);
-    fn renounce_ownership(ref self: TContractState);
+trait IOwnable<TState> {
```
augustbleeds, line 6: just verifying it's not necessary to change the name of this generic variable?

contracts/src/libraries/ownable.cairo
```diff
@@ -1,110 +1,155 @@
...
     use starknet::ContractAddress;
-    use starknet::contract_address_const;
-    use zeroable::Zeroable;
+    use starknet::get_caller_address;
 
     #[storage]
     struct Storage {
-        _owner: ContractAddress,
-        _proposed_owner: ContractAddress,
+        Ownable_owner: ContractAddress,
```
augustbleeds, line 21: nit: I think we can do without the `Ownable_` prefix

contracts/src/tests/test_aggregator.cairo
```diff
@@ -107,7 +107,7 @@ fn setup() -> (
 }
 
 #[test]
-#[available_gas(2000000)]
+#[available_gas(3000000)]
```
augustbleeds, line 110: reminder: we can get rid of all the gas macros in the tests

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -211,20 +264,21 @@ func (txm *starktxm) confirmLoop() {
...
-					if !ok {
-						txm.lggr.Errorw("wrong receipt type", "type", reflect.TypeOf(response))
+					response, err := client.Provider.GetTransactionStatus(ctx, f)
+					if err != nil {
+						txm.lggr.Errorw("failed to fetch transaction status", "hash", hash, "error", err)
 						continue
 					}
 
-					status := receipt.Status
+					status := response.FinalityStatus
```
augustbleeds, line 300: If it's REJECTED, then the tx won't be included in the block. So the nonce becomes wrong for all subsequent txs... therefore we need to updat them by recalculating the nonce and resubmit subsequent transactions with re-calculated nonce

augustbleeds (commented):
biggest item is about the txm's handling of REJECTED txs. Otherwise looks good!

.github/actions/install-cairo/action.yml
ivpavici: hi! maybe this could help:
https://github.com/cairo-nix/cairo-nix

relayer/pkg/chainlink/txm/txm.go
archseer: Can you add a task for this to the next sprint?

contracts/src/libraries/ownable.cairo
archseer: I copied this from OZ and figured I'd keep the naming so it'll be a smaller diff if we choose to migrate to their helper later. I can revert if needed

contracts/src/libraries/ownable.cairo
archseer: Yeah the name doesn't matter

contracts/src/tests/test_erc677.cairo
archseer: It's in the book! https://book.cairo-lang.org/ch99-01-05-03-testing-components.html#testing-components-without-deploying-a-contract

contracts/src/libraries/token/erc677.cairo
archseer: I'm not quite sure why but it's required to get the code compiling. There's some examples here https://github.com/ArielElp/component-examples

contracts/src/token/link_token.cairo
archseer: it'll flatten the subvariants into the parent event rather than having it nested

augustbleeds (commented):
image
Should we target 2.4.0 instead of 2.4.3 since it's not technically supported yet

contracts/src/ocr2/aggregator.cairo
```diff
@@ -292,20 +313,23 @@ mod Aggregator {
...
-        AccessControl::check_read_access(@state, caller);
+    #[generate_trait]
+    impl AccessHelperImpl of AccessHelperTrait {
+        fn _require_read_access(self: @ContractState) {
+            let caller = starknet::info::get_caller_address();
+            self.access_control.check_read_access(caller);
+        }
     }
 
     #[external(v0)]
```
augustbleeds, line 336: shouldn't this be `#[abi(embed_v0)]`? Goes for all the other contracts as well (ref: https://book.cairo-lang.org/ch99-01-03-02-contract-functions.html#2-public-functions)

archseer: good catch, `external(v0)` is being deprecated but seems to still work right now

archseer: looks like deprecation warning is added in 2.5.0

archseer:
> Should we target 2.4.0 instead of 2.4.3 since it's not technically supported yet

Minor versions only impact the compiler and can be used without changes. We can even upgrade to 2.5.0:

> [Cairo 2.5.0 16](https://github.com/starkware-libs/cairo/releases/tag/v2.5.0) was just released. This version only involves high-level compiler (Cairo→Sierra) changes and thus can be used to deploy contracts on Starknet Testnet and Mainnet without delay.

archseer:
I'm going to merge this branch into develop, it's at this point usable and has most tests passing as opposed to develop that's utterly broken.

PR #349, opened 2024-01-31
Fixing E2E tests

smickovskid:


ops/devnet/devnet.go
```diff
@@ -105,7 +105,7 @@ func (devnet *StarknetDevnetClient) FundAccounts(l2AccList []string) error {
 	for _, key := range l2AccList {
 		res, err := devnet.client.R().SetBody(map[string]any{
 			"address": key,
-			"amount":  1e21,
+			"amount":  9000000000000000000000.0,
```
archseer, line 108: btw this whole file could be replaced with starknet.go helpers: https://github.com/NethermindEth/starknet.go/blob/main/devnet/devnet.go#L122

archseer:
Merged into #343

PR #354, opened 2024-02-08
add nix to go test command in ci

tateexon:


archseer:
Assuming it's obsoleted by now

PR #356, opened 2024-02-14
Fix ci

smickovskid:
Fix ci


integration-tests/docker/test_env/stark.go
```diff
@@ -0,0 +1,105 @@
...
+		Str("containerName", s.ContainerName).
+		Msgf("Started Starknet container")
+
+	return nil
+}
+
+func (ms *Starknet) getContainerRequest() (*tc.ContainerRequest, error) {
+	return &tc.ContainerRequest{
+		Name:         ms.ContainerName,
+		Image:        "shardlabs/starknet-devnet-rs:latest",
```
archseer: We should pin this since latest can just start failing at some point

integration-tests/common/gauntlet_common.go
```diff
@@ -33,54 +32,60 @@ func (testState *Test) fundNodes() ([]string, error) {
...
+		// The starknet provided mint method does not work so we send a req directly
+		for _, key := range nAccounts {
+			res, err := m.TestConfig.Resty.R().SetBody(map[string]any{
+				"address": key,
+				"amount":  900000000000000000,
+			}).Post("/mint")
+			m.TestConfig.L.Info().Msg(fmt.Sprintf("Funding account: %s", string(res.Body())))
+			if err != nil {
+				return nil, err
+			}
```
archseer: Needs to also mint STRK now, but should be resolved after you rebase

integration-tests/infra_deployments/deployer_test.go
```diff
@@ -1,100 +1,101 @@
...
-	require.NoError(testState, err, "Deploying contracts should not fail")
-	t.SetBridgeTypeAttrs(&client.BridgeTypeAttributes{
-		Name: "bridge-coinmetrics",
-		URL:  "ADAPTER_URL", // ADAPTER_URL e.g https://adapters.main.sand.cldev.sh/coinmetrics
-	})
-
-	err = t.Common.CreateJobsForContract(t.Cc, observationSource, juelsPerFeeCoinSource, t.OCRAddr, t.AccountAddresses)
-	require.NoError(testState, err)
-}
+//
```
archseer: Can we just delete this file?

integration-tests/common/test_common.go
```diff
@@ -3,135 +3,237 @@ package common
...
-		require.NoError(testState.T, err, "Setting account address should not fail")
-		testState.Devnet.AutoDumpState() // Auto dumping devnet state to avoid losing contracts on crash
+		m.Account.Account = account.Address
+		m.Account.PrivateKey = account.PrivateKey
 	}
 }
 
 // DeployEnv Deploys the environment
-func (testState *Test) DeployEnv() {
-	testState.Common.SetLocalEnvironment(testState.T)
```
archseer, line 102: If this replaces the scripts I built then let's delete them from `scripts/` and update the integration-tests readme so it's clear what the dependencies are and how to run this

integration-tests/common/common.go
```diff
@@ -18,39 +22,139 @@ import (
...
-	serviceKeyL1        = "Hardhat"
-	serviceKeyL2        = "chainlink-starknet.starknet-devnet"
-	serviceKeyChainlink = "chainlink"
-	chainName           = "starknet"
-	chainId             = "SN_GOERLI"
-	defaultNodeUrl      = "http://127.0.0.1:5050"
+	chainName            = "starknet"
+	chainId              = "SN_GOERLI"
+	DefaultNodeCount     = 5
+	DefaultL2RPCInternal = "http://starknet-dev:5000"
```
archseer: This probably won't work for soak?

integration-tests/docker/test_env/stark.go
archseer: Use the same commit we use under `scripts/`

integration-tests/testconfig/ocr2/example.toml
```diff
@@ -0,0 +1,96 @@
+# Example of full config with all fields
```
archseer: Why two configs?

ops/gauntlet/gauntlet_starknet.go
```diff
@@ -88,6 +90,10 @@ func (sg *StarknetGauntlet) InstallDependencies() error {
 	if err != nil {
 		return err
 	}
+	_, err = sg.G.ExecCommand([]string{"build"}, *sg.options)
```
archseer: What does this do, `yarn build`?

archseer:
```
  09:15:06.28 INF gauntlet stderr="unknown error: Error: ENOENT: no such file or directory, open '/home/runner/work/chainlink-starknet/chainlink-starknet/packages-ts/starknet-gauntlet-oz/dist/lib/../../../../contracts/target/release/chainlink_Account.contract_class.json'\n"
```

Test failure looks like it's missing `scarb --profile release build` before running gauntlet

integration-tests/common/gauntlet_common.go
smickovskid: Accepted remote changes

integration-tests/testconfig/configs_embed.go
```diff
@@ -0,0 +1,21 @@
...
+//go:embed log_poller/log_poller.toml
+//go:embed node/node.toml
+//go:embed ocr/ocr.toml
+//go:embed vrfv2/vrfv2.toml
+//go:embed vrfv2plus/vrfv2plus.toml
+var embeddedConfigsFs embed.FS
+
+func init() {
+	areConfigsEmbedded = true
+}
```
Tofel, line 14: you don't need that embedded and noembed configs if you are not planning to compile the tests to a binary like we do in core in some workflows.

if you do need it, though, you should remove all embeds but these two:
```
//go:embed default.toml
//go:embed ocr2/ocr2.toml
```

Because other files do not exist and that will make compilation fail.

integration-tests/testconfig/testconfig.go
```diff
@@ -0,0 +1,360 @@
...
+	GetCommonConfig() *Common
+}
+
+type Ocr2TestConfig interface {
+	GetOCR2Config() *ocr2_config.Config
+}
+
+type NamedConfiguration interface {
+	GetConfigurationName() string
+}
```
Tofel: curious: are you using any of these interfaces anywhere? if not, delete them. I added them to core for a very specific reason. @gheorghestrimtu was using some of the helper functions from `core`, but he wanted to use his own `TestConfig` struct with them. That's why I defined the interfaces in `core`, so that his `TestConfig` could implement them.

So even if you need to call functions from `core` you should implement interfaces defined there instead of declaring new ones.

Unless I'm missing something?

integration-tests/testconfig/testconfig.go
```diff
@@ -0,0 +1,360 @@
...
+	return nil
+}
+
+type Product string
+
+const (
+	OCR2 Product = "ocr2"
+)
+
+var TestTypesWithLoki = []string{"Soak", "Smoke"}
```
Tofel: pls delete, I'm removing it from `core` now.

integration-tests/testconfig/testconfig.go
```diff
@@ -0,0 +1,360 @@
...
+const TestTypeEnvVarName = "TEST_TYPE"
+
+func GetConfigurationNameFromEnv() (string, error) {
+	testType := os.Getenv(TestTypeEnvVarName)
+	if testType == "" {
+		return "", fmt.Errorf("%s env var not set", TestTypeEnvVarName)
+	}
+
+	return cases.Title(language.English, cases.NoLower).String(testType), nil
+}
```
Tofel: do you need that? it's only required if you need to select configuration depending on test type

smickovskid: Removed

integration-tests/testconfig/testconfig.go
smickovskid: Removed

integration-tests/testconfig/testconfig.go
smickovskid: Removed

integration-tests/testconfig/configs_embed.go
smickovskid: Kept it in case we need it in the future

.github/workflows/integration-tests-smoke.yml
```diff
@@ -147,9 +149,39 @@ jobs:
...
         with:
           nix_path: nixpkgs=channel:nixos-unstable
+      - name: Setup Scarb for Linux
+        id: install-scarb
+        shell: bash
+        run: |
+          wget https://github.com/software-mansion/scarb/releases/download/${{ env.scarb_version }}/scarb-${{ env.scarb_version }}-x86_64-unknown-linux-musl.tar.gz
+          tar -xvzf scarb-${{ env.scarb_version }}-x86_64-unknown-linux-musl.tar.gz
+          mv -vf scarb-${{ env.scarb_version }}-x86_64-unknown-linux-musl scarb-build
+          echo "$GITHUB_WORKSPACE/scarb-build/bin" >> $GITHUB_PATH
```
archseer: Why not just use the action we use on all the other workflows

https://github.com/smartcontractkit/chainlink-starknet/blob/57cedea9c926a041ab341839311cc3dac57362c1/.github/workflows/contracts.yml#L41-L42

integration-tests/common/gauntlet_common.go
```diff
@@ -4,25 +4,24 @@ import (
...
-	"os"
-
 	"github.com/smartcontractkit/chainlink-starknet/integration-tests/utils"
+	"os"
 )
 
 var (
-	ethAddressGoerli = "0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7"
-	nAccount         string
+	strkAddressSepolia = "0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d"
```
archseer: is this only used on testnet?

archseer: Can we group this and `chainIdTestnet` (+ maybe RPC) in a config struct somewhere? These constants are very brittle and split across several files

integration-tests/docker/test_env/stark.go
```diff
@@ -0,0 +1,105 @@
...
+		Str("containerName", s.ContainerName).
+		Msgf("Started Starknet container")
+
+	return nil
+}
+
+func (s *Starknet) getContainerRequest() (*tc.ContainerRequest, error) {
+	return &tc.ContainerRequest{
+		Name:         s.ContainerName,
+		Image:        "shardlabs/starknet-devnet-rs:b41e566a3f17aa0e51871f02d5165959e50ce358",
```
archseer: this too, it would be nice if it was pulled out in a constants file since it has to be kept up to date

ops/gauntlet/gauntlet_starknet.go
archseer: My question still stands here

contracts/.tool-versions
```diff
@@ -0,0 +1 @@
+scarb 2.5.4
```
archseer: There's already a `.tool-versions` in the root directory that specifies this

.github/workflows/integration-tests-smoke.yml
```diff
@@ -125,7 +125,9 @@ jobs:
       TEST_DURATION: 15m
       NODE_COUNT: 5
       CHAINLINK_IMAGE: ${{ secrets.QA_AWS_ACCOUNT_NUMBER }}.dkr.ecr.${{ secrets.QA_AWS_REGION }}.amazonaws.com/chainlink
-      CHAINLINK_VERSION: starknet.${{ github.sha }}${{ matrix.image.tag-suffix }}
+      CHAINLINK_VERSION: starknet.${{ github.sha }}
```
cfal: is the matrix suffix not necessary now? the push_tag on L69 still push with a `matrix.image.tag-suffix`?

.github/workflows/integration-tests-smoke.yml
```diff
@@ -125,7 +125,9 @@ jobs:
...
+      INTERNAL_DOCKER_REPO: ${{ secrets.QA_AWS_ACCOUNT_NUMBER }}.dkr.ecr.${{ secrets.QA_AWS_REGION }}.amazonaws.com
+      scarb_version: "v2.5.4"
```
cfal: +1 to using the install-cairo action, then we don't need `scarb_version` in multiple places

integration-tests/.root_dir
cfal, line 1: what does this file do?

integration-tests/common/gauntlet_common.go
```diff
@@ -33,54 +32,69 @@ func (testState *Test) fundNodes() ([]string, error) {
...
 			if err != nil {
 				return nil, err
 			}
 		}
-
 	} else {
-		err = testState.Devnet.FundAccounts(nAccounts)
-		if err != nil {
-			return nil, err
+		// The starknet provided mint method does not work so we send a req directly
```
cfal, line 35: can we delete ops/devnet.go if we're switching to nethermind's devnet client?

integration-tests/smoke/ocr2_test.go
```diff
@@ -3,59 +3,87 @@ package smoke_test
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		err = os.Setenv("CHAINLINK_ENV_USER", *config.Common.User)
+		require.NoError(t, err, "Could not set CHAINLINK_ENV_USER")
+		err = os.Setenv("INTERNAL_DOCKER_REPO", *config.Common.InternalDockerRepo)
+		require.NoError(t, err, "Could not set INTERNAL_DOCKER_REPO")
+		test := test
+		if test.name == "plugins" {
+			pluginImage := *config.ChainlinkImage.Image + "-plugins"
```
cfal: not familiar with this code - but could this clause be removed if we readded the tag suffix to the expected image name in the yaml?

integration-tests/smoke/ocr2_test.go
```diff
@@ -3,59 +3,87 @@ package smoke_test
...
-	err = testState.DeployGauntlet(0, 100000000000, decimals, "auto", 1, 1)
-	require.NoError(t, err, "Deploying contracts should not fail")
-	if !testState.Common.Testnet {
-		testState.Devnet.AutoLoadState(testState.OCR2Client, testState.OCRAddr)
-	}
-	testState.SetUpNodes()
+			// K8s specific config and cleanup
+			if *config.Common.InsideK8s {
+				t.Cleanup(func() {
+					if err = actions.TeardownSuite(t, state.Common.Env, state.ChainlinkNodesK8s, nil, zapcore.PanicLevel, nil); err != nil {
```
cfal, line 58: previously we called `TearDownLocalEnvironment` for the docker setup - i don't see a teardown call for non-k8s now, does that still occur?

integration-tests/common/common.go
```diff
@@ -378,7 +284,7 @@ func (c *Common) CreateJobsForContract(cc *ChainlinkClient, mockUrl string, obse
 		JobType:        "bootstrap",
 		OCR2OracleSpec: oracleSpec,
 	}
-
+	fmt.Println(jobSpec.String())
```
cfal: log?

cfal (commented):
do we need to maintain both non-k8s and k8s? it seems to complicate the code a bit, curious if we couldn't do this only with the docker environment, and connecting to the testnet from it if necessary.

integration-tests/.root_dir
smickovskid: This is related to the testconfig and TOML changes, that is how the logic knows where to look for the toml configurations.

integration-tests/smoke/ocr2_test.go
smickovskid: Not needed for docker, when the parent process terminates all containers will do as well

integration-tests/smoke/ocr2_test.go
smickovskid: I assume you are referring to CI, for local we would still need to override if we want to run it

.github/workflows/integration-tests-smoke.yml
smickovskid: Yes we need to maintain both if we want to run long running soak tests, if we only use docker there is no reliable way to run our tests for a longer period of time.

integration-tests/common/gauntlet_common.go
smickovskid: Grouped under `integration-tests/config/config.go`

integration-tests/docker/test_env/stark.go
smickovskid: Added as mandatory var in the toml config and value in `default.toml`

contracts/.tool-versions
smickovskid: Both removed

integration-tests/common/gauntlet_common.go
smickovskid: Removed

integration-tests/common/common.go
smickovskid: Removed

integration-tests/smoke/ocr2_test.go
cfal: just to clarify - if i'm running this locally, and i already have docker running on my machine with unrelated containers, does it clean up these terst containers that are started here?

cfal (commented):
looks good, left one more comment on docker cleanup. also, go relayer lint is failing (perhaps caused by my fix 😅)

PR #357, opened 2024-02-14
Working local soaktests

augustbleeds:


integration-tests/common/common.go
```diff
@@ -157,12 +159,79 @@ HTTPSPort = 0
 	c.ChainlinkConfig = chainlinkConfig
 	log.Debug().Str("toml", chainlinkConfig).Msg("Created chainlink config")
 
-	envConfig := &environment.Config{NamespacePrefix: "chainlink-ocr-starknet", TTL: c.TTL, Test: t}
-
-	c.Env = environment.New(envConfig)
```
augustbleeds, line 162: this relies on k3d and k8s

integration-tests/common/common.go
```diff
@@ -157,12 +159,79 @@ HTTPSPort = 0
...
+	w.buf.Write(d)
+	return w.w.Write(d)
+}
+
+// Bytes returns bytes written to the writer
+func (w *CapturingPassThroughWriter) Bytes() []byte {
+	return w.buf.Bytes()
+}
+
+func debug(cmd *exec.Cmd) error {
```
augustbleeds, line 190: made debugging a lot easier. I can remove and replace with qa helper method, but it'd be nice to have something like this to debug on CI and locally

scripts/core.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -euo pipefail
+set -euox pipefail
```
augustbleeds, line 3: personally, I prefer to have as much info

PR #359, opened 2024-02-14
[BCI 2027]: Adds multisig cairo tests

chris-de-leon-cll:
This PR adds some new test cases for the following functions:

1. `set_threshold`
2. `set_signers`
3. `set_signers_and_threshold`

For each function above, there are 2 new corresponding test cases: a recursive version and a non-recursive version.

In the recursive version, a transaction is created which calls one of the Multisig functions above. This transaction is submitted to the Multisig contract, which is then confirmed by the correct number of signers and executed. The test case verifies the state before and after the transaction execution.

In the non-recursive version, the Multisig contract is deployed, and the methods on it are directly called using a dispatcher. There is no transaction submission/confirmation. The test simply calls a function and verifies that the result is correct.
I believe the recursive version is what the ticket [here](https://smartcontract-it.atlassian.net/browse/BCI-2027) is referring to, but I included both versions for completeness.


archseer:
Can you also drop `packages-ts/integration-multisig`? These changes should now obsolete them :)

Also delete the mention in the makefile: https://github.com/smartcontractkit/chainlink-starknet/blob/1272736c2ac03bd0080aa01f3381e1553f0a2202/Makefile#L224-L225

augustbleeds (approved):

chris-de-leon-cll:
@archseer I've deleted the `packages-ts/integration-multisig` folder and updated the Makefile accordingly - feel free to approve and merge if everything else looks good!

archseer (approved):
LGTM, thanks for working on this!

PR #360, opened 2024-02-20
Sonar bump

smickovskid:
- Sonar bump

### DIFF
--- a/.github/workflows/sonar-scan.yml
+++ b/.github/workflows/sonar-scan.yml
@@ -101,7 +101,7 @@ jobs:
         run: sed -i 's+/home/runner/work/feeds-manager/feeds-manager/+/github/workspace/+g' ${{ steps.sonarqube_report_paths.outputs.sonarqube_eslint_report_paths }}
 
       - name: SonarQube Scan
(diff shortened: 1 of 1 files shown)


.github/workflows/sonar-scan.yml
```diff
@@ -101,7 +101,7 @@ jobs:
         run: sed -i 's+/home/runner/work/feeds-manager/feeds-manager/+/github/workspace/+g' ${{ steps.sonarqube_report_paths.outputs.sonarqube_eslint_report_paths }}
 
       - name: SonarQube Scan
-        uses: sonarsource/sonarqube-scan-action@a6ba0aafc293e03de5437af7edbc97f7d3ebc91a # v1.2.0
+        uses: sonarsource/sonarqube-scan-action@2.0.1
```
archseer: I think for security reasons we're always supposed to pin to a sha rather than a tag

archseer:
```suggestion
        uses: sonarsource/sonarqube-scan-action@69c1a75940dec6249b86dace6b630d3a2ae9d2a7 # v2.0.1
```

PR #363, opened 2024-02-23
update contracts and tests for cairo 2.5.x

augustbleeds:
Used cairo 2.5.4, the last version of 2.5.x

Changes include
* using abi_embed(v0) macro. All public functions must be inside of an implementation of a trait with the said macro.
* using OwnableTwoStepImpl instead of our own. Only difference between the two is `pending_owner` is used instead of `proposed_owner` and there is no 0 address check during a pending_owner proposal. This is fine because there will be never anyone who can confirm the ownership (no one will be 0 address) so the original owner will still be the owner and can try again.
* fixed some of the tests that were broken. 
* got rid of most compiler warnings
* changed loop {} to while {} where appropriate


chris-de-leon-cll (dismissed):

contracts/src/account.cairo
```diff
@@ -1,57 +1,233 @@
-// copied from https://github.com/OpenZeppelin/cairo-contracts/blob/v0.8.1/src/presets/account.cairo
+// copied from https://github.com/OpenZeppelin/cairo-contracts/blob/release-v0.9.0/src/access/ownable/ownable.cairo
```
archseer: Let's use a sha since tags can change

contracts/src/libraries.cairo
```diff
@@ -1,4 +1,5 @@
 mod ownable;
+mod zownable;
```
archseer: This was added together with an empty file?

contracts/src/ocr2/aggregator.cairo
```diff
@@ -107,6 +107,9 @@ trait Billing<TContractState> {
     fn link_available_for_payment(
         self: @TContractState
     ) -> (bool, u128); // (is negative, absolute difference)
+    fn set_link_token(
```
archseer, line 110: good catch 👍🏻

contracts/src/tests/test_multisig.cairo
```diff
@@ -697,23 +666,19 @@ fn test_recursive_set_signers_and_threshold() {
         );
 
     // Signer 1 confirms the transaction
-    set_caller_address(s1);
     set_contract_address(s1);
     multisig.confirm_transaction(0);
 
     // Signer 2 confirms the transaction
-    set_caller_address(s2);
```
archseer, line 705: Hmm was this API removed?

contracts/src/libraries.cairo
```diff
@@ -1,4 +1,5 @@
 mod ownable;
```
archseer: Let's drop our ownable module if it's now obsolete

contracts/src/token/link_token.cairo
```diff
@@ -102,7 +103,7 @@ mod LinkToken {
         }
```
archseer, line 102: There's a TODO to expose the minter method a couple lines above, can we take care of that here too?

archseer (commented):
Some small nits, great work otherwise!

contracts/src/tests/test_multisig.cairo
augustbleeds: nah, it wasn't removed. we weren't using it the correct way. set_contract_address should only be used if the test contract is deployed in the testing suite. Otherwise we should use set_caller_address.

I noticed when testing that there is some undefined behavior if you do both

contracts/src/libraries.cairo
augustbleeds: whoops, i forgot to remove that

archseer (approved):

archseer (changes requested):
Lints and tests still fail. I'll quickly resolve

contracts/src/account.cairo
```diff
@@ -1,57 +1,233 @@
...
 
 // SPDX-License-Identifier: MIT
-// OpenZeppelin Contracts for Cairo v0.8.1 (presets/account.cairo)
+// OpenZeppelin Contracts for Cairo v0.9.0 (access/ownable/ownable.cairo)
 
-/// # Account Preset
+/// # Ownable Component
```
archseer: You seem to have replaced the account contract with the ownable contract?

archseer (changes requested):

PR #364, opened 2024-02-28
[BCI-2538]: Updates Examples Directory

chris-de-leon-cll:
## Changes

- [x] removes `proxy-consumer`
- [x] adds a new folder `aggregator_consumer` with `starknet-foundry` integrated
- [x] removes the original `aggregator-consumer` example folder (keeping this around for reference - will remove it once everything has been updated)
- [x] removes typescript, node.js, hardhat, starknet.js, etc.
- [x] updates the original `consumer.cario` and `price_consumer.cairo` contracts for compatibility with cairo v2.5.4
- [x] adds new cairo tests for `consumer.cairo` and `price_consumer.cairo`- every function on both of these contracts has a corresponding test case
- [x] updates the README
	- [x] adds a section on what packages/software needs to be installed and includes relevant links
	- [x] adds section on what command to use for running tests via snforge (local devnet)
- [x] converts the original typescript scripts to cairo tests that can be called via `snforge` (so that users can see how the contracts work via a local devnet)
	- [x] `consumerValidator.ts` - not applicable (this doesn't use `consumer.cairo` or `price_consumer.cairo`)
	- [x] `deploy_accounts.ts` - not applicable (handled programmatically in cairo test cases)
	- [x] `deploy_contracts.ts` - the test cases have helper functions which use the snforge library to compile and deploy contracts
	- [x] `readContinuously.ts` - not applicable (this runs an infinite loop that reads the latest round data every 3 seconds)
	- [x] `readDecimals.ts` - has a corresponding cairo test case
	- [x] `readLatestRound.ts` - has a corresponding cairo test case
	- [x] `updateLatestRound.ts` - has a corresponding cairo test case
- [x] update CI pipelines and github actions to automatically run the test cases
	- [x] adds a new github action (`.github/install-starknet-foundry/action.yml`) that installs `starknet-foundry`
	- [x] adds a new github action (`.github/workflow/examples.yml`) that runs the `aggregator_consumer` example test cases via `snforge test`
	- [x] removes `packages-ts/starknet`
	- [x] removes Makefile command `test-integration-contracts`

## Delayed (will be tackled in a follow up PR)
- [ ] update the README
	- [ ] add a section on how to use sncast scripts (live testnet)
- [ ] convert the original typescript scripts to `sncast` scripts that interact with starknet testnet (so that users can see how the contracts work against a live network)
	- [ ] `consumerValidator.ts`
	- [ ] `deploy_accounts.ts`
	- [ ] `deploy_contracts.ts`
	- [ ] `readContinuously.ts`
	- [ ] `readDecimals.ts`
	- [ ] `readLatestRound.ts`
	- [ ] `updateLatestRound.ts`

## Known Issues / Limitations
- Null address checks of the form `address.is_zero()` have been commented out due to compile errors - looking into why these are occurring even when the `zeroable` module is imported
	- <i>**This has been resolved**</i>. The solution was to remove the `edition` property from `Scarb.toml`, which was automatically added by `snforge init`:
		```toml
    	# Scarb.toml

	    # ...

    	[package]
		edition = 2023_11

		# ...
    	```
- Some of the contracts from the `chainlink-starknet/contracts` folder have been copied into the examples directory and modified slightly for compatibility with `snforge`. All the modifications are minor - they mostly involved creating new interfaces for dispatchers and renaming import paths. Also, to elaborate more on "compatibility for `snforge`", the test cases use a function `declare(<ContractName>)` for contract deployment, and this function only works if <ContractName> references a contract that is exported from the `lib.cairo` file in the `src` folder. I'm curious to understand if there's a better way to reference the original contracts without having to duplicate them in the `examples` folder
	- <i>**This has been resolved**</i>. The solution was to remove the `edition` field from `Scarb.toml` (see above) and to also add the following to `Scarb.toml` so that [the tests could find the contract artifacts](https://docs.swmansion.com/scarb/docs/extensions/starknet/contract-target.html#compiling-external-contracts):
		```toml
		# Scarb.toml
		
		# ...

		[[target.starknet-contract]]
		build-external-contracts = [
			"chainlink::emergency::sequencer_uptime_feed::SequencerUptimeFeed",
			"chainlink::ocr2::mocks::mock_aggregator::MockAggregator",
		]
		
		# ...
		```


Makefile
```diff
@@ -224,6 +224,11 @@ test-integration-contracts: build-ts env-devnet-hardhat
 	cd packages-ts/starknet/ && \
 		yarn test
 
+.PHONY test-examples
```
archseer:
```suggestion
.PHONY: test-examples
```

archseer (approved):
Nice work!

PR #365, opened 2024-03-01
Nonce Error Handling

augustbleeds:
Summary:

* Handle invalid nonce Error during broadcast and confirmation by clearing the txstore of all unconfirmed txs and resubmitting the txs with new nonces. During broadcast invalid nonce error can happen during fee estimation or tx invocation.
* Have to use the centralized feeder gateway client to get the status of a rejected transaction (in order to determine if we should retry txs with new nonces) because that is the only way to do so today (asked Starkware)
* You must now provide the FeederURL on all environments besides testnet and sepolia testnet
* Added an exponential backoff timer to prevent handling nonce error too quickly

Sepolia Testnet: "https://alpha-sepolia.starknet.io/feeder_gateway"
Mainnet: "https://alpha-mainnet.starknet.io/feeder_gateway"
Integration: "https://external.integration.starknet.io/feeder_gateway"
Sepolia Integration: "https://integration-sepolia.starknet.io/feeder_gateway"
Goerli Testnet: "https://alpha4.starknet.io/feeder_gateway"
Goerli Integration: "https://external.integration.starknet.io"

Some less related changes
* If you use a nethermind node, they set the x-apikey header you you have to add `APIKey = '<>'` in the Node TOML config
* Removed chain id for nonce manager since there's only 1 nonce manager per chain


relayer/pkg/chainlink/txm/nonce.go
```diff
@@ -64,6 +64,23 @@ func (nm *nonceManager) HealthReport() map[string]error {
...
+	}
+	nm.lock.Lock()
+	defer nm.lock.Unlock()
+
+	n, err := client.AccountNonce(ctx, address)
+	if err != nil {
+		return err
+	}
+
+	nm.n[address.String()][chainId] = n
```
cfal: could we remove chainId? perhaps it was possible before with dynamic configs, but with TOML configs, could the chainId still possibly change once we've initialized the relayer? or was there another reason we needed it? cc @aalu1418

relayer/pkg/chainlink/txm/nonce.go
```diff
@@ -64,6 +64,23 @@ func (nm *nonceManager) HealthReport() map[string]error {
 	return map[string]error{nm.Name(): nm.starter.Healthy()}
 }
 
+func (nm *nonceManager) Sync(ctx context.Context, address *felt.Felt, chainId string, client NonceManagerClient) error {
+	if err := nm.validate(address, chainId); err != nil {
```
cfal: nit: we could avoid locking/unlocking twice, eg by refactoring and adding a `validateLocked` function which requires the lock to be held as a precondition

relayer/pkg/chainlink/txm/txstore.go
```diff
@@ -57,8 +66,11 @@ func (s *TxStore) Confirm(hash string) error {
 	defer s.lock.Unlock()
 
 	if nonce, exists := s.hashToNonce[hash]; exists {
-		delete(s.hashToNonce, hash)
 		delete(s.nonceToHash, nonce)
+
```
cfal, line 80: i'm curious about the logic in `Confirm` and `Save`. `Save()` will keep incrementing `currentNonce` by 1 until it finds a nonce that doesn't exist in the map, while `Confirm` will delete confirmed nonces from the same map. looking at our `confirmLoop` in txm.go, where we iterate through `GetAllUnconfirmed`, isn't it possible that we check for a newer nonce to be confirmed first? because a) `maps.Values` (which `GetAllUnconfirmed` relies on) doesn't provide any guarantees on ordering, and b) because even if the return values were ordered, we could see transaction with nonce=1 as unconfirmed, and then immediately transaction with nonce=2 as confirmed eg. when the node confirms both at once between the RPC calls

cc @aalu1418

relayer/pkg/chainlink/txm/nonce.go
aalu1418: if i remember correctly, it's not that chainID can change - it's that you could theoretically have a single CL node connected to multiple chains

this was similar to the EVM nonce management - but I don't know if that changes with LOOP? if there's a new plugin on a per chain basis or a per chain family basis. if it's a per chain basis, then the chainID could probably be removed 🤔

relayer/pkg/chainlink/txm/txstore.go
aalu1418: correct - ordering it would help or could also assume that if a higher nonce is confirmed, lower nonces must also have been confirmed (like how EVM does it - `confirmed without receipt`)

cfal: have we seen this error @augustbleeds ? it seems like `Save()` could error out without a fix, perhaps we can fix it as part of this PR?

relayer/pkg/chainlink/txm/nonce.go
augustbleeds: yeah, i noticed that too! each nonce manager is tied to one txm and there's only one txm per chain. So I think it's safe to remove the chain id.

relayer/pkg/chainlink/txm/nonce.go
augustbleeds: oh nice, i didn't realize validate was also locking. i'll leave for now but good catch

relayer/pkg/chainlink/txm/txstore.go
augustbleeds: I haven't seen this happen in the logs of the soak test. tracking the highest nonce seems to be easier. I can do that

augustbleeds: Actually, I'm not following. So is the issue is that nonce=2 could be confirmed before nonce=1 is confirmed? That can definitely happen, but that wouldn't break any logic of Save() because that higher nonce had to have been Saved() first in the broadcast so even though the nonce is gone from the store the store's currentNonce has to be higher than the confirmed tx

augustbleeds: **t=0 (add two txs)**

tx_store:
{nonce-1, nonce-2}

currNonce:
3

**t=1 (later tx is confirmed first)**

tx_store:
{nonce-1}

currNonce:
3

**t=2 (broadcast new tx)**

tx_store:
{nonce-1, nonce-3}

currNonce:
4

augustbleeds: ~oh i think i see what you mean, the Save() logic doesn't save currentNonce based off of the nonce it was given~ I think that because, at the beginning of t=2, the currentNonce is 3, the most recently stored nonce will be in the map so the currentNonce will be updated to 4. So I don't think we have this problem

augustbleeds: it doesn't start counting off at nonce=1. It starts counting off at the highest unconfirmed nonce ever saved in the txm. So the fact that a nonce was confirmed doesn't remove the tx store's memory that currentNonce is higher than any of the unconfirmed nonces seen

relayer/pkg/starknet/feeder_client.go
```diff
@@ -0,0 +1,211 @@
+package starknet
```
augustbleeds, line 1: most code was repurposed from juno (they didn't support some of the response fields)

relayer/pkg/starknet/feeder_client.go
```diff
@@ -0,0 +1,211 @@
...
+	url        string
+	client     *http.Client
+	backoff    Backoff
+	maxRetries int
+	maxWait    time.Duration
+	minWait    time.Duration
+	log        utils.SimpleLogger
+}
+
+func (c *FeederClient) WithBackoff(b Backoff) *FeederClient {
```
cfal, line 40: nit: imo `WithXX` seem to imply that you're going to build a new FeederClient with that setting and not just modify the current one, ie you typically see it in factory-type classes that are going to end up building an instance of something, i'd think `SetXX` is clearer

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -302,13 +370,42 @@ func (txm *starktxm) confirmLoop() {
...
+
+						rejectedTx, err := feederClient.TransactionFailure(ctx, f)
+						if err != nil {
+							txm.lggr.Errorw("failed to fetch reason for transaction failure", "hash", hash, err)
+							continue
+						}
+
+						if isInvalidNonce(rejectedTx.ErrorMessage) {
+							// resubmits all unconfirmed transactions (includes the current one that just failed)
+							txm.handleNonceErr(ctx, addr)
```
cfal: do we need to handle when this returns an error?

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -302,13 +370,42 @@ func (txm *starktxm) confirmLoop() {
...
+							break
+						}
+
+						rejectedTx, err := feederClient.TransactionFailure(ctx, f)
+						if err != nil {
+							txm.lggr.Errorw("failed to fetch reason for transaction failure", "hash", hash, err)
+							continue
+						}
+
+						if isInvalidNonce(rejectedTx.ErrorMessage) {
```
cfal, line 487: do we need to handle the case where it's not an invalid nonce error?

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -125,7 +129,52 @@ func (txm *starktxm) broadcastLoop() {
...
+	}
+
+	return nil
+}
+
 const FEE_MARGIN uint32 = 115
+const BROADCASTER_NONCE_ERR = "Invalid transaction nonce"
+const CONFIRMER_NONCE_ERR = "InvalidNonce"
+
+func isInvalidNonce(err string) bool {
```
cfal, line 273: isn't this grouping different expected responses from different RPCs? could this be separated into one for the broadcaster error, and one for the feeder gateway string?

relayer/pkg/chainlink/txm/nonce.go
cfal: feeling like maybe this isn't a nit - could we handle this so it doesn't lock twice?

relayer/pkg/starknet/feeder_client.go
```diff
@@ -0,0 +1,211 @@
...
+
+	txStatus := new(TransactionFailureReason)
+	if err = json.NewDecoder(body).Decode(txStatus); err != nil {
+		return nil, err
+	}
+	return txStatus, nil
+}
+
+// Only responds on valid /get_transaction?transactionHash=<TRANSACTION_HASH> requests. fails otherwise
+func NewTestServer() *httptest.Server {
```
cfal, line 164: can these be moved to the test file?

relayer/pkg/starknet/feeder_client.go
```diff
@@ -0,0 +1,211 @@
...
+		defaultErr := []byte(`{
+			"code": "SOME_ERROR",
+			"error_message": "some error was encountered"
+		}`)
+
+		w.Write(defaultErr)
+	}))
+}
+
+// NewTestClient returns a client and a function to close a test server.
```
cfal, line 189: nit: mention it starts up a server?

relayer/pkg/chainlink/txm/nonce.go
cfal: if we're planning on cleaning this up, might be nice to do it in this PR since we'd be adding more code that refers to `chainId` otherwise

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -125,7 +129,52 @@ func (txm *starktxm) broadcastLoop() {
...
+	txm.nonce.Sync(ctx, accountAddress, chainId, client)
+
+	// todo: in the future, revisit resetting txm.txStore.currentNonce
+
+	unconfirmedTxs, err := txm.txStore.GetUnconfirmedSorted(accountAddress)
+	if err != nil {
+		return err
+	}
+
+	// delete all unconfirmed txs and resubmit them to the txm queue
```
cfal, line 254: since we know both the hash and the nonce of the tx, would it make sense to onlysubmit those with unconfirmed larger nonces? in fact, should we be doing that for any error, nonce or otherwise?

relayer/pkg/chainlink/txm/nonce.go
augustbleeds: will do!

relayer/pkg/chainlink/txm/txm.go
augustbleeds: good catch! i found out the hard way when testing that my errors weren't propagating

relayer/pkg/chainlink/txm/txm.go
augustbleeds: nope! at least not for now. i think in the future we can think about rebroadcasting. but part of having the feeder client gives us the ability to log and analyze what errors we can mitigate in the future

relayer/pkg/starknet/feeder_client.go
augustbleeds: yeah, it started out in the feeder_client_test file. but then i had to put it here because i needed to import it in the integration pkg for integration tests. Go doesn't import exported variables from *_test.go files outside of the package.

relayer/pkg/chainlink/txm/txm.go
augustbleeds: i had similar thoughts, but then realized that you could have tx4 and tx5 both having nonce issues but you process tx5 first. So potentially all txs need to be retried. 

If there were another tx3 in the queue which was completly fine and was accepted, it's ok to retry that one as well because it'll just fail on retry

relayer/pkg/chainlink/txm/txm.go
cfal: why don't we handle it now? another question i guess is, why not just skip handling InvalidNonce as well then?

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -302,13 +401,53 @@ func (txm *starktxm) confirmLoop() {
...
+
 					if err != nil {
 						txm.lggr.Errorw("failed to fetch transaction status", "hash", hash, "error", err)
 						continue
 					}
 
 					status := response.FinalityStatus
 
+					// currently, feeder client is only way to get rejected reason
+					if status == starknetrpc.TxnStatus_Rejected {
```
cfal, line 472: what's FinalityStatus when a transaction is reverted? `response.ExecutionStatus` is `REVERTED`, i'm curious if it hits this clause as well, and if we can't skip hitting the feeder client if so

relayer/pkg/chainlink/txm/txm.go
augustbleeds: that's a good question. originally, this problem had a ticket for it because starknet made some changes to tx life cycle management and as a result the txmgr could enter a state with an out of sync nonce.

i guess there is a potential optional where we just resync the nonce manager without rebroadcasting. basically, resync but don't retry txs. it would be a heck of a lot simpler. but the counterargument is we already have the solution to rebroadcast for nonces.

so tradeoff is simplicity with less robustness over complexity with robustness. 

i think it's worth discussing @archseer @cfal @aalu1418

relayer/pkg/chainlink/txm/txm.go
augustbleeds: finality status would be accepted_on_l2 (or l1)

relayer/pkg/chainlink/txm/txm.go
aalu1418: i would hesitate to allow the txm to just "drop" a tx - that behavior has only been limited to solana data feeds because there are no nonces + continuous/consistent updates + broadcasting schedule

typical data feeds only report on deviation changes so that makes broadcasting individuals reports is more critical since there may not be another for some time. also DONs usually only have 1 node broadcast at a time which is a cost saving method assuming a single tx can reliably included on chain

relayer/pkg/chainlink/txm/nonce.go
augustbleeds: actually, i think it is better to leave it as is because confirmer and broadcaster both access the nonce manager concurrently. Therefore, even if you check that the lock is held in a hypothetical "validateLocked", you don't know whether the lock was held by the caller of "validateLocked" or by another goroutine. 

If it's by another goroutine, then we could be reading the lock when the other goroutine is changing it.

augustbleeds: i could just always call validate() within the bounds of the caller's lock. maybe that's what you were referring to all along

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -125,7 +197,82 @@ func (txm *starktxm) broadcastLoop() {
...
+
+	return nil
+}
+
+func (txm *starktxm) handleNonceErr(ctx context.Context, accountAddress *felt.Felt, publicKey *felt.Felt) error {
+
+	txm.lggr.Debugw("Handling Nonce Validation Error By Resubmitting Txs...", "account", accountAddress)
+
+	// exponential backoff wait for rpc starknet_estimateFee to catch up with nonce returned by starknet_getNonce
+	txm.nonceRetryTimer.Wait(ctx)
```
augustbleeds, line 226: look at avoiding waiting for nonce to resolve in both confirmer and broadcaster.

PR #371, opened 2024-03-14
Cairo 2.6.3 + OZ 0.10

archseer:


contracts/src/tests/test_ownable.cairo
```diff
@@ -4,155 +4,10 @@ use starknet::testing::set_caller_address;
...
-    fn constructor(ref self: ContractState, owner: ContractAddress) {
-        self.ownable.initializer(owner);
-    }
-}
-
-type ComponentState = OwnableComponent::ComponentState<OwnableMock::ContractState>;
-
-fn STATE() -> ComponentState {
-    OwnableComponent::component_state_for_testing()
-}
```
archseer, line 47: This function started failing since I think the method isn't publicly visible outside of the package anymore. We were duplicating OZ tests anyway so I removed them

PR #373, opened 2024-03-14
Starknet Chain Client: Add Block Methods

augustbleeds:
https://smartcontract-it.atlassian.net/browse/BCI-2676

Adding 3 block-related methods so RTSP can begin playing around with fetching blocks and the Batch() method to lay the groundwork for the rest of the methods. Rest of the process should be straightforward after this PR.

Added the ChainId method as well. Turns out the starknet.go provider automatically caches the chainId after first request, so our implementation uses the batchCall always

Original Design Doc: https://docs.google.com/document/d/10jsfp7_VqDv2GFGILmTX2ey-p_yaFyvxPaVszL6kg6g/edit

Open Questions:
* ~For batch calls, the user will need to iterate through the batchElem array and[ cast the Result interface ](https://github.com/smartcontractkit/chainlink-starknet/pull/373/files#diff-eea92d0b7802160d247d8ea230688c30cb2849810928b097331222df6ac5e7a3R171) to the expected return type. This seems like a weird user experience so I'm open to feedback on how to improve this.~ RESOLVED

### DIFF
--- a/relayer/pkg/starknet/chain_client.go
+++ b/relayer/pkg/starknet/chain_client.go
@@ -16,6 +16,7 @@ type FinalizedBlock = starknetrpc.Block
 type StarknetBatchBuilder interface {
 	RequestBlockByHash(h *felt.Felt) StarknetBatchBuilder
 	RequestBlockByNumber(id uint64) StarknetBatchBuilder
+	RequestChainId() StarknetBatchBuilder
 	// RequestLatestPendingBlock() (StarknetBatchBuilder)
 	RequestLatestBlockHashAndNumber() StarknetBatchBuilder
 	// RequestEventsByFilter(f starknetrpc.EventFilter) (StarknetBatchBuilder)
@@ -33,6 +34,15 @@ func NewBatchBuilder() StarknetBatchBuilder {
 	return &batchBuilder{
 		args: nil,
 	}
+}
+
+func (b *batchBuilder) RequestChainId() StarknetBatchBuilder {
+	b.args = append(b.args, gethrpc.BatchElem{
+		Method: "starknet_chainId",
+		Args:   nil,
+		Result: new(string),
+	})
+	return b
 }
 
 func (b *batchBuilder) RequestBlockByHash(h *felt.Felt) StarknetBatchBuilder {
@@ -75,6 +85,7 @@ type StarknetChainClient interface {
 	BlockByHash(ctx context.Context, h *felt.Felt) (FinalizedBlock, error)
 	// only finalized blocks have numbers
 	BlockByNumber(ctx context.Context, id uint64) (FinalizedBlock, error)
+	ChainId(ctx context.Context) (string, error)
 	// only way to get the latest pending block (only 1 pending block exists at a time)
 	// LatestPendingBlock(ctx context.Context) (starknetrpc.PendingBlock, error)
 	// returns block number and block has of latest finalized block
@@ -86,6 +97,25 @@ type StarknetChainClient interface {
 }
 
 var _ StarknetChainClient = (*Client)(nil)
+
+func (c *Client) ChainId(ctx context.Context) (string, error) {
+	// we do not use c.Provider.ChainID method because it caches
+	// the chainId after the first request
+
+	results, err := c.Batch(ctx, NewBatchBuilder().RequestChainId())
+
+	if err != nil {
+		return "", fmt.Errorf("error in ChainId: %w", err)
+	}
+
+	chainId, ok := results[0].Result.(*string)
+
+	if !ok {
+		return "", fmt.Errorf("expected type string block but found: %T", chainId)
+	}
+
+	return *chainId, nil
+}
 
 func (c *Client) BlockByHash(ctx context.Context, h *felt.Felt) (FinalizedBlock, error) {
 	if c.defaultTimeout != 0 {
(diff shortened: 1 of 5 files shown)


relayer/pkg/starknet/chain_client_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+)
+
+var (
+	myChainID = "SN_SEPOLIA"
+	myTimeout = 10 * time.Second
+)
+
+func TestChainClient(t *testing.T) {
```
augustbleeds, line 25: wonder if i should explictly test v0.7 responses?

relayer/pkg/starknet/chain_client.go
```diff
@@ -0,0 +1,60 @@
...
+	if c.defaultTimeout != 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, c.defaultTimeout)
+		defer cancel()
+	}
+
+	block, err := c.Provider.BlockWithTxs(ctx, starknetrpc.BlockID{Hash: h})
+
+	if err != nil {
+		return FinalizedBlock{}, errors.Wrap(err, "error in BlockByHash")
```
krehermann: let's stop use pkg/errors for new development. it's been deprecated for some time now. use the %w directive to wrap errors: fmt.Errorf("error in BlockByHash: %w")

augustbleeds: thanks, didn't know that

relayer/pkg/starknet/chain_client.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+}
+
+func (b *batchBuilder) RequestBlockByHash(h *felt.Felt) (StarknetBatchBuilder, error) {
+	b.args = append(b.args, gethrpc.BatchElem{
+		Method: "starknet_getBlockWithTxs",
+		Args: []interface{}{
+			starknetrpc.BlockID{Hash: h},
+		},
+		Result: &FinalizedBlock{},
```
augustbleeds, line 54: i learned the result is stored within the same BatchElem type. 

and that you need to allocate the memory for the type you expect to be returned, which is json deserialized from the RPC

relayer/pkg/starknet/chain_client.go
```diff
@@ -0,0 +1,106 @@
...
+	gethrpc "github.com/ethereum/go-ethereum/rpc"
+)
+
+// type alias for readibility
+type FinalizedBlock = starknetrpc.Block
+
+// used to create batch requests
+type StarknetBatchBuilder interface {
+	RequestBlockByHash(h *felt.Felt) (StarknetBatchBuilder, error)
+	// RequestBlockByNumber(id uint64) (StarknetBatchBuilder, error)
```
augustbleeds: will be uncommented as the corresponding methods in the chain client are implemented

relayer/pkg/starknet/client.go
```diff
@@ -56,8 +57,9 @@ func NewClient(_chainID string, baseURL string, lggr logger.Logger, timeout *tim
 	}
 
 	client := &Client{
-		Provider: starknetrpc.NewProvider(c),
-		lggr:     lggr,
+		Provider:  starknetrpc.NewProvider(c),
+		EthClient: c,
```
augustbleeds, line 72: add the ethrpc client to the struct so we can take advantage of its batch call capabilities!

relayer/pkg/starknet/chain_client.go
```diff
@@ -0,0 +1,106 @@
...
+		Result: &FinalizedBlock{},
+	})
+	return b, nil
+}
+
+func (b *batchBuilder) Build() []gethrpc.BatchElem {
+	return b.args
+}
+
+type StarknetChainClient interface {
```
archseer, line 83: How different is this from using what we have in `client.go`/`Reader`? We should be able to share the interfaces

augustbleeds: there's definitely some possibility for re-using the reader client methods. but if you mean to combine the interfaces, we've made it a design decision that chain client interface is distinct and set apart because it is dictated by the customer (atlas now) needs.

augustbleeds: For this PR specifically, we introduced BlockByHash and BlockHashAndNumber. 

In terms of replacing, re-using the existing calls, I didn't remove/replace BlockByHash (returns txs) with the already written BlockWithTxHashes (returns tx hashes) because we need the block with txs.

I can simplify the logic of LatestBlockHeight to call BlockHashAndNumber, though. Would you like me to do that?

skippaDaBitFlippa:
> For batch calls, the user will need to iterate through the batchElem array and[ cast the Result interface ](https://github.com/smartcontractkit/chainlink-starknet/pull/373/files#diff-eea92d0b7802160d247d8ea230688c30cb2849810928b097331222df6ac5e7a3R171)to the expected return type. This seems like a weird user experience so I'm open to feedback on how to improve this.

Based on our previous patterns, we typically develop separate apps for different pieces of chain information—such as one app for fetching blocks and another for fetching logs. Therefore, for the app designed to fetch blocks, I anticipate that we will use a `StarknetBatchBuilder` to batch together multiple `BlockByNumber` calls. This approach ensures that all calls are of the same type, facilitating consistent data casting. So I think for us, deducing return type should be easy since they should all be the same type

chris-de-leon-cll (dismissed):

chris-de-leon-cll (dismissed):

chris-de-leon-cll (approved):

PR #376, opened 2024-03-20
add re-estimation on nonce error, cleanup TxStore, misc fixes

cfal:


relayer/pkg/chainlink/txm/txm.go
```diff
@@ -287,17 +287,22 @@ func (txm *starktxm) broadcast(ctx context.Context, publicKey *felt.Felt, accoun
...
+		txm.lggr.Debugw("estimate NextSequence", "unconfirmed", unconfirmed)
 	}
 
 	tx := starknetrpc.InvokeTxnV3{
 		Type:          starknetrpc.TransactionType_Invoke,
 		SenderAddress: account.AccountAddress,
 		Version:       starknetrpc.TransactionV3,
 		Signature:     []*felt.Felt{},
-		Nonce:         nonce,
+		Nonce:         unconfirmed,
```
augustbleeds: will the mempool rejects this transaction if the nonce exists already in the mempool? (asking cause im not sure)

cfal: we only use a nonce known to be unused when we submit (from nonce manager's NextSequence)

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -171,18 +171,11 @@ func (txm *starktxm) broadcastLoop() {
...
 			txm.lggr.Debugw("broadcastLoop: stopped")
 			return
-		default:
-			// skip processing if queue is empty
-			if len(txm.queue) == 0 {
-				continue
-			}
-
-			// preserve tx queue: don't pull tx from queue until client is known to work
+		case tx := <-txm.queue:
```
augustbleeds, line 102: nice catch

relayer/pkg/chainlink/txm/nonce.go
```diff
@@ -64,60 +64,72 @@ func (nm *nonceManager) HealthReport() map[string]error {
...
+	defer nm.lock.RUnlock()
+
+	if err := nm.validate(publicKey); err != nil {
 		return nil, err
 	}
 
-	nm.lock.RLock()
-	defer nm.lock.RUnlock()
-	return nm.n[addr.String()][chainId], nil
+	return nm.n[publicKey.String()], nil
```
archseer: Minor nit, in both cases we validate which is a map lookup, then immediately fetch again. We could rename 'validate' into 'fetch' and have it return the value (we can ignore the value returned in 'Sync')

relayer/pkg/chainlink/txm/txstore.go
```diff
@@ -109,6 +201,41 @@ func (c *ChainTxStore) Confirm(from *felt.Felt, hash string) error {
...
+func (c *ChainTxStore) GetSmallestUnconfirmedNonce(from *felt.Felt) *felt.Felt {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	// empty slice if address isn't found
+	if err := c.validate(from); err != nil {
+		return nil
+	}
+
+	return c.store[from].GetSmallestUnconfirmedNonce()
```
archseer: Same with the validate implementation here

archseer (approved):

PR #391, opened 2024-03-25
relayer/pkg/chainlink/txm: resync nonce on rejection and broadcast errors

cfal:


cfal:
all checks pass - but it'd be nice if we could test the rejected/broadcast error cases easily

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -383,6 +416,44 @@ func (txm *starktxm) logFeederError(ctx context.Context, hash string, f *felt.Fe
...
+	rpcNonce, err := client.AccountNonce(ctx, accountAddress)
+	if err != nil {
+		return fmt.Errorf("failed to check nonce during resync: %+w", err)
+	}
+
+	txStore := txm.accountStore.GetTxStore(accountAddress)
+	currentNonce := txStore.GetNextNonce()
+
+	if rpcNonce.Cmp(currentNonce) == 0 {
+		txm.lggr.Infow("resync nonce skipped, nonce value is the same", "accountAddress", accountAddress, "nonce", currentNonce)
```
augustbleeds, line 446: so because the new estimation is resyncing the nonce already, after a successful estimation, if you get a nonce error during a failed broadcast, then resyncing will lead to the same nonce being returned from the rpc (I think). 

The only time a different nonce would be returned between the estimation sync and broadcast sync is if an earlier tx (from the same address) was just accepted into the mempool during the time after the estimation but before the broadcast. But this can't happen because broadcasts are synchronous, right? So is it even necessary to re-sync the nonce for failed broadcast nonce error?

augustbleeds: although, i guess it doesn't hurt to resync. it'd just be a no-op and would be the safe thing to do.

augustbleeds (approved):

relayer/pkg/chainlink/txm/txstore.go
```diff
@@ -32,18 +32,28 @@ func NewTxStore(initialNonce *felt.Felt) *TxStore {
...
 		}
 	}
+
+	sort.Slice(staleTxs, func(i, j int) bool {
+		a := staleTxs[i]
+		b := staleTxs[j]
+		return a.Nonce.Cmp(b.Nonce) < 0
+	})
+
+	return staleTxs
```
augustbleeds, line 56: just returning this strictly for logging purposes right

relayer/pkg/chainlink/txm/txm.go
cfal: great point, i had a similar thought. we could possibly skip the broadcast re-sync and rely on the confirmLoop/estimate resync only. 

> if you get a nonce error during a failed broadcast, then resyncing will lead to the same nonce being returned from the rpc (I think).

the nonce error during a failed broadcast could be because our nonce is too high, due to a rejected transaction. and then confirmLoop resync has not yet occurred. in which case, the nonce would be different on resync.

> The only time a different nonce would be returned between the estimation sync and broadcast sync is if an earlier tx (from the same address) was just accepted into the mempool during the time after the estimation but before the broadcast.

note the estimation resync only happens when the value is greater - not when it's less, since it could be an RPC issue. eg. we have a local nonce of 9 and the estimate nonce value returned is 8, it wouldn't resync even if it's an error on our end. which it could be, due to a rejected tx.

this is all very theoretical, because we haven't seen a rejected tx after the estimate fee changes yet. but i agree that it seems safe to leave it in for now.

relayer/pkg/chainlink/txm/txstore.go
cfal: yes

relayer/pkg/chainlink/txm/txm.go
cfal: another idea, probably not worth iterating on atm since we are ok with the idea of skipping a few broadcasts - is that for `RPC_NONCE_ERR` on broadcast, we could retry immediately with the value returned from the RPC, if it's different. it's possible that there was no recent resync, the confirmLoop rejected nonce resync hasn't yet occured, and it'd allow us to recover immediately for the transaction being processed.

PR #392, opened 2024-03-26
add field

augustbleeds:
Forgot to add the FeederURL if you copy it from another TOMLConfig.


cfal (commented):
lgtm, do we need any core changes for this to propagate from toml configs?

augustbleeds:
> lgtm, do we need any core changes for this to propagate from toml configs?

yeah, I need to bump the version in core, which I can do after this is merged. Could you approve? 🙏

cfal (approved):

PR #393, opened 2024-03-26
Create monitor without kafka

augustbleeds:
Had to create a new method to create a monitor without Kafka since the previously used `NewMonitor` function imported from chainlink-common comes with Kafka.


sdrug (commented):
1. I wonder if we should get rid of this `withKafka`/`withoutKafka`/`prometheusOnly` etc wording and just stick to what we have. They encourage [here](https://github.com/smartcontractkit/chainlink-common/blob/d529c556ff6b98573d6e442b91d26d322c1066ac/pkg/monitoring/monitor.go#L41) to use this func as a template, so that's what you do and how Starknet monitor looks like.
2. Would you please update the docs too?

augustbleeds:
> 1. I wonder if we should get rid of this `withKafka`/`withoutKafka`/`prometheusOnly` etc wording and just stick to what we have. They encourage [here](https://github.com/smartcontractkit/chainlink-common/blob/d529c556ff6b98573d6e442b91d26d322c1066ac/pkg/monitoring/monitor.go#L41) to use this func as a template, so that's what you do and how Starknet monitor looks like.
> 2. Would you please update the docs too?

The issue is if we use the `NewMonitor` function as a template, it comes with Kafka out of box, which we don't want. Will update the docs though 👍

archseer (approved):

PR #404, opened 2024-04-03
Audit fixes

archseer:
Could use a bit of help determining if I missed any fields to mark as `#[key]`. I'm also not sure if there's a cost difference associated or not?


contracts/src/emergency/sequencer_uptime_feed.cairo
```diff
@@ -161,6 +162,7 @@ mod SequencerUptimeFeed {
 
     #[l1_handler]
     fn update_status(ref self: ContractState, from_address: felt252, status: u128, timestamp: u64) {
+        let from_address: EthAddress = from_address.try_into().unwrap();
```
archseer, line 174: Cairo enforces `from_address` to be a felt252 on the method signature, but we can cast it right after

relayer/pkg/chainlink/ocr2/events.go
```diff
@@ -40,33 +41,38 @@ type NewTransmissionEvent struct {
...
-		const constNumOfElements = 11
+		const observationsLenIndex = 3
+		const constNumOfElements = 9
+		const constNumOfKeys = 2
 
 		if len(eventData) < constNumOfElements {
 			return NewTransmissionEvent{}, errors.New("invalid: event data")
 		}
 
+		if len(event.Keys) < constNumOfKeys {
```
cfal, line 55: would this (and the above `constNumOfElements`) check make more sense as `len(event.Keys) < constNumOfKeys + 1` or even `!=`?

cfal (dismissed):

relayer/pkg/chainlink/ocr2/events.go
archseer: This one should be `+ 1` because of the event id being present, the other one can have a variable number of elements (because of arrays)

PR #414, opened 2024-04-17
Node Balance Monitoring for STOM

augustbleeds:
https://smartcontract-it.atlassian.net/browse/BCI-3128

This PR adds node balance monitoring on Starknet Nodes to monitor the STRK balances
* bumped chainlink-common to get @aalu1418 's NetworkExporter/NetworkSources change
* creates a new NodeBalanceSourceFactory type + ContractBalancesSource type. it gets the node data from RDD
* creates a new NodeBalancesExporterFactory type + NodeBalancesExporter type. it pushes the data to prometheus
* creates a new erc20 client to use for the source factory
* fixed the broken monitoring tests


relayer/pkg/chainlink/erc20/client.go
```diff
@@ -0,0 +1,79 @@
...
+			Selector:        starknetutils.GetSelectorFromNameFelt("decimals"),
+		}
+
+		decimalsRes, err := c.r.CallContract(ctx, ops)
+
+		if err != nil {
+			return nil, fmt.Errorf("couldn't call decimals on erc20: %w", err)
+		}
+
+		c.decimals = starknetutils.FeltToBigInt(decimalsRes[0])
```
aalu1418, line 79: could this ever return a zero len slice?

relayer/pkg/chainlink/erc20/client.go
```diff
@@ -0,0 +1,79 @@
...
+		Selector:        starknetutils.GetSelectorFromNameFelt("balance_of"),
+		Calldata:        []*felt.Felt{accountAddress},
+	}
+
+	balanceRes, err := c.r.CallContract(ctx, ops)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't call balance_of on erc20: %w", err)
+	}
+
+	return starknetutils.FeltToBigInt(balanceRes[0]), nil
```
aalu1418, line 57: might be safer to quickly validate length?

monitoring/pkg/monitoring/exporter_contract_balance.go
```diff
@@ -0,0 +1,80 @@
...
+	log     commonMonitoring.Logger
+	metrics Metrics
+}
+
+func (f *nodeBalancesExporterFactory) NewExporter(params commonMonitoring.ExporterParams) (commonMonitoring.Exporter, error) {
+	return &nodeBalancesExporter{
+		log:         f.log,
+		metrics:     f.metrics,
+		chainConfig: params.ChainConfig,
+		addrsSet:    []ContractAddressWithBalance{},
```
aalu1418: nit: i don't think you need to initialize the empty slice - default behavior

monitoring/pkg/monitoring/exporter_contract_balance.go
```diff
@@ -0,0 +1,80 @@
...
+	defer e.addrsMu.Unlock()
+
+	for _, c := range e.addrsSet {
+		e.metrics.CleanupBalance(c.Address.String(), c.Name, e.chainConfig.GetNetworkID(), e.chainConfig.GetNetworkName(), e.chainConfig.GetChainID())
+	}
+}
+
+func toFloat64(bignum *big.Int) float64 {
+	val, _ := new(big.Float).SetInt(bignum).Float64()
+	return val
```
aalu1418, line 78: ignoring the error? seems important if the number is too large to fit into a Float64

how unlikely is it to happen?

aalu1418 (commented):
overall looks good, just some small concerns

relayer/pkg/chainlink/erc20/client.go
augustbleeds: if it's not an erc20 contract, possibly. I can add a check to make sure it's len == 1

monitoring/pkg/monitoring/exporter_contract_balance.go
augustbleeds: the ignored value is not an error, but it just represents where the actual value was (up or down from the float64) because you lose precision from big.Float to float64.

but in general, it is very unllikely for contract balance to be too large to fit into float64 (	-1.7e+308 to +1.7e+308), especially since this is the number of STRK and not FRI (so we've already divided by 10^18)

aalu1418: that's fair, and yeah the max is a very large number

my personal preference is to throw an error/log if something is too big / small to fit - but it also makes sense that it is highly unlikely

aalu1418 (approved):

PR #418, opened 2024-04-25
BCI-3127: report length

augustbleeds:
Adds the `report_observations` metric, like solana did in https://github.com/smartcontractkit/chainlink-solana/pull/658

### DIFF
--- a/monitoring/pkg/monitoring/exporter_transmission_details.go
+++ b/monitoring/pkg/monitoring/exporter_transmission_details.go
@@ -1,0 +1,78 @@
+package monitoring
+
+import (
+	"context"
+	"fmt"
+
+	relayMonitoring "github.com/smartcontractkit/chainlink-common/pkg/monitoring"
+)
+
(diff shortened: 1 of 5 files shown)


monitoring/pkg/monitoring/exporter_transmission_details.go
```diff
@@ -0,0 +1,78 @@
...
+		observationLength := float64(t.ObservationLength)
+		p.metrics.SetReportObservations(
+			observationLength,
+			p.feedConfig.ContractAddress,
+			p.feedConfig.GetID(),
+			p.chainConfig.GetChainID(),
+			p.feedConfig.GetContractStatus(),
+			p.feedConfig.GetContractType(),
+			p.feedConfig.Name,
+			p.feedConfig.Path,
```
aalu1418, line 59: probably doesn't matter but why use two different methods of getting `Name` or `Path`?
(the cleanup uses the `Get<>` methods instead)

aalu1418 (approved):

PR #421, opened 2024-04-26
BCI-3127: erc20 client reads u256 not felt252

augustbleeds:
We have to join the two felt252s and interpret one as the lower 128 bits + higher 128 bits as per https://www.starknetjs.com/docs/guides/define_call_message/#u256


relayer/pkg/chainlink/erc20/client_test.go
```diff
@@ -57,7 +57,7 @@ func TestERC20Client(t *testing.T) {
 					out = []byte(`{"result":["0x1"]}`)
 				case starknetutils.GetSelectorFromNameFelt("balance_of").String():
 					// latest transmission details response
-					out = []byte(`{"result":["0x0"]}`)
```
aalu1418, line 60: can you leave a comment here to indicate why there is two hex strings rather than just one?

aalu1418 (approved):

PR #423, opened 2024-04-30
BCI-3127: add gas price metrics

augustbleeds:
Starknet happens to report gas price inside of the transmission event, so it's rather easy to retrieve that data for metrics reporting


monitoring/pkg/monitoring/metrics.go
```diff
@@ -19,12 +23,30 @@ type Metrics interface {
...
+		},
+	)
 	reportObservations = promauto.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Name: "report_observations",
 			Help: "Reports # of observations included in a transmission report",
 		},
 		[]string{
+			// uses "account_address" instead of "contract_address" for consistency with solana dashboards
 			"account_address",
```
aalu1418, line 50: hmmm...maybe `contract_address` is the best name for starknet? that way it's consistent with the other starknet metrics?

is there a specific reason it has to align with solana naming?

augustbleeds: it should align with solana naming conventions because Data Feeds is using the same dashboard for both chains

aalu1418 (approved):

PR #430, opened 2024-05-14
golangci-lint: enable containedctx

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-3203

> containedctx is a linter that detects struct contained context.Context field. This is discouraged technique in favour of passing context as first argument of method or function. For rationale please read [Contexts and structs](https://go.dev/blog/context-and-structs) the Go blog post.


relayer/pkg/chainlink/ocr2/client.go
```diff
@@ -171,10 +168,10 @@ func (c *Client) LinkAvailableForPayment(ctx context.Context, address *felt.Felt
 		Selector:        starknetutils.GetSelectorFromNameFelt("link_available_for_payment"),
 	})
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to call the contract with selector 'link_available_for_payment'")
+		return nil, fmt.Errorf("failed to call the contract with selector 'link_available_for_payment': %w", err)
 	}
-	if len(results) != 2 {
-		return nil, errors.Wrap(err, "insufficient data from selector 'link_available_for_payment'")
+	if l := len(results); l != 2 {
+		return nil, fmt.Errorf("insufficient data from selector 'link_available_for_payment': need 2 results but got %d", l)
```
jmank88, line 175: This was accidentally wrapping a nil error, and so always returning nil. Now that this failure path is fixed, `TestOCR2Client/get_link_available_for_payment` is failing.

relayer/pkg/chainlink/config/config.go
jmank88, line 1: @aalu1418 Do I need to change these ValidateConfig implementations?

aalu1418: i don't think so once this PR is merged: https://github.com/smartcontractkit/chainlink/pull/13310

.golangci.yml
```diff
@@ -5,12 +5,17 @@ linters:
     - exhaustive
     - exportloopref
     - revive
-    # - goimports
+    - goimports
```
augustbleeds, line 8: thanks 👍 , these are just the same configs as in `chainlink` repo i'm guessing? wonder if it's overkill to include the config in the github action

augustbleeds: by the way, what code editor do you use? is your code editor able to use golangci-lint ? i guess that's one reason to keep it as is

relayer/pkg/starknet/chain_client.go
```diff
@@ -13,30 +13,30 @@ import (
...
 // used to create batch requests
-type StarknetBatchBuilder interface {
-	RequestBlockByHash(h *felt.Felt) StarknetBatchBuilder
-	RequestBlockByNumber(id uint64) StarknetBatchBuilder
-	RequestChainId() StarknetBatchBuilder
-	// RequestLatestPendingBlock() (StarknetBatchBuilder)
-	RequestLatestBlockHashAndNumber() StarknetBatchBuilder
-	// RequestEventsByFilter(f starknetrpc.EventFilter) (StarknetBatchBuilder)
-	// RequestTxReceiptByHash(h *felt.Felt) (StarknetBatchBuilder)
+type BatchBuilder interface {
```
augustbleeds, line 16: yeah I guess this makes sense. it's already in starknet repo so no need to add the starknet-prefix. 

the linter didn't pick that up though right?! that's just your eye for detail i presume?

relayer/pkg/chainlink/relay.go
```diff
@@ -3,34 +3,31 @@ package chainlink
...
-	relaytypes "github.com/smartcontractkit/chainlink-common/pkg/types"
 )
 
 var _ relaytypes.Relayer = (*relayer)(nil) //nolint:staticcheck
 
 type relayer struct {
-	chain starkchain.Chain
-	ctx   context.Context
+	chain  starkchain.Chain
+	stopCh services.StopChan
```
augustbleeds, line 21: Is there a reason you prefer a channel over ctx? With ctx we can easily stack ctxs on top of each other.

also, I just noticed that ctx (and stopCh) aren't being used 🤣 . Do other chain relayer pass them into the providers?  I notice solana[ attempts to](https://github.com/smartcontractkit/chainlink-solana/blob/develop/pkg/solana/relay.go), at least passes it to the configprovider constructor...but it remains unused.

augustbleeds (approved):
had some questions but looks good to me. thanks for doing this 🥇

relayer/pkg/chainlink/relay.go
augustbleeds: i see ~ currently reading https://go.dev/blog/context-and-structs 🔍

augustbleeds: so it's mainly about not being ambiguous with the ctx? makes sense to me.

augustbleeds (approved):

relayer/pkg/starknet/chain_client.go
jmank88: The linter flags these now.

.golangci.yml
jmank88: The configs overlap a lot, but not entirely (e.g. depguard list), and we have needed to make gradual changes in the past too.

relayer/pkg/chainlink/relay.go
jmank88: That Solana link just looks like something leftover that could be cleaned up.

PR #433, opened 2024-05-21
Batching MVP

chris-de-leon-cll:
# Gauntlet Batching MVP

## Context
- Batching is a necessary requirement for Starknet production mainnet support, and should be considered a blocker since it is not currently supported
- Adding support for batching to gauntlet classic could take a full sprint to complete (this includes implementation + code testing), which could lead to delays
- To reduce this risk, the proposed workaround is to add batching support to the commands that matter the most - this should take less than a sprint to complete and helps keep us on track with the launch date

## Overview of Changes
- This PR adds rudimentary batching support to `set_config`, `set_billing`, `accept_ownership`, and `transfer_ownership`, which should cover the most frequent use cases
- When multiple contract addresses are passed as arguments to an execute command, the operation will be applied to each of the contracts using a single transaction
- New test cases have been added for each of these commands as well

### DIFF
--- a/packages-ts/starknet-gauntlet/src/commands/base/executeCommand.ts
+++ b/packages-ts/starknet-gauntlet/src/commands/base/executeCommand.ts
@@ -188,13 +188,27 @@ export const makeExecuteCommand = <UI, CI>(config: ExecuteCommandConfig<UI, CI>)
 
     // TODO: This will be required for Multisig
     makeMessage = async (): Promise<Call[]> => {
-      const contract = new Contract(this.contract.abi, this.contractAddress, this.provider.provider)
-      const invocation = contract.populate(
-        config.internalFunction || config.action,
-        this.input.contract as any,
-      )
+      const makeInvocation = (contractAddress: string) => {
+        const contract = new Contract(this.contract.abi, contractAddress, this.provider.provider)
+        return contract.populate(
+          config.internalFunction || config.action,
+          this.input.contract as any,
+        )
+      }
 
-      return [invocation]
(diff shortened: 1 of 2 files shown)


packages-ts/starknet-gauntlet/src/commands/base/executeCommand.ts
```diff
@@ -188,13 +188,19 @@ export const makeExecuteCommand = <UI, CI>(config: ExecuteCommandConfig<UI, CI>)
...
-      const invocation = contract.populate(
-        config.internalFunction || config.action,
-        this.input.contract as any,
-      )
-
-      return [invocation]
+      // If more than one argument is passed in, then the following assumptions will be made:
+      //
+      //  - All `args` are valid Starknet contract addresses
+      //  - All contract addresses reference contracts with the same ABI
```
augustbleeds, line 194: looks good to me, but how often is it that you will have the same inputs to different contracts?

chris-de-leon-cll: I actually have the same question here 👀 

If we do need to support batching with different inputs, then my understanding is that we'd need to modify the flags for the relevant commands to take in an array of values (e.g. `--address addr1` would now need to be `--address addr1,addr2,..`)

In my discussions with @archseer, he mentioned that this would be a lot of extra work and that we shouldn't change any of the existing flags

Based on this, I think the answer here is that it should happen frequently enough to make this change worth it. Other cases which involve different inputs to different contracts would have to be covered by G++

augustbleeds (approved):

PR #443, opened 2024-05-31
RDD Batching

chris-de-leon-cll:
This PR builds off the work in https://github.com/smartcontractkit/chainlink-starknet/pull/433. Consider the following Gauntlet command:

```sh
yarn gauntlet ocr2:set_billing \
  --rdd=/reference-data-directory/directory-ethereum-testnet-sepolia-starknet-1.json \
  "0x79c0bc2a03570241c27235a2dca7696a658cbdaae0bad5762e30204b2791aba" \
  "0x31b3f3475b58a55a707b5990a4328d7ac026fede2df4b0aaedca042aeb351a3"
  ...
```

There's two things to note here:
1. Multiple contract addresses are passed as input (so it is assumed we want to execute a batched call)
2. Only one RDD file can be passed to this command

In the current implementation, the Gauntlet runtime will construct one transaction with a series of `set_billing` calls (one for each input argument / address). The parameters to each call will be the same, and the RDD config values corresponding to the first input argument (in this case `0x79c0bc2a03570241c27235a2dca7696a658cbdaae0bad5762e30204b2791aba`) will be used as the parameters for each call.

This PR modifies the above behavior such that the RDD configs for each respective input address are extracted individually to each call and forwarded to the `set_billing` function. 

More generally, any execute-type commands that support the RDD flag now follow the behavior outlined above.

### DIFF
--- a/packages-ts/starknet-gauntlet/src/commands/base/executeCommand.ts
+++ b/packages-ts/starknet-gauntlet/src/commands/base/executeCommand.ts
@@ -63,8 +63,8 @@ export interface ExecuteCommandInstance<UI, CI> {
   executionContext: ExecutionContext
   contract: CompiledContract
   compiledContractHash?: string
-
   input: Input<UI, CI>
+  batchInput?: Map<string, Input<UI, CI>>
 
   makeMessage: () => Promise<Call[]>
   execute: () => Promise<Result<TransactionResponse>>
@@ -87,8 +87,8 @@ export const makeExecuteCommand = <UI, CI>(config: ExecuteCommandConfig<UI, CI>)
     executionContext: ExecutionContext
     contract: CompiledContract
     compiledContractHash?: string
-
     input: Input<UI, CI>
+    batchInput?: Map<string, Input<UI, CI>>
 
     beforeExecute: () => Promise<void>
     afterExecute: (response: Result<TransactionResponse>) => Promise<any>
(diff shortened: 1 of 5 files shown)


packages-ts/starknet-gauntlet/src/commands/base/executeCommand.ts
```diff
@@ -124,27 +129,93 @@ export const makeExecuteCommand = <UI, CI>(config: ExecuteCommandConfig<UI, CI>)
...
+        }
       }
 
       c.input = await c.buildCommandInput(flags, args, env)
 
-      c.beforeExecute = config.hooks?.beforeExecute
-        ? config.hooks.beforeExecute(c.executionContext, c.input, {
-            logger: deps.logger,
-            prompt: deps.prompt,
+      if (c.batchInput != null) {
```
augustbleeds: wondering, to clean up this logic can we just always assume we have a batch? even if it's of size 1?

augustbleeds (dismissed):

packages-ts/starknet-gauntlet/src/commands/base/executeCommand.ts
chris-de-leon-cll: Yup good point - I'll clean this up in a moment

augustbleeds (approved):

PR #444, opened 2024-05-31
implement starknet_getEvent in chain client

skippaDaBitFlippa:
# What
See title

# Why
For realtime to be able to pull block logs using the starknet chain client


augustbleeds (approved):
awesome, good job!

PR #450, opened 2024-06-03
Fix OCR2 deploy command variable casing

chris-de-leon-cll:


augustbleeds (dismissed):

packages-ts/starknet-gauntlet-ocr2/src/commands/ocr2/deploy.ts
```diff
@@ -38,8 +38,8 @@ const makeUserInput = async (flags, args, env): Promise<UserInput> => {
       decimals: aggregator.decimals,
       description: aggregator.name,
       billingAccessController: flags.billingAccessController || env.BILLING_ACCESS_CONTROLLER || '',
-      linkToken: flags.link || env.LINK || '',
```
augustbleeds, line 41: so it needs to be lower case? what about `env.BILLING_ACCESS_CONTROLLER` ?

chris-de-leon-cll: Yup and nice catch - updated the casing for that variable as well!

PR #457, opened 2024-06-06
BCI-3349: Add latest_answer

augustbleeds:
The proxy and aggregator contract should have a view-only function latest_answer. This is the standard for EVM contracts and it helps simplify external adapter logic

### DIFF
--- a/contracts/src/emergency/sequencer_uptime_feed.cairo
+++ b/contracts/src/emergency/sequencer_uptime_feed.cairo
@@ -161,6 +161,13 @@ mod SequencerUptimeFeed {
         fn decimals(self: @ContractState) -> u8 {
             0_u8
         }
+
+        fn latest_answer(self: @ContractState) -> u128 {
+            self._require_read_access();
+            let latest_round_id = self._latest_round_id.read();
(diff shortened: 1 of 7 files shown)


contracts/src/emergency/sequencer_uptime_feed.cairo
chris-de-leon-cll, line 1: Overall looks good! Just have a couple questions:
1. I noticed that `latest_answer` was also added to the sequencer uptime feed, but no test case was created for it - is that intentional?
2. Looks like there's a lint error reported [here](https://github.com/smartcontractkit/chainlink-starknet/actions/runs/9405982207/job/25908370130?pr=457#step:6:16) - could you please reformat the lines when you have a chance?

augustbleeds: thanks, looks like i forgot a file.

chris-de-leon-cll (approved):

archseer (approved):

obelisk:
Seems ok to me. Not a Cairo person but the change seems sensible

PR #463, opened 2024-06-11
Fix Multisig tests, `inspect` op, and `set_signers` op

chris-de-leon-cll:


packages-ts/starknet-gauntlet-multisig/src/wrapper/index.ts
```diff
@@ -75,18 +75,24 @@ export const wrapCommand = <UI, CI>(
...
       c.account = env.account
       c.multisigAddress = env.multisig
+
+      // NOTE: all multisig commands require the multisig address as an argument.
+      //
+      // There's two ways to provide this:
+      //   - set the MULTISIG environment variable to the address of the multisig contract
+      //   - explicitly pass the address to the command (this will override the MULTISIG env var)
+      //
+      c.contractAddress = args[0] ?? c.multisigAddress
```
augustbleeds, line 88: shouldn't we check the multisigAddress first? because for a batch command (the way it is written) wouldn't it take the first rdd contract always?

augustbleeds (approved):

packages-ts/starknet-gauntlet-multisig/src/wrapper/index.ts
chris-de-leon-cll: For batching, there's an additional field called `batchInput` which is used instead of `contractAddress` to access RDD, so I believe we should be good. The original `contractAddress` field is only relevant when one address is passed in

I can change the ordering around to make `env.multisig` go first, but if I'm not mistaken, the consequence is that if you have the `MULTISIG` env var exported and you try to override it by passing a different multisig contract address to a multisig command, then the `MULTISIG` env var will always take priority regardless of the input argument

PR #465, opened 2024-06-13
read pending block not latest

augustbleeds:
the pending block contains finalized txs so it should be used for the most recent data

if you read from the pending block, there's a problem where the block number on chain is actually not queryable via the rpc because the block hasn't been assigned a block number since it's in the `pending` state. a fix has been made to account for this edge case


archseer (approved):

archseer (dismissed):
Let's fix the tests, likely needs an update of the devnet/RPC version

augustbleeds:
> Let's fix the tests, likely needs an update of the devnet/RPC version

thanks, it was missing a mocked response

chris-de-leon-cll (approved):

PR #475, opened 2024-06-28
[BCI-3438] Add stub chain writer method onto starknet relayer

nickcorin:


aalu1418 (dismissed):

silaslenihan (dismissed):

aalu1418 (dismissed):

relayer/pkg/chainlink/relay.go
```diff
@@ -55,6 +55,10 @@ func (r *relayer) HealthReport() map[string]error {
 	return map[string]error{r.Name(): r.Healthy()}
 }
 
+func (r *relayer) NewChainWriter(_ context.Context, _ []byte) (relaytypes.ContractReader, error) {
```
archseer: Shouldn't this be relaytypes.ChainWriter rather than ContractReader?

aalu1418 (approved):

PR #498, opened 2024-08-01
bump ctf

jmank88:


integration-tests/go.mod
```diff
@@ -79,7 +79,7 @@ require (
 	github.com/cespare/xxhash v1.1.0 // indirect
 	github.com/cespare/xxhash/v2 v2.2.0 // indirect
 	github.com/chai2010/gettext-go v1.0.2 // indirect
-	github.com/chaos-mesh/chaos-mesh/api/v1alpha1 v0.0.0-20220226050744-799408773657 // indirect
+	github.com/chaos-mesh/chaos-mesh/api v0.0.0-20240709130330-9f4feec7553f // indirect
```
jmank88, line 82: I had to manually align this one with CTF to get the mod resolution to work.

aalu1418: just a guess, do you need to add this line? I had to add a bunch of things to the integration test go.mod a while back for solana

https://github.com/smartcontractkit/chainlink/blob/3b4c2b58c3ebb04a2261108e758a3419de436a71/integration-tests/go.mod#L488
```
exclude github.com/chaos-mesh/chaos-mesh/api/v1alpha1 v0.0.0-20220226050744-799408773657
```

jmank88: That would have helped since I replaced that version with this one. I think it is only a problem because the tooling assumes that import paths won't step on each other, so each package only belong to one module. Should be fine now that we are past it.

aalu1418 (approved):

PR #503, opened 2024-08-05
update the cache used for LatestRoundRequested

augustbleeds:


relayer/pkg/chainlink/ocr2/transmissions_cache.go
```diff
@@ -133,9 +134,8 @@ func (c *transmissionsCache) LatestRoundRequested(
 	c.tdLock.RLock()
 	defer c.tdLock.RUnlock()
 	configDigest = c.transmissionDetails.Digest
-	epoch = c.transmissionDetails.Epoch
-	round = c.transmissionDetails.Round
-	err = c.assertTransmissionsNotStale()
```
augustbleeds, line 138: got rid of the error check because we don't actually care what the config digest is and asserting an error would actually cause you to transmit more rounds

relayer/pkg/chainlink/ocr2/transmissions_cache.go
```diff
@@ -133,9 +134,8 @@ func (c *transmissionsCache) LatestRoundRequested(
...
+	epoch = 0
+	round = 0
```
aalu1418: the changes look fine - but why is `LatestRoundRequested` implemented in two places? it seems very easy to make changes in one place but forget to make changes in the other

augustbleeds: originally i think it was to have a cached response and save time. libocr uses the transmissions_cache to satisfy the `ContractTransmitter` interface. 

now that we know we can always return 0 I think we don't need the underlying method. i'll edit that out

relayer/pkg/chainlink/ocr2/transmissions_cache.go
augustbleeds: oh wait, actually the Reader needs to implement it because it must conform to the interface was well

augustbleeds: i removed the need to query the chain and to use the lock

relayer/pkg/chainlink/ocr2/transmissions_cache.go
augustbleeds: oh wait, actually the Reader needs to implement it because it must conform to the interface was well

augustbleeds: i removed the need to query the chain and to use the lock

aalu1418: I guess my suggestion is - instead of having zero values implemented twice, why not just call `c.reader.LatestRoundRequested` and add a note saying that this will always return zero values?

that way you only have hard-coded values in one place rather than two places

augustbleeds: oh i see what you mean. will do 👍

aalu1418 (approved):

PR #505, opened 2024-08-06
use gomods; update generated; add CI check for tidy+generate

jmank88:
This PR swaps in [gomods](https://github.com/jmank88/gomods) for tidy and generate commands, updates stale generated files, and adds a CI check for tidiness+generation.


.github/workflows/relayer.yml
jmank88, line 1: We should either make `check-tidy` required, or move it up in to an existing required step.

aalu1418 (approved):

PR #506, opened 2024-08-06
starknet monitoring: paginate events

augustbleeds:
https://smartcontract-it.atlassian.net/browse/BCI-3958

monitor was broken because the pagination was never implemented, so for blocks with more than 10 events of the same type, it may be impossible to find the transmission event

### DIFF
--- a/relayer/pkg/chainlink/ocr2/client.go
+++ b/relayer/pkg/chainlink/ocr2/client.go
@@ -184,6 +184,42 @@ func (c *Client) LinkAvailableForPayment(ctx context.Context, address *felt.Felt
 	return ans, nil
 }
 
+func (c *Client) collectAllEvents(ctx context.Context, block starknetrpc.BlockID, address *felt.Felt, eventKey *felt.Felt, pageSize int) (events []starknetrpc.EmittedEvent, err error) {
+	input := starknetrpc.EventsInput{
+		EventFilter: starknetrpc.EventFilter{
+			FromBlock: block,
+			ToBlock:   block,
+			Address:   address,
+			Keys:      [][]*felt.Felt{{eventKey}}, // skip other event types
+		},
+		ResultPageRequest: starknetrpc.ResultPageRequest{
+			ChunkSize: pageSize,
+		},
+	}
+
+	chunk, err := c.r.Events(ctx, input)
+
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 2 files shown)


relayer/pkg/chainlink/ocr2/client.go
```diff
@@ -200,24 +236,7 @@ func (c *Client) fetchEventsFromBlock(ctx context.Context, address *felt.Felt, e
...
-		},
-		ResultPageRequest: starknetrpc.ResultPageRequest{
-			// ContinuationToken: ,
-			ChunkSize: 10,
-		},
-	}
-	chunk, err := c.r.Events(ctx, input)
-	events = chunk.Events
-
-	// TODO: check events.isLastPage, query more if needed
```
augustbleeds, line 220: this todo was never implemented, so this PR is doing it (basically)

relayer/pkg/chainlink/ocr2/client_test.go
```diff
@@ -125,6 +532,12 @@ func TestOCR2Client(t *testing.T) {
 		fmt.Printf("%+v\n", transmissions)
 	})
 
+	t.Run("get new transmission event", func(t *testing.T) {
+		events, err := client.NewTransmissionsFromEventsAt(context.Background(), contractAddress, 123)
+		require.NoError(t, err)
+		assert.Len(t, events, 15)
```
augustbleeds, line 538: just assert that we did in fact paginate to get 15 results. if we didn't then we'd only get 10 results

relayer/pkg/chainlink/ocr2/client_test.go
```diff
@@ -78,7 +474,18 @@ func TestOCR2Client(t *testing.T) {
 					require.False(t, true, "unsupported contract method %s", reqdata.Selector)
 				}
 			case "starknet_getEvents":
-				out = []byte(blockOutput)
+				eventsReq := starknetrpc.EventsInput{}
+				require.NoError(t, json.Unmarshal(call.Params[0], &eventsReq))
+
+				configSetSelector := starknetutils.GetSelectorFromNameFelt("ConfigSet")
+
+				if eventsReq.EventFilter.Keys[0][0].Equal(configSetSelector) {
```
augustbleeds, line 482: way to distinguish what mocked result to return to the client

jmank88 (approved):

PR #514, opened 2024-08-21
.golangci.yml: update rules

jmank88:
Updating rules and fixing existing issues.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,7 +1,10 @@
-{ stdenv, pkgs, lib, scriptDir }:
-
-with pkgs;
-let
+{
+  stdenv,
+  pkgs,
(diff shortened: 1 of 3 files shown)


shell.nix
```diff
@@ -1,58 +1,62 @@
...
+        nodePackages.typescript-language-server
+        nodePackages.npm
+        python3
+
+        python311Packages.ledgerwallet
+        go
+
+        gopls
+        delve
+        (golangci-lint.override {buildGoModule = buildGo122Module;})
```
krebernisak, line 32: This is the only change, the rest is the formatter

aalu1418 (approved):

PR #517, opened 2024-08-29
ManyChainMultiSig

augustbleeds:
- [x] contract done
- [x] set_config tests 
- [x] set_root tests
- [x] execute tests


contracts/src/tests/test_mcms.cairo
```diff
@@ -0,0 +1,1023 @@
...
+        0,
+        0,
+        0,
+        0,
+        0,
+        0
+    ];
+
+    let mut group_parents = array![
+        0,
```
augustbleeds: when we upgrade to cairo >= 2.7.0 we will be able to do something like [0_u8; 32].into() instead of having this ugly thing

contracts/src/mcms.cairo
```diff
@@ -0,0 +1,610 @@
...
+struct RootMetadata {
+    chain_id: u256,
+    multisig: ContractAddress,
+    pre_op_count: u64,
+    post_op_count: u64,
+    override_previous_root: bool
+}
+
+// todo: maybe use copy
+// todo: figure out how this works off-chain with MCMS since we have a new selector field here
```
calvwang9: Although changing the Op struct is possible, can the selector be encoded in the first byte of data instead?

contracts/src/mcms.cairo
```diff
@@ -0,0 +1,610 @@
...
+                .encode(MANY_CHAIN_MULTI_SIG_DOMAIN_SEPARATOR_OP)
+                .encode(op.chain_id)
+                .encode(op.multisig)
+                .encode(op.nonce)
+                .encode(op.to)
+                .encode(op.selector);
+            // encode the data field by looping through
+            let mut i = 0;
+            while i < op.data.len() {
+                encoded_leaf = encoded_leaf.encode(*op.data.at(i));
```
calvwang9: just checking if this encoding does any padding on each element of data? felt252 is smaller than u256 so might result in a different encoding if so

contracts/src/tests/test_mcms.cairo
```diff
@@ -0,0 +1,1023 @@
+use core::array::{SpanTrait, ArrayTrait};
```
calvwang9: any unit tests for the encoding and merkle proof validation util functions?

a full e2e test from set config -> root -> execute would be great to validate as well

calvwang9 (commented):
💯

contracts/src/tests/test_mcms.cairo
augustbleeds: definitely, the plan was to add more tests in follow up PRs. i thought there was enough changes already for one PR.

contracts/src/mcms.cairo
augustbleeds: all of the `.encode()` do ~no~ padding similar to `abi.encode` for solidity, so we'll fit the current standards

(there is an `.encode_packed` if you'd like to do `abi.encodePacked`)

contracts/src/mcms.cairo
augustbleeds: it's doable to make it the first index of the data array, but I figured that the Op struct is already different since it doesn't have "value" in it (nor will sign value offchain) so might as well make it more readable?

contracts/src/mcms.cairo
calvwang9: I think `abi.encode` does do padding - [inputs less than 32 bytes are padded to 32 bytes](https://medium.com/@scourgedev/deep-dive-into-abi-encode-types-padding-and-disassembly-84472f1b4543#5853). `abi.encodePacked` is the non-padding encoding

contracts/src/mcms.cairo
calvwang9: Have we considered the approach of setting/asserting (offchain) and signing `value == 0`, and thus the Op struct would be exactly the same as EVM? This comes with the benefits of offchain tooling compatibility and consistency across chains, at the cost of readability. 

However I don't think readability (including the additional selector field) is too big of a concern here since the readability of MCM ops comes from proposals describing human-readbale g++ commands, which then get converted into non-human readable raw proposals consisting of smart contract inputs. EVM calldata encodes the function selector in the first 4 bytes of the calldata as well, so its not a foreign concept, and the selector is encoded as a felt meaning its not human-readble in the first place.

That being said, this current approach also works fine and there's no clear winner here - I'll leave it to you to decide on the best approach

contracts/src/tests/test_mcms.cairo
calvwang9: For features introduced in a single PR, yep this is plenty. However tests (especially unit tests) are a way to validate the correctness of these features as you cant really tell just by reading code when reviewing PRs. Some implementations, e.g. the encoding of the `data` field I mentioned above, I wouldn't know if it worked correctly or not, so seeing tests for such components helps build confidence for PR approvals and reduces the number of bugfix prs later on.

contracts/src/mcms.cairo
augustbleeds: sorry yeah I meant *.encode does padding

contracts/src/tests/test_mcms.cairo
augustbleeds: yeah that works -- i can keep this PR open for all the tests then

contracts/src/mcms.cairo
augustbleeds: Yeah I considered that too 🤔 , I prefer readability but also not carying over EVM design decisions to non-EVM chains if they're not absolutely necessary 👍 good thoughts though

contracts/src/mcms.cairo
calvwang9: yep so i was wondering if the data will be encoded here like `encoded = (0x00.. + felt252(chunk1)) + (0x00.. + felt252(chunk2)) + ...`. We might want to be using encode_packed instead, since data should already be encoded calldata?

augustbleeds: The calldata is just an array of felts. Whether it represents a serialized array or a u64 or any time, either encode or encode_packed should work. 

Since every other encoding is unpacked (on Starknet and also on EVM) I don't see a reason to use packed encoding specifically for this case.

contracts/src/tests/test_mcms/test_execute.cairo
```diff
@@ -0,0 +1,343 @@
...
+
+    start_cheat_chain_id_global(1231);
+    let result = safe_mcms.execute(op1, op1_proof);
+
+    match result {
+        Result::Ok(_) => panic!("expect 'wrong chain id'"),
+        Result::Err(panic_data) => {
+            assert(*panic_data.at(0) == 'wrong chain id', *panic_data.at(0));
+        }
+    }
```
calvwang9: copy-paste error?

contracts/src/mcms.cairo
```diff
@@ -0,0 +1,655 @@
...
+                signer_addresses.len() != 0 && signer_addresses.len() <= MAX_NUM_SIGNERS.into(),
+                'out of bound signers len'
+            );
+
+            assert(signer_addresses.len() == signer_groups.len(), 'signer groups len mismatch');
+
+            assert(
+                group_quorums.len() == NUM_GROUPS.into()
+                    && group_quorums.len() == group_parents.len(),
+                'group quorums/parents mismatch'
```
calvwang9: we could update this error message to more accurately reflect the potential other error (len != 32)

contracts/src/tests/test_mcms/test_set_config.cairo
```diff
@@ -0,0 +1,942 @@
...
+#[test]
+#[feature("safe_dispatcher")]
+fn test_set_config_group_tree_malformed() {
+    // 7. test if group_parents[i] is greater than or equal to i (when not 0) there is revert
+    let (_, _, mcms_safe) = setup_mcms_deploy();
+
+    let signer_addresses = array![EthAddressZeroable::zero()];
+    let signer_groups = array![0];
+
+    let mut group_quorums = array![
```
calvwang9: perhaps a util function like `fill_array` will be useful to help make these cleaner as well as other while loops elsewhere

contracts/src/mcms.cairo
calvwang9: Ah ok i think i understand now - i was confused as to how calldata is handled on Starknet.
Calldata on EVM and most other chains usually represents the encoded raw data to be passed in a transaction and is in the format of bytes or a hex string, e.g. EVM calldata is `0x + <4 bytes func selector> + <encoded args>`. However on Starknet it seems like calldata is more akin to "raw args" rather than encoded data and is represented by an array of felts corresponding to args (or arg metadata). I realized that since there's no encoding necessary as it gets passed straight to the called contract, you're right - we are free to choose the encoding method here for the purposes of generating a hash for MCM and just have to make sure it is consistent between the contract and tooling.

calvwang9 (commented):
nice to see in-depth tests 🎉 just minor comments

contracts/src/tests/test_mcms/test_execute.cairo
augustbleeds: good catch

augustbleeds:
thanks @calvwang9 👍 , i pushed a change to address all the comments

calvwang9 (approved):

PR #528, opened 2024-09-25
NONEVM-298: timelock contract

augustbleeds:
This is the last set of contract changes necessary for the upcoming audit. There are some optional contract changes as well we can add further down the road.

- [x] timelock contract (it implements the SRC-5 standard (since we are on Starknet instead of EVM), which is similar to the EIP-165 standard for knowing if a contract adheres to supports certain interfaces. in this case, timelock on EVM is EIP-165 compatible because it needs to be able to perform safe transfers of certain standards of tokens)
- [x] enumerable set component (encapsulates set-related functionality)

### DIFF
--- a/contracts/src/access_control/rbac_timelock.cairo
+++ b/contracts/src/access_control/rbac_timelock.cairo
@@ -37,8 +37,8 @@ trait IRBACTimelock<TContractState> {
     );
     fn cancel(ref self: TContractState, id: u256);
     fn execute_batch(ref self: TContractState, calls: Span<Call>, predecessor: u256, salt: u256);
-    fn update_delay(ref self: TContractState, new_delay: u256);
     fn bypasser_execute_batch(ref self: TContractState, calls: Span<Call>);
+    fn update_delay(ref self: TContractState, new_delay: u256);
     fn block_function_selector(ref self: TContractState, selector: felt252);
     fn unblock_function_selector(ref self: TContractState, selector: felt252);
     fn get_blocked_function_selector_count(self: @TContractState) -> u256;
@@ -219,15 +219,14 @@ mod RBACTimelock {
         ref self: ContractState,
         min_delay: u256,
         admin: ContractAddress,
-        proposers: Span<ContractAddress>,
-        executors: Span<ContractAddress>,
-        cancellers: Span<ContractAddress>,
-        bypassers: Span<ContractAddress>
+        proposers: Array<ContractAddress>,
+        executors: Array<ContractAddress>,
+        cancellers: Array<ContractAddress>,
+        bypassers: Array<ContractAddress>
     ) {
         self.access_control.initializer();
         self.erc1155_receiver.initializer();
         self.erc721_receiver.initializer();
-
         self.access_control._set_role_admin(ADMIN_ROLE, ADMIN_ROLE);
         self.access_control._set_role_admin(PROPOSER_ROLE, ADMIN_ROLE);
         self.access_control._set_role_admin(EXECUTOR_ROLE, ADMIN_ROLE);
@@ -352,20 +351,6 @@ mod RBACTimelock {
             self._after_call(id);
         }
 
-        fn update_delay(ref self: ContractState, new_delay: u256) {
-            self.access_control.assert_only_role(ADMIN_ROLE);
-
-            self
-                .emit(
-                    Event::MinDelayChange(
-                        MinDelayChange {
-                            old_duration: self._min_delay.read(), new_duration: new_delay,
-                        }
-                    )
-                );
-            self._min_delay.write(new_delay);
-        }
-
         fn bypasser_execute_batch(ref self: ContractState, calls: Span<Call>) {
             self._assert_only_role_or_admin_role(BYPASSER_ROLE);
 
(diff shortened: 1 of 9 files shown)


contracts/src/access_control/rbac_timelock.cairo
```diff
@@ -0,0 +1,500 @@
...
+use alexandria_encoding::sol_abi::encode::SolAbiEncodeTrait;
+
+#[derive(Copy, Drop, Serde)]
+struct Call {
+    target: ContractAddress,
+    selector: felt252,
+    data: Span<felt252>,
+}
+
+fn _hash_operation_batch(calls: Span<Call>, predecessor: u256, salt: u256) -> u256 {
```
calvwang9, line 13: im not across the latest Cairo standards, why does this need to be split out into its own fn outside of the module?

contracts/src/access_control/rbac_timelock.cairo
```diff
@@ -0,0 +1,500 @@
...
+            );
+        }
+
+        fn _after_call(ref self: ContractState, id: u256) {
+            assert(self.is_operation_ready(id), 'rbact: operation not ready');
+            self._timestamps.write(id, _DONE_TIMESTAMP);
+        }
+
+        fn _execute(ref self: ContractState, call: Call) {
+            let _response = call_contract_syscall(call.target, call.selector, call.data)
```
calvwang9: does the response contain any useful information? if not, is there any point in assigning it to a variable (or is this a linter thing?)

contracts/src/access_control/rbac_timelock.cairo
```diff
@@ -0,0 +1,500 @@
...
+                            old_duration: self._min_delay.read(), new_duration: new_delay,
+                        }
+                    )
+                );
+            self._min_delay.write(new_delay);
+        }
+
+        //
+        // ONLY ADMIN
+        //
```
calvwang9: update_delay should be under ONLY ADMIN as well right

contracts/src/tests/test_enumerable_set.cairo
```diff
@@ -0,0 +1,151 @@
...
+    assert(mock.at(MOCK_SET_ID, 3) == 0, 'no entry at 3rd index');
+    assert(mock.values(MOCK_SET_ID) == array![300, 200], 'values should match');
+
+    // [200]
+    assert(mock.remove(MOCK_SET_ID, 300), 'remove 300');
+    assert(mock.length(MOCK_SET_ID) == 1, 'length should equal 1');
+    assert(!mock.contains(MOCK_SET_ID, 300), 'does not contain 300');
+    assert(mock.contains(MOCK_SET_ID, 200), 'contains 200');
+    assert(mock.at(MOCK_SET_ID, 1) == 200, 'indexes match');
+    assert(mock.at(MOCK_SET_ID, 2) == 0, 'no entry at 2nd index');
```
calvwang9: an interesting behaviour of this enumerable set implementation - you wouldn't be able to tell if the entry at an index is actually zero or if there is no entry?

contracts/src/tests/test_rbac_timelock.cairo
```diff
@@ -0,0 +1,865 @@
...
+                )
+            ]
+        );
+
+    let (actual_value, actual_toggle) = target.read();
+    assert(actual_value == 0x56162, 'value equal');
+    assert(actual_toggle, 'toggle true');
+
+    assert(timelock.is_operation_done(id), 'operation is done');
+}
```
calvwang9, line 642: could also add another schedule/execute batch with predecessor = first batch to this test to demonstrate that works as intended too

contracts/src/tests/test_rbac_timelock.cairo
```diff
@@ -0,0 +1,865 @@
...
+    let result = safe_timelock.execute_batch(calls, predecessor, salt);
+    match result {
+        Result::Ok(_) => panic!("expect 'rbact: missing dependency'"),
+        Result::Err(panic_data) => {
+            assert(*panic_data.at(0) == 'rbact: missing dependency', *panic_data.at(0));
+        }
+    }
+}
+
+// snforge does not treat contract invocation failures as a panic (yet)
```
calvwang9, line 494: so executing this on a real environment would result in a panic? is the timelock contract able to tell which op failed?

contracts/src/tests/test_enumerable_set.cairo
augustbleeds: yeah, you can't store the 0 value in this implementation. I guess I could solve for this by introducing a special map that only stores 0? EVM doesn't have this problem because storage values can be arrays (not the case in Starknet).

contracts/src/tests/test_rbac_timelock.cairo
augustbleeds: in a real environment you'd get a panic / revert. in general, the EVM version also does not explictly tell you which operation failed

contracts/src/tests/test_enumerable_set.cairo
augustbleeds: actually, i think as long as i do a length check, we can assume that 0's retrieved from indexes less than or equal to the length are valid

contracts/src/access_control/rbac_timelock.cairo
augustbleeds: it doesn't. I split it out so that it could be used as a utility function in tests. I didn't end up using it though

contracts/src/tests/test_rbac_timelock.cairo
augustbleeds: good point, i'll add that 👍

contracts/src/tests/test_enumerable_set.cairo
augustbleeds: i fixed it so that .at() does a length check and panics if the index is invalid. that way if you get back a 0 from .at() it will be valid

calvwang9 (approved):

PR #535, opened 2024-10-01
NONEVM-629: Link Token Changes

augustbleeds:
Based off of the design here: https://docs.google.com/document/d/1VxcXMJtfqh2Sej6NruQ2KtCuNz8t07AKOaaXsyZY2LQ/edit?usp=sharing 

For reviewing, doing a diff between v2/erc667.cairo and v1/erc667.cairo will make the changes more apparent
Likewise, do a diff between v2/LinkToken.cairo and v1/LinkToken.cairo

- [x] contract changes

legacy gauntlet is not needed to change because the new token will only be used on G++

Creates version 2 of the link token, which 
- alters the LINK token constructor to play nicely with the bridge
- adds a set_minter function
- changes ERC677 to use SRC-5 instead of ERC-165
- adds an ERC677ReceiverComponent


contracts/src/libraries/token/v2/erc677.cairo
```diff
@@ -0,0 +1,79 @@
+use starknet::ContractAddress;
```
augustbleeds, line 1: doing a diff  between `v2/erc667.cairo` and `v1/erc667.cairo` will make the changes more apparent

calvwang9 (approved):
have limited context but based off the pr description this looks fine to me 👍

PR #548, opened 2024-10-25
Add Gauntlet++ Op Client and Op Requests into E2E Integration Tests

chray-zhang:
### What
- Added G++ Client into E2E Tests
- Ported all legacy Gauntlet Ops to G++ ops except for Account Deploy
     - Tracking this with: https://smartcontract-it.atlassian.net/browse/NONEVM-654
- Added new configs for G++ Client

**Ops are not yet integrated into E2E Test workflow. That will be captured in this ticket: https://smartcontract-it.atlassian.net/jira/software/c/projects/NONEVM/boards/357/backlog?selectedIssue=NONEVM-291**

### Why:
https://smartcontract-it.atlassian.net/browse/NONEVM-330


ops/gauntlet/gauntlet_plus_plus_starknet.go
```diff
@@ -0,0 +1,313 @@
...
+	sgpp *StarknetGauntletPlusPlus
+)
+
+type Request struct {
+	Input map[string]any 		 `json:"input"`
+	Command string					 `json:"command"`
+}
+
+type StarknetGauntletPlusPlus struct {
+	client				*g.ClientWithResponses
```
augustbleeds: you can run linting with `make format-go` but before that it seems like you need to tidy the go mods with `make gomodtidy` ~ a version conflict came up when I did that so you just need to fix the version bumping

ops/gauntlet/gauntlet_plus_plus_starknet.go
```diff
@@ -0,0 +1,313 @@
...
+	gr      			*http.Response
+	providers 		*[]g.Provider
+}
+
+func (sgpp *StarknetGauntletPlusPlus) BuildProviders(address string, rpcUrl string, privateKey string) (*[]g.Provider) {
+	var input map[string]*interface{}
```
augustbleeds: people tend to use `any` or `interface{}` now

augustbleeds: main benefit is to make the code look more readable

ops/gauntlet/gauntlet_plus_plus_starknet.go
```diff
@@ -0,0 +1,313 @@
...
+	input = make(map[string]*interface{})
```
augustbleeds: it is probably best practice to use a new inputMap for each provider instead of passing the same input map to each one. it doesn't matter here but if you had the same keys they'd overwrite each other

ops/gauntlet/gauntlet_plus_plus_starknet.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+
+	providers := []g.Provider{AccountProvider, SignerProvider, RpcProvider}
+
+	return &providers
+}
+
+// New StarknetGauntletPlusPlus creates a default g++ client with responses
+func NewStarknetGauntletPlusPlus(gauntletPPEndpoint string, rpcUrl string, address string, privateKey string) (*StarknetGauntletPlusPlus, error) {
+	fmt.Println("rpcUrl: " + rpcUrl)
```
augustbleeds: you can use the logger instead of printing it out

ops/gauntlet/gauntlet_plus_plus_starknet.go
```diff
@@ -0,0 +1,313 @@
...
+)
+
+type Request struct {
+	Input map[string]any 		 `json:"input"`
+	Command string					 `json:"command"`
+}
+
+type StarknetGauntletPlusPlus struct {
+	client				*g.ClientWithResponses
+	gr      			*http.Response
```
augustbleeds: what's `gr` stand for and where is it used? is it the acronym for "gauntlet response"?

ops/gauntlet/gauntlet_plus_plus_starknet.go
```diff
@@ -0,0 +1,313 @@
...
+			Args: &args,
+			Name: request.Command,
+		},
+	}
+
+	return &body
+}
+
+func (sgpp *StarknetGauntletPlusPlus) executeRequest(command string, inputMap map[string]interface{}) error {
+	request := Request{
```
augustbleeds: if you're doing this each time inside of an `execute*` method , I would just enforce the argument to be of Request type.

augustbleeds: you can also name this method to just `execute` since the arg is of type Request and is self-explanatory what type it expects

ops/gauntlet/gauntlet_plus_plus_starknet.go
```diff
@@ -0,0 +1,313 @@
...
+	if err != nil {
+		return err // Handle post execution error
+	}
+
+	// Show Response Status
+	log.Info().Str("Response Status:", string(response.Status())).Msg("Gauntlet++")
+	return nil
+}
+
+func (sgpp *StarknetGauntletPlusPlus) executeRequestReturnsReport(command string, inputMap map[string]interface{}) (g.Report, error) {
```
augustbleeds: if you change the parameter to just Request this function can be called `executeWithReport`

ops/gauntlet/gauntlet_plus_plus_starknet.go
```diff
@@ -0,0 +1,313 @@
...
+	headers := &g.PostExecuteParams{}
+	response, err := sgpp.client.PostExecuteWithResponse(context.Background(), headers, *body)
+	if err != nil {
+		return g.Report{}, err // Handle post execution error
+	}
+
+	return *response.JSON200, nil
+}
+
+func (sgpp *StarknetGauntletPlusPlus) executeDeployRequest(command string, inputMap map[string]interface{}) (string, error) {
```
augustbleeds: likewise, this can just be `executeDeploy` or just `deploy`

ops/gauntlet/gauntlet_plus_plus_starknet.go
```diff
@@ -0,0 +1,313 @@
...
+	if report.Output != nil {
+		// Attempt to assert the Output as a map
+		if outputMap, ok := (*report.Output).(map[string]interface{}); ok {
+			if value, exists := outputMap[key]; exists {
+				// Assert value to a string
+				if strValue, ok := value.(string); ok {
+					fmt.Println("Value:", strValue)
+					return strValue
+				} else {
+					fmt.Println("Value is not of type string")
```
augustbleeds: will the code function work properly if it doesn't find the value? i'd think no, since you'd be missing info that the tests are dependent on.

so, instead of just printing out the issue, you can return an error here

so you'll change your return type of the method to `string, error` 

and in the function that calls ExtractValueFromResponseBody you check if there is an error. If so, you can log the error and either bubble the error up or just do nothing depending on what behavior you want. In this case, I think we should probably bubble the error up since we expect to get a value

ops/gauntlet/gauntlet_plus_plus_starknet.go
chray-zhang: Yup. Do you think I should expand it to `response`?

ops/gauntlet/gauntlet_plus_plus_starknet.go
chray-zhang: >people tend to use any or interface{} now

g.Provider requires input to be of type `*map[string]interface{}`

So I have to pass in some sort of pointer

chray-zhang: I'll clean it up and push a revision

ops/gauntlet/gauntlet_plus_plus_starknet.go
chray-zhang: Added better error catching

augustbleeds:
looks good! now we're just waiting for the github token to pull in the private go modules

.github/actions/build-test-image/action.yml
```diff
@@ -32,6 +35,11 @@ runs:
         tag: ${{ inputs.tag }}
         AWS_REGION: ${{ inputs.QA_AWS_REGION }}
         AWS_ROLE_TO_ASSUME: ${{ inputs.QA_AWS_ROLE_TO_ASSUME }}
+    - name: Setup Go with private repo access
```
augustbleeds, line 38: because you do this step a couple times, let's roll it up into it's on github action in the local .github repo

.github/workflows/golangci-lint.yml
```diff
@@ -74,10 +89,25 @@ jobs:
...
     name: Golang Lint Integration Tests
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
     needs: [golangci-lint-version]
     steps:
       - name: Checkout sources
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - name: Setup GitHub Token
```
augustbleeds, line 98: this as well (making it into its own github action)

.golangci.yml
```diff
@@ -147,4 +147,9 @@ issues:
     - path: test
       text: "^G404:"
       linters:
-        - gosec
\ No newline at end of file
+        - gosec
+    # TODO: re-add gpp op file once we integrate into workflows
```
augustbleeds, line 151: can this todo be deleted?

chray-zhang: No it fails the linter because those ops are still not used in the workflow. That'll be in the next PR after this [one](https://github.com/smartcontractkit/chainlink-starknet/pull/550) is merged

augustbleeds (approved):

PR #612, opened 2025-04-29
Bump Starknet to 0.8.0, juno to 0.12.5, and starknet js package to 0.7.0

chray-zhang:
- Bump juno to 0.12.5
- Bump starknet go module to 0.8.0
- Bump starknet js to 0.7.0

Default use v3 transactions and v08 RPC which aligns with starknet June changes 
core ref: 5fa0dace39f754f4b10c94b5fcc8ffdf762886b6


jmank88 (dismissed):

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -212,9 +212,14 @@ func (txm *starktxm) broadcast(ctx context.Context, publicKey *felt.Felt, accoun
 				MaxAmount:       "0x0",
 				MaxPricePerUnit: "0x0",
 			},
+			// New starknet cannot resolve amounts as 0x0. Minimal max price per unit is: 0x3b9aca00
+			L1DataGas: starknetrpc.ResourceBounds{
+				MaxAmount:       "0x01",
+				MaxPricePerUnit: "0x01",
```
augustbleeds: why is Max set at 0x1?

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -250,47 +259,54 @@ func (txm *starktxm) broadcast(ctx context.Context, publicKey *felt.Felt, accoun
 
 	// TODO: consider making this configurable
 	// pad estimate to 250% (add extra because estimate did not include validation)
-	gasConsumed := friEstimate.GasConsumed.BigInt(new(big.Int))
+	gasConsumed := friEstimate.L1GasConsumed.BigInt(new(big.Int))
 	expandedGas := new(big.Int).Mul(gasConsumed, big.NewInt(250))
 	maxGas := new(big.Int).Div(expandedGas, big.NewInt(100))
-	tx.ResourceBounds.L1Gas.MaxAmount = starknetrpc.U64(starknetutils.BigIntToFelt(maxGas).String())
+	broadCastTxnV3.InvokeTxnV3.ResourceBounds.L1Gas.MaxAmount = starknetrpc.U64(starknetutils.BigIntToFelt(maxGas).String())
```
augustbleeds: ~if ETH/L1 gas is not accepted anymore for v3 tx, is it necessary to specify the parameters for L1 gas?~

edit: not actually certain L1 gas means ETH because apparently it depends on the tx version.
```
// The gas price (in wei or fri, depending on the tx version) that was used in the cost estimation.
```

I guess it's safer to use both

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -250,47 +259,54 @@ func (txm *starktxm) broadcast(ctx context.Context, publicKey *felt.Felt, accoun
...
+
+	L2gasConsumed := friEstimate.L2GasConsumed.BigInt(new(big.Int))
```
augustbleeds: I don't think we need to specify L2 gas. NOPs currently pay in STRK/FRI and that is using the L1Gas fields

chray-zhang: Confirmed offline that L2 is STRK/FRI and L1 is ETH

jmank88 (dismissed):

relayer/pkg/chainlink/txm/txm.go
chray-zhang: Works with 0x0

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -250,47 +260,54 @@ func (txm *starktxm) broadcast(ctx context.Context, publicKey *felt.Felt, accoun
 
 	// TODO: consider making this configurable
 	// pad estimate to 250% (add extra because estimate did not include validation)
-	gasConsumed := friEstimate.GasConsumed.BigInt(new(big.Int))
+	gasConsumed := friEstimate.L1GasConsumed.BigInt(new(big.Int))
 	expandedGas := new(big.Int).Mul(gasConsumed, big.NewInt(250))
 	maxGas := new(big.Int).Div(expandedGas, big.NewInt(100))
-	tx.ResourceBounds.L1Gas.MaxAmount = starknetrpc.U64(starknetutils.BigIntToFelt(maxGas).String())
+	broadCastTxnV3.InvokeTxnV3.ResourceBounds.L1Gas.MaxAmount = starknetrpc.U64(starknetutils.BigIntToFelt(maxGas).String())
```
augustbleeds: i think we had a bug here. we were setting L1Gas.MaxAmount once here to be overwritten later on again. So we need to remove one of them. 

We might as well keep this one because we already have the GasConsumed so it makes sense to calculate the padding based off of that instead of from calculating backwards from the fee

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -250,47 +260,54 @@ func (txm *starktxm) broadcast(ctx context.Context, publicKey *felt.Felt, accoun
...
+
+	L2gasConsumed := friEstimate.L2GasConsumed.BigInt(new(big.Int))
+	L2expandedGas := new(big.Int).Mul(L2gasConsumed, big.NewInt(250))
```
augustbleeds: this can be changed to 150 since the 250% was never used for the L1Gas field

relayer/pkg/chainlink/txm/txm.go
augustbleeds: You'll want to update the padding to be 150 instead of 250 though

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -250,47 +260,54 @@ func (txm *starktxm) broadcast(ctx context.Context, publicKey *felt.Felt, accoun
...
+	L2gasPrice := friEstimate.L2GasPrice.BigInt(new(big.Int))
 	overallFee := friEstimate.OverallFee.BigInt(new(big.Int)) // overallFee = gas_used*gas_price + data_gas_used*data_gas_price
 
 	// TODO: consider making this configurable
 	// pad estimate to 150% (add extra because estimate did not include validation)
 	gasUnits := new(big.Int).Div(overallFee, gasPrice)
 	expandedGasUnits := new(big.Int).Mul(gasUnits, big.NewInt(150))
 	maxGasUnits := new(big.Int).Div(expandedGasUnits, big.NewInt(100))
-	tx.ResourceBounds.L1Gas.MaxAmount = starknetrpc.U64(starknetutils.BigIntToFelt(maxGasUnits).String())
+	broadCastTxnV3.InvokeTxnV3.ResourceBounds.L1Gas.MaxAmount = starknetrpc.U64(starknetutils.BigIntToFelt(maxGasUnits).String())
```
augustbleeds: And here, we can remove this calculation for `ResourceBounds.L1Gas.MaxAmount` entirely since we already did this earlier

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -250,47 +260,54 @@ func (txm *starktxm) broadcast(ctx context.Context, publicKey *felt.Felt, accoun
...
-	tx.Nonce = nonce
-	// Re-sign transaction now that we've determined MaxFee
-	// TODO: SignInvokeTransaction for V3 is missing so we do it by hand
-	hash, err := account.TransactionHashInvoke(tx)
-	if err != nil {
-		return txhash, err
-	}
-	signature, err := account.Sign(ctx, hash)
+	broadCastTxnV3.InvokeTxnV3.ResourceBounds.L1DataGas.MaxAmount = starknetrpc.U64(friEstimate.L1DataGasConsumed.String())
+	broadCastTxnV3.InvokeTxnV3.ResourceBounds.L1DataGas.MaxPricePerUnit = starknetrpc.U128(friEstimate.L1DataGasPrice.String())
```
augustbleeds: both the L1DataGas.MaxAmount and L1DataGas.MaxPricePerUnit should have the padding as well to be safe. 

Usually the MaxAmount will be 0 so multiplying it by a padding is still 0.

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -250,47 +260,54 @@ func (txm *starktxm) broadcast(ctx context.Context, publicKey *felt.Felt, accoun
 
 	// TODO: consider making this configurable
 	// pad estimate to 250% (add extra because estimate did not include validation)
-	gasConsumed := friEstimate.GasConsumed.BigInt(new(big.Int))
+	gasConsumed := friEstimate.L1GasConsumed.BigInt(new(big.Int))
```
augustbleeds: in this section it's getting hard to track what values we're updating. but we're essentially updating the resourceBounds for the L1Gas, L2Gas, and L1DataGas. And for each one the MaxAmount and MaxPricePerUnit fields. So 6 fields in total.

Maybe you can create a helper method here so it's less confusing and we can avoid similar bugs uncovered in the future. 

You could create a function like

func updateMaxAmountBounds(maxAmount *U64, estimate *big.Int) {}

and 

func updateMaxPricePerUnit(maxAmount *U128, estimate *big.Int) {}

and the logic would be the same, just multiplying by the pad.

or something like that where we can consolidate some of the logic and make it more clear which values are being updated.

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -343,6 +328,20 @@ func (txm *starktxm) broadcast(ctx context.Context, publicKey *felt.Felt, accoun
...
 
+func (txm *starktxm) updateMaxAmountBounds(gasConsumed *big.Int, padding int64) starknetrpc.U64 {
+	expandedGas := new(big.Int).Mul(gasConsumed, big.NewInt(padding))
+	maxGas := new(big.Int).Div(expandedGas, big.NewInt(100))
+
+	return starknetrpc.U64(starknetutils.BigIntToFelt(maxGas).String())
+}
+
+func (txm *starktxm) updateMaxPriceUnitBounds(gasPrice *big.Int, padding int64) starknetrpc.U128 {
+	expandedGasPrice := new(big.Int).Mul(gasPrice, big.NewInt(150))
```
augustbleeds: `padding` instead of `big.NewInt(150)`

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -258,43 +258,28 @@ func (txm *starktxm) broadcast(ctx context.Context, publicKey *felt.Felt, accoun
...
 	}
 
-	// TODO: consider making this configurable
-	// pad estimate to 250% (add extra because estimate did not include validation)
-	gasConsumed := friEstimate.L1GasConsumed.BigInt(new(big.Int))
-	expandedGas := new(big.Int).Mul(gasConsumed, big.NewInt(250))
-	maxGas := new(big.Int).Div(expandedGas, big.NewInt(100))
-	broadCastTxnV3.InvokeTxnV3.ResourceBounds.L1Gas.MaxAmount = starknetrpc.U64(starknetutils.BigIntToFelt(maxGas).String())
-
 	L2gasConsumed := friEstimate.L2GasConsumed.BigInt(new(big.Int))
```
augustbleeds: nit: `L2GasConsumed` (capitalize the g)

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -258,43 +258,28 @@ func (txm *starktxm) broadcast(ctx context.Context, publicKey *felt.Felt, accoun
...
 	L2gasPrice := friEstimate.L2GasPrice.BigInt(new(big.Int))
 	overallFee := friEstimate.OverallFee.BigInt(new(big.Int)) // overallFee = gas_used*gas_price + data_gas_used*data_gas_price
 
 	// TODO: consider making this configurable
 	// pad estimate to 150% (add extra because estimate did not include validation)
-	gasUnits := new(big.Int).Div(overallFee, gasPrice)
-	expandedGasUnits := new(big.Int).Mul(gasUnits, big.NewInt(150))
-	maxGasUnits := new(big.Int).Div(expandedGasUnits, big.NewInt(100))
-	broadCastTxnV3.InvokeTxnV3.ResourceBounds.L1Gas.MaxAmount = starknetrpc.U64(starknetutils.BigIntToFelt(maxGasUnits).String())
+	gasUnits := new(big.Int).Div(overallFee, L1gasPrice)
```
augustbleeds: You should get the l1 gas consumed directly from the estimate. (like you did for L2GasConsumed above)
```
 L1GasConsumed = friEstimate.L1GasConsumed.BigInt(new(big.Int))

 broadCastTxnV3.InvokeTxnV3.ResourceBounds.L1Gas.MaxAmount = txm.updateMaxAmountBounds(L1GasConsumed, 150)
 ```

File: relayer/pkg/chainlink/txm/txm.go [hunk 12]
```diff
@@ -258,43 +258,28 @@ func (txm *starktxm) broadcast(ctx context.Context, publicKey *felt.Felt, accoun
...
-	L2maxGasPrice := new(big.Int).Div(L2expandedGasPrice, big.NewInt(100))
-	broadCastTxnV3.InvokeTxnV3.ResourceBounds.L2Gas.MaxPricePerUnit = starknetrpc.U128(starknetutils.BigIntToFelt(L2maxGasPrice).String())
+	broadCastTxnV3.InvokeTxnV3.ResourceBounds.L1Gas.MaxPricePerUnit = txm.updateMaxPriceUnitBounds(L1gasPrice, 150)
+	broadCastTxnV3.InvokeTxnV3.ResourceBounds.L2Gas.MaxPricePerUnit = txm.updateMaxPriceUnitBounds(L2gasPrice, 150)
 
 	txm.lggr.Infow("Set resource bounds", "L1MaxAmount", tx.ResourceBounds.L1Gas.MaxAmount, "L1MaxPricePerUnit", tx.ResourceBounds.L1Gas.MaxPricePerUnit)
 
-	broadCastTxnV3.InvokeTxnV3.ResourceBounds.L1DataGas.MaxAmount = starknetrpc.U64(friEstimate.L1DataGasConsumed.String())
-	broadCastTxnV3.InvokeTxnV3.ResourceBounds.L1DataGas.MaxPricePerUnit = starknetrpc.U128(friEstimate.L1DataGasPrice.String())
+	L1dataGasConsumed := friEstimate.L1DataGasConsumed.BigInt(new(big.Int))
```
augustbleeds: nit: L1DataGasConsumed

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -258,43 +258,28 @@ func (txm *starktxm) broadcast(ctx context.Context, publicKey *felt.Felt, accoun
...
+	L1dataGasPrice := friEstimate.L1DataGasPrice.BigInt(new(big.Int))
```
augustbleeds: nit: L1DataGasPrice

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -258,43 +258,28 @@ func (txm *starktxm) broadcast(ctx context.Context, publicKey *felt.Felt, accoun
...
-	gasConsumed := friEstimate.L1GasConsumed.BigInt(new(big.Int))
-	expandedGas := new(big.Int).Mul(gasConsumed, big.NewInt(250))
-	maxGas := new(big.Int).Div(expandedGas, big.NewInt(100))
-	broadCastTxnV3.InvokeTxnV3.ResourceBounds.L1Gas.MaxAmount = starknetrpc.U64(starknetutils.BigIntToFelt(maxGas).String())
-
 	L2gasConsumed := friEstimate.L2GasConsumed.BigInt(new(big.Int))
-	L2expandedGas := new(big.Int).Mul(L2gasConsumed, big.NewInt(250))
-	L2maxGas := new(big.Int).Div(L2expandedGas, big.NewInt(100))
-	broadCastTxnV3.InvokeTxnV3.ResourceBounds.L2Gas.MaxAmount = starknetrpc.U64(starknetutils.BigIntToFelt(L2maxGas).String())
+	broadCastTxnV3.InvokeTxnV3.ResourceBounds.L2Gas.MaxAmount = txm.updateMaxAmountBounds(L2gasConsumed, 150)
```
augustbleeds: nit: broadcastTxnV3

augustbleeds (approved):

jmank88 (approved):

PR #646, opened 2025-10-07
Prevent nonce gaps for untracked transactions

dimriou:
### RPC nonce error check
The code fixes how the TXM checks for a nonce error from the RPC by comparing the error message. The old check included an additional object that attempted to deserialize the error, but it resulted in nil, which effectively rendered the check useless.  

### Updated transaction confirmation logic
Transactions are no longer confirmed individually per transaction hash; instead, it uses the latest nonce to confirm transactions with a nonce lower than that (similar to EVM TXMv2). This was done for two reasons. First, the RPC responses are slow and unreliable for transaction status. Second, the old confirmation logic doesn't cover the case where a transaction is no longer tracked (either because the RPC evicted the transaction from mempool or failed internally during broadcast, but returned success).

### Added transaction throttling
Unfortunately, not all RPCs will return an error when broadcasting a transaction with a nonce higher than the pending, and will keep the transaction in the mempool up to a large threshold, similar to EVM. That means, if there is a nonce gap, the TXM will be very slow to detect and resync. This is very crucial, especially after a nonce gap is created and there aren't many transactions in the queue. To fix this, the confirmation loop adds a transaction throttling mechanism for in-flight transactions via a constant value. If that threshold is exceeded, the TXM will do a resync of the local nonce. This doesn't necessarily mean that previous transactions will be overwritten, since the broadcasting loop has its own nonce fast-forwarding mechanism. You can think of this throttling mechanism as a "chance" for the broadcast loop and the TXM to go back and replay the nonce and find gaps. That doesn't mean transactions won't be overwritten, but that's a property the original TXM didn't support anyway.

### Transaction overwrite
A major issue with the old design was that the TXM assumed that a successful message during broadcast means the transaction will always be either on the mempool or mined, which, as explained above, it's not the case. This has led to nonce gap incidents. To protect against that, and enable the `nonceResync` mechanism, we allow the TxStore to overwrite unconfirmed transactions with the same nonce if the nonce resyncs. On some edge cases, it means that a past transaction was replaced by a more recent one, which is already expected by the TXM, but in the majority, this means an untracked transaction was dropped for a new one that will fill the nonce gap.

### DIFF
--- a/relayer/pkg/chainlink/txm/txm.go
+++ b/relayer/pkg/chainlink/txm/txm.go
@@ -188,7 +188,7 @@ func (txm *starktxm) broadcast(ctx context.Context, publicKey *felt.Felt, accoun
 		if accountNonceErr != nil {
 			return txhash, fmt.Errorf("failed to check account nonce during TxStore creation: %+w", accountNonceErr)
 		}
-		newTxStore, createErr := txm.accountStore.CreateTxStore(accountAddress, initialNonce)
+		newTxStore, createErr := txm.accountStore.CreateTxStore(accountAddress, initialNonce, txm.lggr)
 		if createErr != nil {
 			return txhash, fmt.Errorf("failed to create TxStore: %+w", createErr)
 		}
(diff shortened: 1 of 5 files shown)


relayer/pkg/chainlink/txm/txm.go
```diff
@@ -293,17 +294,8 @@ func (txm *starktxm) broadcast(ctx context.Context, publicKey *felt.Felt, accoun
...
-		if !errors.As(err, &dataErr) {
-			return txhash, fmt.Errorf("failed to read EstimateFee error: %T %+v", err, err)
-		}
-		data := dataErr.Data
-		dataStr = fmt.Sprintf("%+v", data)
-		txm.lggr.Errorw("failed to invoke tx", "accountAddress", accountAddress, "error", err, "data", dataStr)
-
-		if strings.Contains(dataStr, RPCNonceErrMsg) {
+		txm.lggr.Errorw("failed to invoke tx", "accountAddress", accountAddress, "error", err)
+		if strings.Contains(err.Error(), RPCNonceErrMsg) {
```
augustbleeds, line 298: this seems like a clean change, but we should verify that err.Error() has the expected error string we're looking for

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -362,56 +354,34 @@ func (txm *starktxm) confirmLoop() {
...
-							txm.lggr.Errorw("resync failed for rejected tx", "error", resyncErr)
-						}
-
-						go txm.logFeederError(ctx, hash, f)
-					}
-
-					if executionStatus == starknetrpc.TxnExecutionStatusREVERTED {
-						// TODO: get revert reason?
-						txm.lggr.Errorw("transaction reverted", "hash", hash)
+				nonce, err := client.AccountNonceLatest(ctx, accountAddress)
```
augustbleeds, line 364: I thought that this was going to supplement the existing confirmation logic. any reason why you wanted to remove it? 

It's much cleaner this way, but we should make a note this will need to be rolled out slowly. it'd actually be helpful to introduce a feature flag service for NOPs for cases like this.

augustbleeds (commented):
couple of comments

relayer/pkg/chainlink/txm/txm.go
dimriou: We verified it during manual testing. `dataStr` was empty, while `err.Error()` had the required error message.

relayer/pkg/chainlink/txm/txm.go
dimriou: That was the initial goal, but confirming transactions individually proved to be much more unreliable than we thought. `GetTransactionStatus` is very slow, and almost every time it will return an `unknown transaction` error when fetching an already broadcasted transaction. This not only doesn't provide any benefits over the new design, but it can potentially generate a lot of false positives for freshly broadcast transactions.

Re feature flag, I don't have any strong preference.

augustbleeds (approved):

augustbleeds:
Soaked in stage and prod testnet. Good to merge! Will follow up with the metrics PR now

PR #655, opened 2025-10-28
Feature/add v2 txm metrics

cawthorne:
# Add v2 TXM Metrics with Per-Account Tracking

Metric can be seen for Starknet Testnet nodes at the endpoint:
[http://10.13.80.235:6688/plugins/starknet.SN_SEPOLIA/metrics](http://10.13.80.235:6688/plugins/starknet.SN_SEPOLIA/metrics)
(Pod IP may rotate).

infra-k8s `preview-stage` PR is here:
https://github.com/smartcontractkit/infra-k8s/pull/32851/files

ArgoCD Staging is here:
https://argo.main.stage.cldev.sh/applications/cl-ocr-starknet-testnet?view=tree&resource=
https://argo.main.stage.cldev.sh/applications/cl-ocr-starknet-testnet?view=list&resource=&node=%2FPod%2Fchainlink%2Fcl-ocr-starknet-testnet-0-67b97d78b5-pkkg5%2F0

## Summary

Implements comprehensive transaction manager metrics following Chainlink-EVM patterns, with per-account granularity and dual telemetry support (Prometheus + OpenTelemetry/Beholder).

## Metrics

All metrics are labeled by `chainID` and `accountAddress` for per-account monitoring:

- **txm_num_broadcasted_transactions**: Counter for successful transaction broadcasts
- **txm_num_confirmed_transactions**: Counter for confirmed transactions (handles re-orgs)
- **txm_num_nonce_gaps**: Counter for nonce gap events requiring resync
- **txm_nonce_rebroadcast**: Counter for nonce rebroadcasts (indicates resync frequency)
- **txm_time_until_tx_confirmed**: Histogram of confirmation durations
- **txm_enqueue_failed**: Counter for enqueue failures (queue full)
- **txm_next_nonce**: Gauge for current next nonce per account

## Implementation

- **Dual telemetry**: Prometheus (via `promauto`) and OpenTelemetry/Beholder metrics recorded in parallel
- **Thread-safe**: Uses `sync.Once` for safe metric initialization
- **Per-account tracking**: All metrics labeled with both `chainID` and `accountAddress`
- **Nonce-based tracking**: Internal transaction tracking uses nonce instead of hash for reliability

## Files Changed

- **relayer/pkg/chainlink/txm/txm.go** - Metric integration with per-account tracking
- **relayer/pkg/chainlink/txm/txm_metrics.go** - Prometheus and Beholder metrics implementation
- **relayer/pkg/chainlink/txm/txstore.go** - Nonce-based transaction tracking
- **Test files** - Comprehensive test coverage

### DIFF
--- a/relayer/pkg/chainlink/config/testdata/config-full.toml
+++ b/relayer/pkg/chainlink/config/testdata/config-full.toml
@@ -6,6 +6,7 @@ OCR2CacheTTL = '3m0s'
 RequestTimeout = '1m3s'
 TxTimeout = '13s'
 ConfirmationPoll = '42s'
+MaxAttempts = 15
 
 [[Nodes]]
 Name = 'primary'
--- a/relayer/pkg/chainlink/config/testdata/config-full.toml
+++ b/relayer/pkg/chainlink/config/testdata/config-full.toml
@@ -6,7 +6,6 @@ OCR2CacheTTL = '3m0s'
 RequestTimeout = '1m3s'
 TxTimeout = '13s'
 ConfirmationPoll = '42s'
-MaxAttempts = 15
 FeeEstimationMaxAttempts = 8
 
 [[Nodes]]
--- a/relayer/pkg/chainlink/config/testdata/config-full.toml
+++ b/relayer/pkg/chainlink/config/testdata/config-full.toml
@@ -7,6 +7,7 @@ RequestTimeout = '1m3s'
 TxTimeout = '13s'
 ConfirmationPoll = '42s'
 MaxAttempts = 15
+FeeEstimationMaxAttempts = 8
 
 [[Nodes]]
 Name = 'primary'
--- a/relayer/pkg/chainlink/txm/txm.go
+++ b/relayer/pkg/chainlink/txm/txm.go
@@ -188,7 +188,7 @@ func (txm *starktxm) broadcast(ctx context.Context, publicKey *felt.Felt, accoun
 		if accountNonceErr != nil {
 			return txhash, fmt.Errorf("failed to check account nonce during TxStore creation: %+w", accountNonceErr)
 		}
-		newTxStore, createErr := txm.accountStore.CreateTxStore(accountAddress, initialNonce)
+		newTxStore, createErr := txm.accountStore.CreateTxStore(accountAddress, initialNonce, txm.lggr)
 		if createErr != nil {
 			return txhash, fmt.Errorf("failed to create TxStore: %+w", createErr)
 		}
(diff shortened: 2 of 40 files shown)


relayer/pkg/chainlink/txm/txm_metrics.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+var (
+	promNumBroadcastedTxs    *prometheus.CounterVec
+	promNumConfirmedTxs      *prometheus.CounterVec
+	promNumNonceGaps         *prometheus.CounterVec
+	promReachedMaxAttempts   *prometheus.GaugeVec
+	promTimeUntilTxConfirmed *prometheus.HistogramVec
+	promEnqueueFailed        *prometheus.CounterVec
+	metricsOnce              sync.Once
```
cll-gg: I think we wanted to do otel metrics from now on instead of prometheus metrics, since otel metrics will be sent to Beholder automatically. 

@pkcll @jmank88: could you confirm?

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -55,9 +65,23 @@ type starktxm struct {
...
 	feederClient *utils.LazyLoad[*starknet.FeederClient]
 	accountStore *AccountStore
+	metrics      TxMetrics
+	chainID      string
+
+	// Track broadcast times for confirmation duration metrics
+	broadcastTimes sync.Map // map[string]time.Time (tx hash -> broadcast time)
+
+	// Track retry attempts for max attempts detection
+	retryAttempts sync.Map // map[string]int (tx hash -> attempt count)
```
augustbleeds: I prefer using nonces instead of hashes to uniquely identify a transaction for broadcastTimes and retryAttempts. EVM txmv2 stores/fetches by nonce (though there is an option to fetch by txm attempt, but that's because you can associated one tx to many tx attempts in EVM txmv2). Since starknet txm also is nonce-centric we should do the same thing here.

By using nonce, these intermediate data structures won't have abandoned/stale txs in the map since they will be overwritten if we resync the nonce in the case of a reorg or sequencer degradation like we saw recently. 

Tracking by nonce also helps because if we tag the metrics by nonce and see if a nonce is being broadcasted multiple times but has no confirmation.

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -371,6 +407,16 @@ func (txm *starktxm) confirmLoop() {
 				txm.lggr.Infow("Confirmation loop", "accountAddress", accountAddress, "latestNonce", nonce,
 					"transactionsConfirmed", confirmed, "highestUnconfirmed", highestUnconfirmed)
 
+				// Increment confirmed transactions metric
+				if confirmed > 0 {
+					txm.metrics.IncrementNumConfirmedTxs(ctx, confirmed)
+					// Record confirmation duration for confirmed transactions
+					// Since we can't track individual transactions, we'll estimate based on confirmation poll interval
```
augustbleeds: If you track by nonce, or even if you track by hash, it should be doable to track individual transactions. We can change `Confirm` to return the list of nonces that were just confirmed on the latest loop or the list of `UnconfirmedTx` though we should probably rename `UnconfirmedTx` to just `Tx` or `TxAttempt` to make it less confusing if it's extracted in the confirmation loop.

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -109,8 +136,16 @@ func (txm *starktxm) broadcastLoop() {
 			hash, err := txm.broadcast(ctx, tx.publicKey, tx.accountAddress, tx.call)
 			if err != nil {
 				txm.lggr.Errorw("transaction failed to broadcast", "error", err, "tx", tx.call)
+				// Track retry attempts for max attempts detection
+				txm.trackRetryAttempt(ctx, hash)
```
augustbleeds: I'm not sure if this metric is tracking "attempts" as we think of them in evm. in the starknet txm at most a transaction is only received once in the txm.queue. If the broadcast is not successful on the first time within the txm (withiin the .broadcast() method) the transaction is dropped . There are not multiple attempts per transaction. Perhaps we can get rid of these metric.

A useful metric in its place would be tracking how many times a nonce is broadcasted (as mentioned in one of the other comments) because that tells us how often we are resyncing and how wide the impact of the resync is. Tagging the metrics with nonce value is an important addition we can add here but not necessary. I think the more metadata we can add the better!

Also the "maxRetryAttempt" value doesn't actually stop any effect anything on the txm side.

relayer/pkg/chainlink/txm/txm_metrics.go
augustbleeds: If adding beholder is not super complex then we should. Do we get beholder for free in all the new chainlink images? 

But the primary point of the PR is to just get metrics on internal nodes first. It'd be nice to have them in prod as well but not a requirement for now.

cawthorne: added beholder metrics

relayer/pkg/chainlink/config/testdata/config-full.toml
```diff
@@ -6,6 +6,7 @@ OCR2CacheTTL = '3m0s'
 RequestTimeout = '1m3s'
 TxTimeout = '13s'
 ConfirmationPoll = '42s'
+FeeEstimationMaxAttempts = 8
```
augustbleeds, line 9: note to self: Comms to Node Ops need to specify this config change.

augustbleeds:
resolved some merge conflicts

relayer/pkg/chainlink/txm/txm.go
```diff
@@ -367,9 +424,25 @@ func (txm *starktxm) confirmLoop() {
...
+					// Record actual confirmation duration for each confirmed transaction
+					now := time.Now()
+					for _, confirmedTx := range confirmedTxs {
+						nonceStr := confirmedTx.Nonce.String()
+						if broadcastTime, ok := txm.broadcastTimes.Load(nonceStr); ok {
+							duration := now.Sub(broadcastTime.(time.Time)).Seconds()
+							txm.metrics.RecordTimeUntilTxConfirmed(ctx, accountAddress.String(), duration)
+							// Clean up the broadcast time entry
+							txm.broadcastTimes.Delete(nonceStr)
+						}
```
augustbleeds, line 454: maybe add a warning log if the broadcastTime is not found?

relayer/pkg/chainlink/txm/txm_metrics.go
```diff
@@ -0,0 +1,161 @@
...
+		}, []string{"chainID", "accountAddress"})
+
+		// Initialize beholder metrics
+		beholderNumBroadcastedTxs, _ = beholder.GetMeter().Int64Counter("txm_num_broadcasted_transactions")
+		beholderNumConfirmedTxs, _ = beholder.GetMeter().Int64Counter("txm_num_confirmed_transactions")
+		beholderNumNonceGaps, _ = beholder.GetMeter().Int64Counter("txm_num_nonce_gaps")
+		beholderTimeUntilTxConfirmed, _ = beholder.GetMeter().Float64Histogram("txm_time_until_tx_confirmed")
+		beholderEnqueueFailed, _ = beholder.GetMeter().Int64Counter("txm_enqueue_failed")
+		beholderNonceRebroadcast, _ = beholder.GetMeter().Int64Counter("txm_nonce_rebroadcast")
+		beholderNextNonce, _ = beholder.GetMeter().Int64Gauge("txm_next_nonce")
```
jmank88: 1. Let's handle these errors, even just with `panic` or `log.Fatal`
2. I don't think you can rely on `beholder.GetMeter()` being set before this `init` func is executed. It also prevents testing smoothly. Why don't we inject `metric.Meter` and construct these where they are used instead?

relayer/pkg/chainlink/txm/txm_metrics.go
```diff
@@ -0,0 +1,175 @@
...
+		}, []string{"chainID", "accountAddress"})
+
+		promNextNonce = promauto.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "txm_next_nonce",
+			Help: "The next nonce that will be used for the account. Updated when transactions are broadcasted or nonce is resynced.",
+		}, []string{"chainID", "accountAddress"})
+
+		// Initialize beholder metrics - these are optional and may not be available in all environments
+		// Errors are ignored as beholder may not be initialized yet or may not be available
+		meter := beholder.GetMeter()
```
jmank88: Please inject this and handle the errors 🙏

jmank88 (changes requested):

relayer/pkg/chainlink/txm/txm_metrics.go
cawthorne: I think this is now resolved, but please re-open if you think it isn't.

jadepark-dev (dismissed):
I'm lacking context of Starknet integration, but approving this so that @cawthorne can unblock o11y

toblich (dismissed):
I lack a lot of context here, but as far as I can tell it LGTM

jmank88 (dismissed):

jmank88 (dismissed):

smickovskid (approved):

PR #656, opened 2025-11-04
bump ctf-build-image for platform option

augustbleeds:
This enables building and testing Chainlink images for both ARM64 and AMD64 architectures. The chainlink stage/prod environments now use arm64 nodes only , but CI tests still depend on AMD64 images so we have to build both. The PR was pointing to some feature branches but now that that the changes have been merged to the main branch of the .github actions repository we can merge these changes in.

I cleaned up the action also to tag the starknet image with the PR commit hash instead of the merged git hash which is confusing since it does not exist and cannot be fetched outside of the initial PR trigger.

After these changes are merged, we'll need to rebase Dimitris's starknet changes and attempt the rollout again. 

Here is evidence the build works in staging without a failed ArgoCD rollout: https://github.com/smartcontractkit/infra-k8s/pull/32462 

Co-pilot's summary of changes:

## Multi-platform Docker builds:
* Changed from single build_chainlink_image job to build_chainlink_images with matrix strategy
* Added support for both linux/arm64 and linux/amd64 platforms
* Uses different runners: ubuntu-24.04-arm for ARM64, ubuntu-24.04 for AMD64

## New multi-platform manifest job:
* Added build_multiplatform_image job that combines ARM64 and AMD64 images into a single manifest
* Uses build-push-docker-manifest action to create unified image

## SHA handling improvements:
* Added SN_SHA environment variable that properly uses PR head SHA instead of merge commit
* Updated all image tags to use SN_SHA instead of github.sha

## Build action upgrade:
* Updated from older ctf-build-image@v0.1.3 to @v1.3.0
* Changed from legacy parameters to new platform, dockerfile, plugin-manifest-overrides format

## Job dependencies updated:
* run_tests now depends on build_multiplatform_image instead of build_chainlink_image

### DIFF
--- a/.github/workflows/integration-tests-smoke.yml
+++ b/.github/workflows/integration-tests-smoke.yml
@@ -21,7 +21,7 @@ concurrency:
 
 env:
   TEST_LOG_LEVEL: debug
-  CL_ECR: ${{ secrets.QA_AWS_ACCOUNT_NUMBER }}.dkr.ecr.${{ secrets.QA_AWS_REGION }}.amazonaws.com/chainlink
+  CL_ECR: ${{ secrets.QA_AWS_ACCOUNT_NUMBER }}.dkr.ecr.${{ secrets.QA_AWS_REGION }}.amazonaws.com
   ENV_JOB_IMAGE: ${{ secrets.QA_AWS_ACCOUNT_NUMBER }}.dkr.ecr.${{ secrets.QA_AWS_REGION }}.amazonaws.com/chainlink-starknet-tests:${{ github.sha }}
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-tests-smoke.yml
```diff
@@ -20,28 +20,42 @@ concurrency:
   cancel-in-progress: true
 
 env:
+  # for PR builds, ${{ github.sha }} is the temporary merge commit, we want the head commit instead
+  SN_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
   TEST_LOG_LEVEL: debug
-  CL_ECR: ${{ secrets.QA_AWS_ACCOUNT_NUMBER }}.dkr.ecr.${{ secrets.QA_AWS_REGION }}.amazonaws.com/chainlink
```
augustbleeds, line 24: these were just unused

cawthorne (approved):

mxiao-cll (approved):

